PR #659, opened 2023-10-30
Optimise default value of number-of-workers

berendt:


osism/commands/worker.py
```diff
@@ -1,17 +1,23 @@
...
 
+import multiprocessing
 import subprocess
 
 from cliff.command import Command
 
 
 class Run(Command):
     def get_parser(self, prog_name):
+        if multiprocessing.cpu_count() <= 8:
```
artificial-intelligence, line 11: if this is expected to work in constrained envs (e.g. docker/vm with partial cpus)
it might be better to use `len(os.sched_getaffinity(0))`
see: https://docs.python.org/3/library/os.html#os.cpu_count

berendt: I think it's fine to rely on the overall usable cores. At the moment we do not have the feature to use limits on the manager processes itself.

PR #743, opened 2024-01-03
Remove all pycache dirs & pyc files

berendt:


osfrickler:
Why do you want to do this? I think it is more useful to have precompiled .pyc in the image rather than python having to do the compilation each time when running the container and thus modifying it?

berendt:
> Why do you want to do this? I think it is more useful to have precompiled .pyc in the image rather than python having to do the compilation each time when running the container and thus modifying it?

I would like the images to be smaller so that the start times are faster. It doesn't take very long to create the bytefiles and it seems to be best practice, at least for Debian packages, to deliver them without pyc/__pycache__.

berendt:
```
with-pyclean                                              latest            331b6916161f   48 seconds ago   546MB
without-pyclean                                           latest            2d24292566c0   5 minutes ago    626MB
```

berendt:
Without pyclean (@Quay): 196.8 MiB (extracted 570MB)
With pyclean (@Quay): 168.2 MiB (extracted 487MB)

PR #760, opened 2024-01-15
Add ng apply commands for collections

berendt:


berendt:
Replaced by https://github.com/osism/python-osism/pull/766.

PR #872, opened 2024-04-26
Add tempest

janhorstmann:
Part of https://github.com/osism/issues/issues/926

### DIFF
--- a/osism/core/enums.py
+++ b/osism/core/enums.py
@@ -110,6 +110,7 @@ VALIDATE_PLAYBOOKS = {
     "ulimits": {"environment": "generic", "runtime": "osism-ansible"},
     "mariadb-backup": {"environment": "kolla", "runtime": "kolla-ansible"},
     "mariadb-recovery": {"environment": "kolla", "runtime": "kolla-ansible"},
+    "tempest": {"environment": "openstack", "runtime": "osism-ansible"},
 }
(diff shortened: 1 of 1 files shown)


osism/core/enums.py
```diff
@@ -110,6 +110,7 @@
     "ulimits": {"environment": "generic", "runtime": "osism-ansible"},
     "mariadb-backup": {"environment": "kolla", "runtime": "kolla-ansible"},
     "mariadb-recovery": {"environment": "kolla", "runtime": "kolla-ansible"},
+    "tempest": {"environment": "openstack", "runtime": "osism-ansible"},
```
berendt, line 113: This is not necessary. You have to add a playbook openstack/tempest.yml in the osism/ansible-playbooks repository. Like for the osism.validations.refstack role: https://github.com/osism/ansible-playbooks/blob/main/playbooks/openstack/refstack.yml

berendt (changes requested):
Change not necessary, wrong place. See inline comment.

osism/core/enums.py
janhorstmann: Got it, [the `validate-tempest.yml` symlink is sufficient](https://github.com/osism/python-osism/blob/main/osism/commands/validate.py#L108)

PR #1149, opened 2024-12-15
Add --delete parameter to the manage images command

berendt:


berendt:
@janhorstmann Please take over. Feel free to close this one and create a new one. Also assigned you the corresponding issue.

janhorstmann:
> @janhorstmann Please take over. Feel free to close this one and create a new one. Also assigned you the corresponding issue.

@berendt ready for review

osism/commands/manage.py
```diff
@@ -251,30 +257,26 @@ def get_parser(self, prog_name):
...
-        if latest:
+        if parsed_args.latest:
             arguments.append("--latest")
-        if hide:
+        if parsed_args.hide:
             arguments.append("--hide")
 
-        if images:
+        if parsed_args.images:
             arguments.append("--images")
```
osfrickler, line 278: nit: this is duplicated in both branches, you could move it in front of the conditional

osism/commands/manage.py
```diff
@@ -225,6 +225,12 @@ def get_parser(self, prog_name):
             help="Hide images that should be deleted",
             action="store_true",
```
osfrickler, line 226: not directly related, but this argument has `store_true` and default=True, which sounds useless to me. more importantly, this should be exclusive in regard to the `--delete` option? Doesn't make sense to both hide and delete images at the same time?

osism/commands/manage.py
```diff
@@ -225,6 +225,12 @@ def get_parser(self, prog_name):
...
         )
+        parser.add_argument(
+            "--delete",
+            default=False,
+            help="Delete images that should be deleted",
+            action="store_true",
```
osfrickler, line 232: the image-manager seems to have an additional `--yes-i-really-know-what-i-do` option, are you sure we want to go with just the simple option in this scenario?

berendt: I think it makes sense to always add `--yes-i-really-know-what-i-do` when `--delete` is set in the OSISM CLI.

PR #1297, opened 2025-04-23
Add CAPI 1.33 image

berendt:


berendt:
Depends on https://github.com/osism/k8s-capi-images/pull/278

PR #1326, opened 2025-05-09
Allow override of ironic parameters in netbox

janhorstmann:
Use custom field `ironic_parameters` of a netbox device to override `ironic_parameters` in `conductor.yml`.
Rather than replacing mappings completely with a given override mapping a recursive merge is done to allow easy extension or override of single parameters.
To allow removal of keys in an override the special `DELETE` keyword may be used.

Part of https://github.com/osism/issues/issues/1224


berendt:
Recheck

PR #1367, opened 2025-05-23
Add baremetal deploy commands

janhorstmann:
Add baremetal deploy commands and generate a configdrive with metadata required for network setup.

Closes:  https://github.com/osism/issues/issues/1256

### DIFF
--- a/osism/commands/baremetal.py
+++ b/osism/commands/baremetal.py
@@ -1,0 +1,63 @@
+# SPDX-License-Identifier: Apache-2.0
+
+from cliff.command import Command
+
+from tabulate import tabulate
+
+from osism.commands import get_cloud_connection
+
+
+class BaremetalList(Command):
+    def get_parser(self, prog_name):
+        parser = super(BaremetalList, self).get_parser(prog_name)
+        parser.add_argument(
+            "--provision-state",
+            default=None,
+            choices=["enroll", "managable", "available", "active", "error"],
+            type=str,
+            help="Only list nodes with the given provision state",
+        )
+        parser.add_argument(
+            "--maintenance",
+            default=False,
+            action="store_true",
+            help="Only list baremetal nodes in maintenance mode",
+        )
+        return parser
+
+    def take_action(self, parsed_args):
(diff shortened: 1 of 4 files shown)


berendt:
@janhorstmann Let's discuss the data format tomorrow in our daily. I just finished the auto generation of the FRR parameters based on the Netbox in the inventory reconciler. We should use the same logic here and have to modify the format of the meta information we currently process in the internal images.

https://github.com/osism/container-image-inventory-reconciler/blob/main/files/netbox/extractors/frr_extractor.py

```
  The FRR extractor automatically generates FRR (Free Range Routing) BGP configuration parameters
  for devices. Here's how it works:

  Two-mode operation:
  1. Manual mode: If device has frr_parameters custom field, use it directly
  2. Automatic mode: Generate parameters from NetBox data

  Automatic generation process:

  1. AS Number Calculation (extractors/frr_extractor.py:14-51)
    - Takes IPv4 from dummy0 interface (e.g., 192.168.45.123)
    - Formula: prefix + 3rd octet (padded) + 4th octet (padded)
    - Example: 192.168.45.123 with prefix 42 → 42045123
    - Can be overridden with frr_local_as custom field
  2. Loopback Address Extraction (extractors/frr_extractor.py:138-186)
    - Finds dummy0 interface on device
    - Extracts first IPv4 and IPv6 addresses
    - Sets frr_loopback_v4 and frr_loopback_v6
  3. Uplink Discovery (extractors/frr_extractor.py:188-261)
    - Finds interfaces with:
        - managed-by-osism tag
      - A label (used as interface name)
      - Connected endpoints
    - Filters to only connections to switches (roles: leaf-switch, spine-switch, switch)
    - For each uplink, calculates remote switch's AS number

  Output format:
  frr_parameters:
    frr_local_as: 42045123
    frr_loopback_v4: 192.168.45.123
    frr_loopback_v6: 2001:db8::123
    frr_uplinks:
      - interface: leaf1
        remote_as: 42042100
      - interface: leaf2
        remote_as: 42042101
```

PR #2163, opened 2026-03-24
Allow burn-in of active nodes through service steps

janhorstmann:


janhorstmann:
> Hey - I've left some high level feedback:
> 
> * When processing multiple nodes (e.g., with `--all`), hitting an active node without `--yes-i-really-really-mean-it` causes an early `return` and stops processing the rest; consider using `continue` so that only the problematic node is skipped instead of aborting the entire operation.
> * The `elif node.provision_state in ["active"]:` condition is effectively a simple equality check; using `== "active"` would be clearer and avoid suggesting that multiple states are being handled.
> 
> Prompt for AI Agents
> Sourcery is free for open source - if you like our reviews please consider sharing them ✨
> Help me be more useful! Please click 👍 or 👎 on each comment and I'll use the feedback to improve your reviews.

Fixed the first point

PR #2172, opened 2026-04-08
Avoid eager vault lookup for YAML secrets

ideaship:
Add load_yaml_file and reuse it in the CLI paths that read OpenStack, RabbitMQ, Octavia, and database secrets from YAML files.

Several commands called get_vault() before checking whether the file was actually vault-encrypted. In deployments that keep secrets as plain YAML, this produced repeated 'Unable to get ansible vault password' errors for a path that is never needed.

Read the file first and only initialize the vault when the content is encrypted. This removes noisy false-error output during checks and deployments while preserving decryption for encrypted secrets.

AI-assisted: Codex

### DIFF
--- a/osism/tasks/openstack.py
+++ b/osism/tasks/openstack.py
@@ -10,7 +10,7 @@ from loguru import logger
 
 from osism import settings, utils
 from osism.tasks import Config, run_command
-from osism.tasks.conductor.utils import get_vault
+from osism.tasks.conductor.utils import load_yaml_file
 
 app = Celery("openstack")
 app.config_from_object(Config)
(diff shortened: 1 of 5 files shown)


osism/tasks/openstack.py
```diff
@@ -432,51 +432,9 @@ def get_cloud_password(cloud):
...
-                    decrypted_secrets = yaml.safe_load(f)
-                logger.debug(
-                    f"Successfully loaded unencrypted secrets file (development mode): {secrets_path}"
-                )
-            except Exception as plain_exc:
-                logger.error(f"Failed to read secrets file as plain YAML: {plain_exc}")
-                return None
+        decrypted_secrets = load_yaml_file(secrets_path)
+        if decrypted_secrets is None:
+            return None
```
osfrickler: this is duplicating and partially obsoleting the check below, I'd suggest to drop this

osfrickler (commented):
lgtm in general

osism/tasks/openstack.py
ideaship: Thank you. Fixed.

osfrickler (approved):

PR #2173, opened 2026-04-08
EVPN Layer-2 features

### LINKED ISSUE #1346: Layer-2 Services via EVPN/VxLAN support in SONiC
We need some layer-2 services on top of EVPN to connect existing firewall setups. This issue shall track all required changes. We will contribute those.

### LINKED ISSUE #1348: Support for Static routes in SONiC
We need the ability to insert static routes (e.g. `0.0.0.0/0` and `::/0`) pointing to a Nexthop reachable through a Layer-2 service (Firewall cluster with VRRP). We can implement this

fzakfeld:
- Support for static default routes: by adding `default_route_ipv4` and `default_route_ipv6` to custom field `sonic_parameters` of a VLAN interface, an IPv4 and IPv6 default route will be installed within the assigned VRF
- Support for L2 VNIs using VxLAN / EVPN: by adding the custom tag `evpn-l2vni` to a VLAN, a VxLAN tunnel entry will be created on all switches using that VLAN. The VNI will be the same as the VLAN ID.
- Support for Static Anycast Gateway (SAG): when assigning an IP address with the role "Anycast" to a VLAN interface, it will be added as an Anycast address. Additionally, SAG will be globally enabled when needed
- Support for EVPN Multihoming: by tagging a portchannel with the custom tag `evpn-lag` it will be shared across other switches

Closes https://github.com/osism/issues/issues/1348
Closes https://github.com/osism/issues/issues/1346

### DIFF
--- a/osism/tasks/conductor/sonic/config_generator.py
+++ b/osism/tasks/conductor/sonic/config_generator.py
@@ -73,6 +73,17 @@ def generate_sonic_config(device, hwsku, device_as_mapping=None, config_version=
     # Get port channel configuration from NetBox first (needed by get_connected_interfaces)
     portchannel_info = detect_port_channels(device)
 
+    # Resolve evpn_system_mac early so it is validated once and passed explicitly later
+    _raw_evpn_mac = device.config_context.get("_evpn_system_mac")
+    evpn_system_mac = (
+        _raw_evpn_mac if isinstance(_raw_evpn_mac, str) and _raw_evpn_mac else None
+    )
+    if _raw_evpn_mac and not evpn_system_mac:
+        logger.warning(
+            f"Device {device.name}: '_evpn_system_mac' in config_context is not a valid string"
+            f" (got {type(_raw_evpn_mac).__name__!r}), ignoring"
+        )
+
     # Get connected interfaces to determine admin_status
     connected_interfaces, connected_portchannels = get_connected_interfaces(
         device, portchannel_info
@@ -274,7 +285,7 @@ def generate_sonic_config(device, hwsku, device_as_mapping=None, config_version=
         config["BREAKOUT_PORTS"].update(breakout_info["breakout_ports"])
 
     # Add port channel configuration
-    _add_portchannel_configuration(config, portchannel_info)
+    _add_portchannel_configuration(config, portchannel_info, evpn_system_mac)
 
     # Add VRF configuration
     _add_vrf_configuration(config, vrf_info, netbox_interfaces)
@@ -2108,17 +2119,20 @@ def _add_vrf_configuration(config, vrf_info, netbox_interfaces):
             )
 
 
-def _add_portchannel_configuration(config, portchannel_info):
+def _add_portchannel_configuration(config, portchannel_info, evpn_system_mac=None):
     """Add port channel configuration from NetBox."""
     if portchannel_info["portchannels"]:
         for pc_name, pc_data in portchannel_info["portchannels"].items():
             # Add PORTCHANNEL configuration
-            config["PORTCHANNEL"][pc_name] = {
+            pc_config = {
                 "admin_status": pc_data["admin_status"],
                 "fast_rate": pc_data["fast_rate"],
                 "min_links": pc_data["min_links"],
                 "mtu": pc_data["mtu"],
             }
+            if pc_data.get("evpn_lag") and evpn_system_mac:
+                pc_config["system_mac"] = evpn_system_mac
+            config["PORTCHANNEL"][pc_name] = pc_config
 
             # Add PORTCHANNEL_INTERFACE configuration to enable IPv6 link-local
             config["PORTCHANNEL_INTERFACE"][pc_name] = {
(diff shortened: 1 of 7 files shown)


osism/tasks/conductor/sonic/config_generator.py
```diff
@@ -73,16 +73,7 @@ def generate_sonic_config(device, hwsku, device_as_mapping=None, config_version=
...
-    _raw_evpn_mac = device.config_context.get("_evpn_system_mac")
-    evpn_system_mac = (
-        _raw_evpn_mac if isinstance(_raw_evpn_mac, str) and _raw_evpn_mac else None
-    )
-    if _raw_evpn_mac and not evpn_system_mac:
-        logger.warning(
-            f"Device {device.name}: '_evpn_system_mac' in config_context is not a valid string"
-            f" (got {type(_raw_evpn_mac).__name__!r}), ignoring"
-        )
+    evpn_system_mac = "02:00:00:00:00:00"
```
osfrickler: where is this mac coming from? is this some random choice or some well known default? might be good to explain in a comment

osism/tasks/conductor/sonic/config_generator.py
```diff
@@ -1736,17 +1739,82 @@ def _add_vlan_configuration(config, vlan_info, netbox_interfaces, device):
...
+                sag_enabled = True
+                config["SAG"][f"{vlan_name}|IPv6"] = {"gwip": ipv6_anycast}
+
+    if sag_enabled:
+        if "SAG_GLOBAL" not in config:
+            config["SAG_GLOBAL"] = {}
+        config["SAG_GLOBAL"]["IP"] = {
+            "IPv4": "enable",
+            "IPv6": "enable",
+            "gwmac": "02:00:10:00:00:00",
```
osfrickler: is this intentionally different from the evpn system mac?

osfrickler (changes requested):
please clean up the commit stack a bit, e.g. https://github.com/osism/python-osism/pull/2173/changes/84b1c84d769eb53bde4514b0686b100b78d6496c should get squashed into the parent, merge commits are not needed ...

that will allow us to apply the commits as they are, without having to squash everything into a single big blob

also some questions regarding the choice of mac addresses

osism/tasks/conductor/sonic/config_generator.py
fzakfeld: Random choice, X2-XX is a locally administered / non vendor range. The port channel MACs are calculated using this + the port channel index. For us it is fine to set this as a default, as this won't be used for far-stretched layer 2 where this could come in conflict with other setups. I'll add a comment regarding this

osism/tasks/conductor/sonic/config_generator.py
fzakfeld: Yes, the system mac is for the port channels / LACP and this one for the (virtual) L3 gateways (what responds to ARP/ND). I'll also add some comment here

jklare:
Hi @osfrickler @fzakfeld , I just stumbled over this and was wondering if there is still something blocking to get this merged and resolve the linked issues. Currently, it looks like this needs a rebase to resolve the merge conflict, but apart from that, I do not see any requested changes any more.

osism/tasks/conductor/sonic/config_generator.py
```diff
@@ -1736,17 +1744,82 @@ def _add_vlan_configuration(config, vlan_info, netbox_interfaces, device):
...
+            if ipv6_anycast:
+                sag_enabled = True
+                config["SAG"][f"{vlan_name}|IPv6"] = {"gwip": ipv6_anycast}
+
+    if sag_enabled:
+        if "SAG_GLOBAL" not in config:
+            config["SAG_GLOBAL"] = {}
+        config["SAG_GLOBAL"]["IP"] = {
+            "IPv4": "enable",
+            "IPv6": "enable",
```
osfrickler, line 1938: I cannot find this in the Sonic User Guide that I have access to, so I wonder whether this might be a new feature that was only recently introduced? This might imply we should better safeguard this by a version check so we don't mistakenly generate config that an older sonic version cannot process?

I also wonder whether we should track ipv4_anycast and ipv6_anycast seperately and only enable the needed AF here?

osfrickler (approved):
I think this needs to wait for https://github.com/osism/python-osism/pull/2297/ and be adapted to it afterwards. It likely would also be helpful to wait for the unit tests to be implemented for this code branch and then be able to add unit tests for the new features (though I don't necessarily expect the submitter to perform that task).

We then also need to run some manual tests that assure that existing usecases do not get broken by these changes.

osism/tasks/conductor/sonic/config_generator.py
fzakfeld: 4.5.0 defenitely supports it, probably earlier versions as well, but I can add a safeguard. Makes sense to track IPv4 and IPv6 seperately, I'll change that as well

berendt:
As discussed, we are closing this since it is no longer needed at this time.

PR #2209, opened 2026-04-25
Add unit tests for osism/tasks/conductor/sonic/{bgp,constants}

berendt:
Covers both modules with 100% statement coverage:

- bgp.py: calculate_local_asn_from_ipv4 (docstring example, CIDR stripping, custom prefix, zero-padding, boundary octets, invalid formats), calculate_minimum_as_for_group (multiple/empty groups, invalid IPs skipped with logger.debug, None/empty primary_ip4 silently skipped, custom prefix, str() conversion of non-str IPs), and find_interconnected_spine_groups delegation to connections.find_interconnected_devices.
- constants.py: invariants of module-level tables (default values, sorted-and-deduped DEFAULT_SONIC_ROLES, sample port-speed mappings across families, HIGH_SPEED_PORTS consistency with PORT_TYPE_TO_SPEED_MAP, HWSKU prefix and uniqueness).

Closes #2198

AI-assisted: Claude Code

PR #2210, opened 2026-04-26
Add unit tests for osism/tasks/conductor/utils

### LINKED ISSUE #2202: Unit tests for osism/tasks/conductor/utils.py
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul infrastructure). First step in **Tier 2** (#2199): the small, mostly-pure helpers in `osism/tasks/conductor/utils.py` that are used everywhere else in the conductor package. Covering them first means later tier issues can rely on them being well-tested.

The five targets are largely pure dict/string logic. The only external dependency is `ansible.parsing.vault.VaultLib` (used by `deep_decrypt` and `load_yaml_file`), which is straightforward to mock.

## Scope

Add `tests/unit/tasks/conductor/test_utils.py` covering five functions in `osism/tasks/conductor/utils.py`. Out of scope (separate issues): `get_redfish_connection`, `_get_conductor_redfish_*`, `check_task_lock_and_exit`, `mask_secrets` / `_mask_secrets_inplace`, `get_vault` (only indirectly via `load_yaml_file`).

### `deep_compare(a, b, updates)` — `utils.py:16`

Recursively walks `a` and writes diffs vs. `b` into `updates`. Pure function, no mocking.

- Identical dicts → `updates` stays empty
- Key in `a`, missing in `b` → recorded in `updates`
- Same key, different scalar values → recorded
- Nested dict where inner value differs → nested entry in `updates`
- Nested dict where inner subtree is fully equal → that branch is removed from `updates` (the cleanup branch)
- Empty `a` → `updates` untouched

### `deep_merge(a, b)` — `utils.py:32`

Recursively merges `b` into `a` in place. Pure function, no mocking.

- Disjoint keys → both kept
- Overlapping scalar key → `b`'s value wins
- Both values are dicts → merged recursively
- Type mismatch (dict vs scalar) → `b`'s value overwrites
- Sentinel `\"DELETE\"` value in `b` → corresponding key removed from `a`
- `\"DELETE\"` for a key that does not exist in `a` → no error
- Mutation: assert `a` is mutated in place and `b` is unchanged

### `deep_decrypt(a, vault)` — `utils.py:47`

Recursively decrypts vault-encrypted strings inside dicts/lists. Mock the `vault` parameter (a `VaultLib` instance) — patch its `is_encrypted` and `decrypt` methods.

- `a is None` → returns immediately, no calls on `vault`
- Dict with one encrypted string value → replaced with decoded/stripped plaintext
- Dict with one plaintext value → untouched (`is_encrypted` returns `False`)
- Nested dict / nested list → recursion descends
- List containing encrypted strings → values replaced in place
- Decrypt raises `Exception` for a dict value → key is removed from the dict
- Decrypt raises `Exception` for a list element → element is left untouched (`pass` branch)
- Verify `decrypt` returns `bytes` and the result is `.decode().strip()`

### `_is_secret_key(key)` — `utils.py:280`

- Non-string input (int, None, bytes) → `False`
- `\"password\"`, `\"PASSWORD\"`, `\"db_password\"` → `True`
- `\"secret\"`, `\"client_secret\"`, `\"SECRET_KEY\"` → `True`
- `\"ironic_osism_foo\"`, `\"IRONIC_OSISM_BAR\"` → `True`
- `\"username\"`, `\"host\"`, `\"ironic_other\"` (does not match the prefix exactly) → `False`
- Empty string → `False`

### `load_yaml_file(path)` — `utils.py:95`

Reads a file, optionally vault-decrypts it, and returns the parsed YAML. Uses `tmp_path` for filesystem cases; patch `osism.tasks.conductor.utils.get_vault` for the encrypted-file case.

- Path does not exist → returns `None` (no exception)
- Plain YAML file → returns parsed dict
- Empty YAML file → returns `None` (yaml.safe_load on empty input)
- Malformed YAML → returns `None` (`yaml.YAMLError` swallowed, debug-logged)
- `OSError` on read (e.g. patch `builtins.open` to raise) → returns `None`
- Encrypted file: patch `VaultLib.is_encrypted` to `True` and patch `get_vault().decrypt` to return valid YAML bytes → returns parsed dict
- Encrypted file where `decrypt` raises `AnsibleError` → returns `None`

## Mocking hints

- For `deep_decrypt`: build a `MagicMock(spec=VaultLib)` and configure `is_encrypted.side_effect` / `decrypt.side_effect` per case — no need to touch the real ansible-vault stack.
- For `load_yaml_file` (encrypted branch): patch `osism.tasks.conductor.utils.get_vault` so the test never reaches `osism.utils.get_ansible_vault_password()` (which would hit Redis).
- For the `is_encrypted` probe inside `load_yaml_file` (line ~105): it constructs a fresh `VaultLib()` just to call `.is_encrypted()`. Either patch `osism.tasks.conductor.utils.VaultLib` or pass content that the real method correctly classifies (a real plaintext file works without any patch).
- File I/O: prefer `tmp_path` over patching `open`, except for the `OSError` case.

## Definition of Done

- [ ] `tests/unit/tasks/__init__.py`, `tests/unit/tasks/conductor/__init__.py` created
- [ ] `tests/unit/tasks/conductor/test_utils.py` covers all listed cases
- [ ] `pytest --cov=osism.tasks.conductor.utils` shows 100 % for the five target functions (other functions in the file may stay uncovered for now)
- [ ] `pipenv run pytest tests/unit/tasks/conductor/test_utils.py` passes locally

## Dependencies

- Part of #2199 (meta). Blocked by #2192 / PR #2193 (test infrastructure).

berendt:
Covers the five small helpers in osism/tasks/conductor/utils.py with 100% line coverage: deep_compare (diff recording with empty-branch cleanup), deep_merge (recursion plus the "DELETE" sentinel), deep_decrypt (VaultLib mocked, exception-drop semantics for dicts vs. lists), _is_secret_key (password/secret/ironic_osism_ matching) and load_yaml_file (missing/empty/malformed/OSError paths plus vault encrypted/decrypt-failure paths). Out-of-scope conductor helpers stay uncovered as the issue prescribes.

Closes #2202

AI-assisted: Claude Code

### DIFF
--- a/osism/tasks/conductor/utils.py
+++ b/osism/tasks/conductor/utils.py
@@ -12,6 +12,8 @@ import sushy
 import urllib3
 import yaml
 
+DELETE_SENTINEL = "DELETE"
+
 
 def deep_compare(a, b, updates):
     """
@@ -31,7 +33,7 @@ def deep_compare(a, b, updates):
 
 def deep_merge(a, b):
     for key, value in b.items():
-        if value == "DELETE":
+        if value == DELETE_SENTINEL:
             # NOTE: Use special string to remove keys
             a.pop(key, None)
         elif (
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -48,8 +48,15 @@ def _install_ansible_stubs() -> None:
         def __init__(self, *args, **kwargs):
             pass
 
-        def is_encrypted(self, *args, **kwargs):
-            return False
+        def is_encrypted(self, data, *args, **kwargs):
+            if isinstance(data, str):
+                data = data.encode("utf-8", errors="ignore")
+            return isinstance(data, (bytes, bytearray)) and bytes(data).startswith(
+                b"$ANSIBLE_VAULT"
+            )
+
+        def decrypt(self, *args, **kwargs):
+            return b""
 
     class VaultSecret:
         def __init__(self, *args, **kwargs):
--- a/tests/unit/tasks/conductor/test_utils.py
+++ b/tests/unit/tasks/conductor/test_utils.py
@@ -1,0 +1,470 @@
+# SPDX-License-Identifier: Apache-2.0
+
+from unittest.mock import MagicMock, patch
+
+import pytest
+from ansible.errors import AnsibleError
+from ansible.parsing.vault import VaultLib
+
+from osism.tasks.conductor.utils import (
+    DELETE_SENTINEL,
+    _is_secret_key,
+    deep_compare,
+    deep_decrypt,
+    deep_merge,
+    load_yaml_file,
+)
+
+# ---------------------------------------------------------------------------
+# deep_compare
+# ---------------------------------------------------------------------------
+
+
+def test_deep_compare_identical_dicts_leaves_updates_empty():
+    a = {"x": 1, "y": "two"}
+    b = {"x": 1, "y": "two"}
+    updates = {}
+
+    deep_compare(a, b, updates)
+
+    assert updates == {}
+
+
+def test_deep_compare_records_missing_key_in_b():
+    a = {"x": 1, "y": 2}
+    b = {"x": 1}
+    updates = {}
+
+    deep_compare(a, b, updates)
+
+    assert updates == {"y": 2}
+
+
+def test_deep_compare_records_changed_scalar_value():
+    a = {"x": 1}
+    b = {"x": 2}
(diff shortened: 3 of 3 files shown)


berendt:
Depends on https://github.com/osism/python-osism/pull/2209.

ideaship (approved):

PR #2237, opened 2026-04-28
Add unit tests for SONiC config_generator orchestrator and service ca…

### LINKED ISSUE #2221: Unit tests for osism/tasks/conductor/sonic/config_generator.py — orchestrator, caches, log/SNMP
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul infrastructure). Part of **Tier 3** (#2199). `osism/tasks/conductor/sonic/config_generator.py` is 2201 LOC and is split across three sub-issues:

- **This issue**: top-level orchestrator (`generate_sonic_config`), metalbox / NTP / DNS caches, log-server and SNMP configuration, `natural_sort_key`.
- Companion issue: port / interface / port-channel / breakout sections.
- Companion issue: BGP / VLAN / Loopback / VRF sections.

The orchestrator pulls together every section helper, so this issue focuses on the high-level glue (delegation, base-config loading, primary-IP / AS routing, version handling) plus the device-independent service caches.

## Scope

Add `tests/unit/tasks/conductor/sonic/test_config_generator_orchestrator.py` covering the items listed below in `osism/tasks/conductor/sonic/config_generator.py`.

### `natural_sort_key(port_name)` — `config_generator.py:56`

- `"Ethernet0"` → `0`
- `"Ethernet120"` → `120`
- `"PortChannel"` (no digits) → `0`

### `generate_sonic_config(device, hwsku, device_as_mapping=None, config_version=None)` — `config_generator.py:62`

This is a glue function. Patch every section helper at its import site:

- `osism.tasks.conductor.sonic.config_generator.get_port_config`
- `...detect_port_channels`, `...detect_breakout_ports`, `...convert_netbox_interface_to_sonic`, `...get_speed_from_port_type`
- `...get_connected_interfaces` (returns `(set(), set())`)
- `...get_device_oob_ip`, `...get_device_vlans`, `...get_device_loopbacks`, `...get_device_interface_ips`
- `...get_device_platform`, `...get_device_hostname`, `...get_device_mac_address`
- `..._get_vrf_info`, `..._get_transfer_role_ipv4_addresses`
- `..._add_port_configurations`, `..._add_interface_configurations`, `..._add_bgp_configurations`
- `..._add_ntp_configuration`, `..._add_dns_configuration`, `..._add_vlan_configuration`, `..._add_loopback_configuration`, `..._add_log_server_configuration`, `..._add_snmp_configuration`, `..._add_vrf_configuration`, `..._add_portchannel_configuration`
- `...get_cached_device_interfaces`
- `..._get_metalbox_ip_for_device`
- `..._add_missing_breakout_ports` (indirectly)

The test focus is on the **orchestrator's own logic**, not the helpers (which have their own issues):

- Returns a dict that contains `DEVICE_METADATA`, `BGP_GLOBALS`, `MGMT_INTERFACE`, `STATIC_ROUTE`, `BREAKOUT_CFG`, `BREAKOUT_PORTS`, `VERSIONS` keys at minimum
- `DEVICE_METADATA["localhost"]` populated with `hostname`, `hwsku`, `platform`, `mac` from the corresponding helpers
- Base config loaded from `/etc/sonic/config_db.json` when present (patch `os.path.exists` + `builtins.open`/`json.load`); resulting `config` is a `deepcopy` (mutating one device's config does **not** affect a second device's)
- Base config absent → starts from a minimal scaffold (verify by patching `os.path.exists` to `False`)
- Base config load raises → starts fresh (warning logged), still produces a usable config
- `device.primary_ip4` set → `BGP_GLOBALS["default"]["router_id"]` is the IPv4 address (without prefix), `local_asn` is set via patched `calculate_local_asn_from_ipv4`
- `device.primary_ip4` set but `calculate_local_asn_from_ipv4` raises `ValueError` → `local_asn` not set, warning logged, function still returns
- `device.primary_ip4` `None` and `device.primary_ip6` set → `router_id` is the IPv6 address, `local_asn` not calculated (only IPv4 path)
- `device_as_mapping` includes the device's id → `local_asn` taken from the mapping, `calculate_local_asn_from_ipv4` not called
- `oob_ip_result` returned by `get_device_oob_ip` → `MGMT_INTERFACE["eth0|<ip>/<prefix>"]` and `STATIC_ROUTE["mgmt|0.0.0.0/0"]` populated; `_get_metalbox_ip_for_device` called for the nexthop
- `oob_ip_result` `None` → `MGMT_INTERFACE`/`STATIC_ROUTE` not populated, `oob_ip` passed to `_add_snmp_configuration` is `None`
- `breakout_info["breakout_cfgs"]` non-empty → merged into `config["BREAKOUT_CFG"]`
- `config_version="4_2_0"` → normalized to `"version_4_2_0"` (prefix added)
- `config_version="version_4_2_0"` → kept as-is
- `config_version` `None` and base config has no DATABASE.VERSION → defaulted to `"version_4_0_1"`
- `config_version` `None` and base config has DATABASE.VERSION already → left unchanged
- Missing optional NetBox fields on interfaces (`speed`, `type`, `tags`) → `netbox_interfaces` still built without raising
- `get_cached_device_interfaces` raises → warning logged, `netbox_interfaces` stays empty, function still completes

### `_load_metalbox_devices_cache()` — `config_generator.py:1400`

Patch `osism.tasks.conductor.sonic.config_generator.utils.nb`. Reset module globals (`_metalbox_devices_cache`, `_metalbox_ip_cache`, `_ntp_servers_cache`) in an `autouse` fixture via `clear_all_caches()`.

- One metalbox with two interfaces (one mgmt-only, one regular) and one Vlan SVI → cache contains all three with `is_vlan` flagged for `Vlan*`, `mgmt_only` skipped
- `nb.dcim.interfaces.filter` raises for one metalbox → that metalbox has empty `interfaces`, warning logged, but other metalboxes still processed
- Top-level `nb.dcim.devices.filter(role="metalbox")` raises → cache reset to `{}`, warning logged
- IP addresses without `address` value → still kept (`if ip_addr.address` filters them)

### `_get_metalbox_ip_for_device(device)` — `config_generator.py:1475`

Use `_load_metalbox_devices_cache` results from a fixture, or set `_metalbox_devices_cache` directly. Patch `get_device_oob_ip`.

- `get_device_oob_ip` returns `None` → returns `None`, cached as `None`
- OOB IP `10.0.0.5/24`, metalbox interface IP `10.0.0.1/24` → returns `"10.0.0.1"`, cached
- OOB IP not in any metalbox subnet → returns `None`, warning logged, cached
- IPv6 metalbox IP (`"2001:db8::1/64"`) skipped (`IPv4Address` raises `ValueError`)
- VLAN interface with matching IP → returns it (logs `"VLAN interface"`)
- Cache hit on second call → no further iteration over `_metalbox_devices_cache`
- `_metalbox_devices_cache` `None` → returns `None`, warning logged
- Outer exception (e.g. malformed OOB IP) → returns `None`, warning logged

### `_get_ntp_servers()` — `config_generator.py:1566`

Patch `utils.nb.dcim.devices.filter`, `utils.nb.dcim.interfaces.filter`, `utils.nb.ipam.ip_addresses.filter`.

- One manager device with `Loopback0` and IPv4 → returned in dict with maxpoll/minpoll/prefer
- One metalbox + one manager → both included
- Loopback0 with IPv6 only → ignored (the `"." in ip_only and ":" not in ip_only` filter)
- No matching devices → empty dict cached
- Exception raised → empty dict cached, warning logged
- Second call → returns cached dict (no second `filter` call)

### `_add_ntp_configuration(config, device)` — `config_generator.py:1624`

Patch `_get_metalbox_ip_for_device`.

- Returns IP → `config["NTP_SERVER"][ip] = {maxpoll, minpoll, prefer}`
- Returns `None` → `config["NTP_SERVER"]` left untouched, warning logged
- Helper raises → warning logged, no exception propagated

### `_add_dns_configuration(config, device)` — `config_generator.py:1670`

Same shape as NTP, but `config["DNS_NAMESERVER"][ip] = {}`. Two cases (IP / no IP) plus exception handling.

### `clear_ntp_cache()` / `clear_metalbox_ip_cache()` / `clear_metalbox_devices_cache()` / `clear_all_caches()` — lines 1649 / 1656 / 1663 / 1691

- Each `clear_*` resets its respective module global to the documented empty value
- `clear_all_caches()` resets all three plus calls `clear_port_config_cache()` (verify by patching `clear_port_config_cache` in the module namespace and asserting it was called)

### `_add_log_server_configuration(config, device)` — `config_generator.py:2105`

Build `device.config_context` as a plain dict.

- `_segment_log_server_hosts=["10.1.1.1", "10.1.1.2"]`, defaults for the rest → both hosts under `SYSLOG_SERVER` with `protocol="UDP"`, `severity="info"`, `vrf_name="mgmt"`, `remote-port="514"`, `message-type="log"`
- `_segment_log_server_proto="tcp"` → `protocol="TCP"` (upper-cased)
- Custom severity/vrf → propagated
- `_segment_log_server_hosts` empty / missing → `SYSLOG_SERVER` not added

### `_add_snmp_configuration(config, device, oob_ip)` — `config_generator.py:2128`

Patch `osism.tasks.conductor.sonic.config_generator.get_vault` and stub the returned vault so `deep_decrypt` is a no-op (or patch `deep_decrypt` directly on the module).

- Defaults: location `"Data Center"`, contact `"info@example.com"`, `traps="enable"` (default `True`)
- `_segment_snmp_server_traps=False` → `traps` key absent from `SNMP_SERVER["SYSTEM"]`
- `oob_ip` set → `SNMP_AGENT_ADDRESS_CONFIG[f"{oob_ip}|161|mgmt"] = {"name": "agentEntry1"}`
- `oob_ip=None` → `SNMP_AGENT_ADDRESS_CONFIG` not added
- `_segment_snmp_server_username` set → user/group-member entries created with `shaKey`/`aesKey` from `node_secrets` (defaults if absent)
- `username` set + `_segment_snmp_server_hosts=["host1"]` → `SNMP_SERVER_PARAMS` and `SNMP_SERVER_TARGET` populated, counter increments per host
- `username` not set → user/group/target sections absent
- `device.custom_fields["secrets"]` `None` → treated as `{}` without raising

## Mocking hints

- Use `pytest` fixtures with `autouse=True` to call `clear_all_caches()` between tests so module-level globals stay isolated.
- Build `device.custom_fields` and `device.config_context` as plain dicts on a `SimpleNamespace`.
- For `generate_sonic_config`, prefer patching the section helpers via `mocker.patch(...)` so the orchestrator test stays focused. Each helper just needs a side-effect that mutates the `config` dict it receives.
- The base config (`/etc/sonic/config_db.json`) test should patch `os.path.exists` and `builtins.open` (with `mock_open(read_data=json.dumps(...))`).
- For `_add_snmp_configuration`, prefer patching `deep_decrypt` and `get_vault` so secrets handling stays a no-op.

## Definition of Done

- [ ] `tests/unit/tasks/conductor/sonic/test_config_generator_orchestrator.py` created
- [ ] `pytest --cov=osism.tasks.conductor.sonic.config_generator` shows ≥ 80 % for the targeted functions (other parts of the file are tracked in companion sub-issues)
- [ ] `pipenv run pytest tests/unit/tasks/conductor/sonic/test_config_generator_orchestrator.py` passes locally

## Dependencies

berendt:
…ches

Covers natural_sort_key, generate_sonic_config (base-config loading, deepcopy, primary-IP / AS routing, OOB management, breakout merge, config_version normalization), the metalbox / NTP / DNS caches, and the log-server / SNMP helpers. Helpers from companion sub-issues are patched at their import sites so this file exercises only the orchestrator's own glue.

Also hardens generate_sonic_config so the top-level scaffold keys it and its helpers index into directly are always present, even when /etc/sonic/config_db.json is absent or fails to load.

Closes #2221

AI-assisted: Claude Code

### DIFF
--- a/osism/tasks/conductor/sonic/config_generator.py
+++ b/osism/tasks/conductor/sonic/config_generator.py
@@ -40,9 +40,6 @@ from .connections import (
 from .cache import get_cached_device_interfaces
 from .constants import BGP_AF_L2VPN_EVPN_TAG, DEFAULT_SONIC_ROLES
 
-# Global cache for NTP servers to avoid multiple queries
-_ntp_servers_cache = None
-
 # Global cache for metalbox IPs per device to avoid duplicate lookups
 _metalbox_ip_cache: dict[int, Optional[str]] = {}
 
@@ -51,6 +48,40 @@ _metalbox_devices_cache: Optional[dict] = None
 
 # VXLAN VTEP name used for VXLAN tunnel configuration
 VXLAN_VTEP_NAME = "vtepServ"
+
+# Top-level scaffold keys that the orchestrator and downstream helpers index
+# into directly. Kept as a single source of truth so that the production code
+# and the test helpers cannot drift apart when keys are added or removed.
+TOP_LEVEL_SCAFFOLD_KEYS = (
+    "DEVICE_METADATA",
+    "BGP_GLOBALS",
+    "BGP_GLOBALS_AF",
+    "BGP_GLOBALS_AF_NETWORK",
+    "BGP_GLOBALS_ROUTE_ADVERTISE",
+    "BGP_NEIGHBOR",
+    "BGP_NEIGHBOR_AF",
+    "MGMT_INTERFACE",
+    "STATIC_ROUTE",
+    "BREAKOUT_CFG",
+    "BREAKOUT_PORTS",
+    "NTP_SERVER",
+    "DNS_NAMESERVER",
+    "PORT",
+    "INTERFACE",
+    "VLAN",
+    "VLAN_INTERFACE",
+    "VLAN_MEMBER",
+    "LOOPBACK",
+    "LOOPBACK_INTERFACE",
+    "PORTCHANNEL",
+    "PORTCHANNEL_INTERFACE",
+    "PORTCHANNEL_MEMBER",
+    "VRF",
+    "VXLAN_TUNNEL",
+    "VXLAN_EVPN_NVO",
+    "VXLAN_TUNNEL_MAP",
+    "VERSIONS",
+)
 
 
 def natural_sort_key(port_name):
@@ -159,6 +190,12 @@ def generate_sonic_config(device, hwsku, device_as_mapping=None, config_version=
         )
         # Ensure we start fresh even on error
         config = {}
+
+    # Ensure the top-level scaffold keys the orchestrator and downstream
+    # helpers index into directly are always present, even when the on-disk
+    # base config is missing or only partially populated.
+    for _scaffold_key in TOP_LEVEL_SCAFFOLD_KEYS:
+        config.setdefault(_scaffold_key, {})
 
     # Update DEVICE_METADATA with NetBox information
     if "localhost" not in config["DEVICE_METADATA"]:
@@ -1563,64 +1600,6 @@ def _get_metalbox_ip_for_device(device):
         return None
 
 
-def _get_ntp_servers():
-    """Get NTP servers from manager/metalbox devices. Uses caching to avoid repeated queries."""
-    global _ntp_servers_cache
-
-    if _ntp_servers_cache is not None:
-        logger.debug("Using cached NTP servers")
-        return _ntp_servers_cache
-
-    ntp_servers = {}
-    try:
-        # Get devices with manager or metalbox device roles
-        devices_manager = utils.nb.dcim.devices.filter(role="manager")
-        devices_metalbox = utils.nb.dcim.devices.filter(role="metalbox")
-
-        # Combine both device lists
-        ntp_devices = list(devices_manager) + list(devices_metalbox)
-        logger.debug(f"Found {len(ntp_devices)} potential NTP devices")
-
-        for ntp_device in ntp_devices:
-            # Get interfaces for this device to find Loopback0
-            device_interfaces = utils.nb.dcim.interfaces.filter(device_id=ntp_device.id)
-
-            for interface in device_interfaces:
-                # Look for Loopback0 interface
-                if interface.name == "Loopback0":
-                    # Get IP addresses assigned to this Loopback0 interface
-                    ip_addresses = utils.nb.ipam.ip_addresses.filter(
-                        assigned_object_id=interface.id,
-                    )
-
-                    for ip_addr in ip_addresses:
-                        if ip_addr.address:
-                            # Extract just the IPv4 address without prefix
-                            ip_only = ip_addr.address.split("/")[0]
-
-                            # Check if it's an IPv4 address (simple check)
-                            if "." in ip_only and ":" not in ip_only:
-                                ntp_servers[ip_only] = {
-                                    "maxpoll": "10",
-                                    "minpoll": "6",
-                                    "prefer": "false",
-                                }
-                                logger.info(
-                                    f"Found NTP server {ip_only} from device {ntp_device.name} with role {ntp_device.role.slug}"
-                                )
-                    break
-
-        # Cache the results
-        _ntp_servers_cache = ntp_servers
-        logger.debug(f"Cached {len(ntp_servers)} NTP servers")
-
-    except Exception as e:
-        logger.warning(f"Could not process NTP servers: {e}")
-        _ntp_servers_cache = {}
-
-    return _ntp_servers_cache
-
-
 def _add_ntp_configuration(config, device):
     """Add NTP_SERVER configuration to device config.
 
@@ -1644,13 +1623,6 @@ def _add_ntp_configuration(config, device):
 
     except Exception as e:
         logger.warning(f"Could not add NTP configuration to device {device.name}: {e}")
-
-
-def clear_ntp_cache():
-    """Clear the NTP servers cache. Should be called at the start of sync_sonic."""
-    global _ntp_servers_cache
-    _ntp_servers_cache = None
-    logger.debug("Cleared NTP servers cache")
 
 
 def clear_metalbox_ip_cache():
@@ -1690,7 +1662,6 @@ def _add_dns_configuration(config, device):
 
 def clear_all_caches():
     """Clear all caches in config_generator module."""
-    clear_ntp_cache()
     clear_metalbox_ip_cache()
     clear_metalbox_devices_cache()
     clear_port_config_cache()
--- a/osism/tasks/conductor/sonic/config_generator.py
+++ b/osism/tasks/conductor/sonic/config_generator.py
@@ -300,7 +300,8 @@ def generate_sonic_config(device, hwsku, device_as_mapping=None, config_version=
         config["MGMT_INTERFACE"]["eth0"] = {"admin_status": "up"}
         config["MGMT_INTERFACE"][f"eth0|{oob_ip}/{prefix_len}"] = {}
         metalbox_ip = _get_metalbox_ip_for_device(device)
-        config["STATIC_ROUTE"] = {}
+        # Write into the existing STATIC_ROUTE dict so any pre-existing
+        # routes loaded from /etc/sonic/config_db.json survive.
         config["STATIC_ROUTE"]["mgmt|0.0.0.0/0"] = {"nexthop": metalbox_ip}
     else:
         oob_ip = None
@@ -320,9 +321,9 @@ def generate_sonic_config(device, hwsku, device_as_mapping=None, config_version=
     # Add VRF configuration
     _add_vrf_configuration(config, vrf_info, netbox_interfaces)
 
-    # Set DATABASE VERSION from config_version parameter or default
-    if "VERSIONS" not in config:
-        config["VERSIONS"] = {}
+    # Set DATABASE VERSION from config_version parameter or default.
+    # VERSIONS itself is guaranteed by the TOP_LEVEL_SCAFFOLD_KEYS setdefault
+    # loop above; only the nested DATABASE entry still needs guarding.
     if "DATABASE" not in config["VERSIONS"]:
         config["VERSIONS"]["DATABASE"] = {}
 
--- a/tests/unit/tasks/conductor/sonic/test_config_generator_ntp.py
+++ b/tests/unit/tasks/conductor/sonic/test_config_generator_ntp.py
@@ -1,0 +1,58 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for ``_add_ntp_configuration`` (per-device NTP wiring)."""
+
+from types import SimpleNamespace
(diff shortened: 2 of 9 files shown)


osism/tasks/conductor/sonic/config_generator.py
```diff
@@ -52,6 +52,32 @@
...
+    "VLAN_INTERFACE",
+    "VLAN_MEMBER",
+    "LOOPBACK",
+    "LOOPBACK_INTERFACE",
+    "PORTCHANNEL",
+    "PORTCHANNEL_INTERFACE",
+    "PORTCHANNEL_MEMBER",
+    "VRF",
+    "VERSIONS",
+)
```
ideaship, line 84: `TOP_LEVEL_SCAFFOLD_KEYS` is missing sections the orchestrator writes into directly: `BGP_NEIGHBOR_AF` (e.g. line 1020), `BGP_NEIGHBOR` (line 1202), `BGP_GLOBALS_AF_NETWORK` (line 1819), `BGP_GLOBALS_AF` (line 2014), `BGP_GLOBALS_ROUTE_ADVERTISE` (line 2035), `VXLAN_TUNNEL`, `VXLAN_EVPN_NVO`, `VXLAN_TUNNEL_MAP` (lines 2067–2086). With a missing or sparse `config_db.json`, any device with connected ports, Loopback0 routes, or VRFs with VNI will raise `KeyError` at runtime. These eight keys need to be added to the tuple.

ideaship (commented):
**[config_generator.py:319] Dead guard introduced by this PR**

The `if "VERSIONS" not in config:` guard at line 319 is now dead: adding `"VERSIONS"` to `TOP_LEVEL_SCAFFOLD_KEYS` guarantees it is present via `setdefault` at line 193. Remove lines 319–320 only — the inner guard at line 321 (`if "DATABASE" not in config["VERSIONS"]:`) is still live because the scaffold only creates an empty dict, not the nested key.

tests/unit/tasks/conductor/sonic/test_config_generator_ntp.py
```diff
@@ -0,0 +1,185 @@
...
+(per-device wiring)."""
+
+from types import SimpleNamespace
+
+import pytest
+
+from osism.tasks.conductor.sonic import config_generator
+from osism.tasks.conductor.sonic.config_generator import (
+    _add_ntp_configuration,
+    _get_ntp_servers,
```
ideaship: `_get_ntp_servers` is not called from production code — `_add_ntp_configuration` reaches the metalbox IP via `_get_metalbox_ip_for_device` directly, never through this function. The associated `_ntp_servers_cache` global and `clear_ntp_cache()` are also unreachable. These tests exercise an orphaned private function. Recommend either removing `_get_ntp_servers`, `_ntp_servers_cache`, and `clear_ntp_cache()` if the NetBox-crawl NTP path is obsolete, or wiring the function back into `_add_ntp_configuration` if it is intended behaviour.

tests/unit/tasks/conductor/sonic/test_config_generator_orchestrator.py
```diff
@@ -0,0 +1,556 @@
...
+    patch_base_config(mocker)
+    patch_orchestrator_helpers.get_device_oob_ip.return_value = ("10.42.0.5", 24)
+    patch_orchestrator_helpers._get_metalbox_ip_for_device.return_value = "10.42.0.1"
+    device = make_orchestrator_device()
+
+    config = generate_sonic_config(device, "HWSKU")
+
+    assert config["MGMT_INTERFACE"]["eth0"] == {"admin_status": "up"}
+    assert "eth0|10.42.0.5/24" in config["MGMT_INTERFACE"]
+    assert config["STATIC_ROUTE"]["mgmt|0.0.0.0/0"] == {"nexthop": "10.42.0.1"}
```
ideaship, line 346: **Missing test** — this test uses an empty scaffold so the `STATIC_ROUTE` overwrite in the production code is invisible. The overwrite itself is fixed in #2238; a companion test is still needed here: seed a base config with a non-empty `STATIC_ROUTE`, trigger the OOB path, and assert the pre-existing entries survive alongside the new `mgmt|0.0.0.0/0` route.

tests/unit/tasks/conductor/sonic/test_config_generator_orchestrator.py
```diff
@@ -0,0 +1,556 @@
...
+    device_b = make_orchestrator_device(device_id=2, name="dev-b")
+
+    config_a = generate_sonic_config(device_a, "HWSKU-A")
+    # Mutate device_a's config; device_b must not see it.
+    config_a["DEVICE_METADATA"]["shared"]["flag"] = "mutated"
+
+    config_b = generate_sonic_config(device_b, "HWSKU-B")
+
+    assert config_b["DEVICE_METADATA"]["shared"] == {"flag": "original"}
+    assert config_a["DEVICE_METADATA"] is not config_b["DEVICE_METADATA"]
```
ideaship: **Weak test** — the docstring says this guards against cross-device base-config contamination, "a real bug we got bitten by", but that scenario cannot occur with the current code: `base_config` is a local variable produced by `json.load` inside `generate_sonic_config`, so each call already gets an independent dict. Removing `deepcopy` here would not cause this test to fail.

The test passes but for the wrong reason — it verifies isolation that `json.load` already provides, not the isolation that `deepcopy` adds.

ideaship (commented):
**Reminder — dead `VERSIONS` guard (flagged in previous round, still open)**

The `if "VERSIONS" not in config:` guard at `config_generator.py:324–325` was called out in the last review and has not been addressed. The scaffold `setdefault` loop added by this PR guarantees `VERSIONS` is always present, so lines 324–325 are now unreachable. Only the inner guard on line 326 (`if "DATABASE" not in config["VERSIONS"]:`) is still live.

---

**New finding — `STATIC_ROUTE` wholesale replacement (correctness bug, `config_generator.py:303`)**

Line 303 does a full dict replacement:
```python
config["STATIC_ROUTE"] = {}                            # replaces the dict
config["STATIC_ROUTE"]["mgmt|0.0.0.0/0"] = {"nexthop": metalbox_ip}
```
The scaffold `setdefault` on line 197–198 preserves any `STATIC_ROUTE` entries from `config_db.json`, but line 303 then throws that away. Any custom static routes in the base config are silently dropped for every device that has an OOB IP. The fix is to drop line 303 — the dict is already present and any pre-existing entries should survive.

ideaship:
Note: the **STATIC_ROUTE production bug** section in the review body above is covered by #2238 (already open against `main`). The inline comment on the test file remains — a regression test for that fix still belongs in this PR.

PR #2238, opened 2026-04-29
sonic: Fix STATIC_ROUTE clobber in OOB management path

ideaship:
## Summary

- `generate_sonic_config()` reset `STATIC_ROUTE` to `{}` before writing the management default route, silently discarding any routes loaded from `/etc/sonic/config_db.json` on every sync.
- The reset was introduced in f4f2290 when base config loading did not exist yet — it was harmless at the time and went unnoticed when base config loading was added later.
- Fix: write the management route directly into the existing dict without resetting it first.

## Test plan

- [ ] Verify existing unit tests pass (`pytest tests/unit/tasks/conductor/sonic/`)
- [ ] Add a test that pre-populates `STATIC_ROUTE` in the base config and asserts entries survive the OOB management path (tracked as follow-up in PR #2237 review)


berendt:
Included in https://github.com/osism/python-osism/pull/2237.

Issue #2258, opened 2026-05-04
Cover Enterprise SONiC tables in the YANG model set


berendt:
## Background

The SONiC ConfigDB validator (`osism/tasks/conductor/sonic/validator.py`) generates Pydantic schemas from the YANG models shipped in `files/sonic/yang_models/`. That directory currently mirrors the **Community SONiC** YANG tree (`sonic-net/sonic-buildimage/src/sonic-yang-models/yang-models`).

When validating real configs, the validator emits warnings of the form:

```
[WARN]  xxxxxx: No YANG schema for table 'SNMP_SERVER_USER' (validation skipped)
```

for 26 ConfigDB tables. Per the design in #2253, these are reported as warnings (not errors), but they leave a real coverage gap: rows in those tables are never schema-checked.

## Why those tables are missing

The 26 tables fall into three groups:

**1. Modeled in Enterprise SONiC (Broadcom/Dell), not in Community.**
The Enterprise SONiC YANG tree (mirrored e.g. in `oceans1908/dell-yang-model` and `oceans1908/sonic-yang`) ships top-level `container` declarations for these tables in dedicated `sonic-*.yang` files that have no Community counterpart:

| ConfigDB table(s) | Enterprise YANG file |
|---|---|
| `SWITCH` (top-level) | `sonic-switch.yang` |
| `COREDUMP` | `sonic-coredump.yang` |
| `HARDWARE` | `sonic-hardware.yang` |
| `LOOPBACK` (separate from `LOOPBACK_INTERFACE`) | `sonic-loopback.yang` |
| `SNMP_SERVER`, `SNMP_SERVER_USER`, `SNMP_SERVER_VIEW`, `SNMP_SERVER_GROUP`, `SNMP_SERVER_PARAMS`, `SNMP_SERVER_TARGET`, `SNMP_SERVER_GROUP_USER`, `SNMP_SERVER_GROUP_ACCESS`, `SNMP_SERVER_GROUP_MEMBER`, `SNMP_SERVER_GROUP_PARAMS`, `SNMP_SERVER_GROUP_TARGET` | `sonic-snmp.yang` (Enterprise variant — distinct from the Community file of the same name, which only models `SNMP`/`SNMP_COMMUNITY`/`SNMP_USER`/`SNMP_AGENT_ADDRESS_CONFIG`) |
| `SSHD_COMMON` | `sonic-ssh-common.yang` |
| `NEIGH_GLOBAL` | `sonic-neighbor.yang` |
[...]

PR #2260, opened 2026-05-05
sonic/config_generator: align syslog and mgmt_port values with schema

berendt:
Three values produced by config generation did not match the SONiC schema enforced by the new Pydantic validator:

- MGMT_PORT.eth0.autoneg in the base config_db.json was "true"; the schema requires "on"|"off".
- _add_log_server_configuration uppercased the syslog protocol ("UDP"), but the schema requires lowercase "tcp"|"udp".
- _add_log_server_configuration passed the severity through unchanged, so common syslog aliases like "warning" or "critical" failed against the SONiC-accepted set (none, debug, info, notice, warn, error, crit).

Lowercase the protocol, lowercase and alias-map the severity (warning -> warn, informational -> info, err -> error, critical -> crit), fix the autoneg value in the base config, and update the existing tests to match.

AI-assisted: Claude Code

### DIFF
--- a/files/sonic/config_db.json
+++ b/files/sonic/config_db.json
@@ -335,7 +335,7 @@
     "MGMT_PORT": {
         "eth0": {
             "admin_status": "up",
-            "autoneg": "true",
+            "autoneg": "on",
             "description": "Management0",
             "mtu": "1500",
             "speed": "1000"
--- a/osism/tasks/conductor/sonic/config_generator.py
+++ b/osism/tasks/conductor/sonic/config_generator.py
@@ -2074,6 +2074,16 @@ def _add_portchannel_configuration(config, portchannel_info):
             )
 
 
+# Map common syslog severity aliases to the SONiC-accepted set.
+# SONiC schema accepts: none, debug, info, notice, warn, error, crit.
+_SYSLOG_SEVERITY_ALIASES = {
+    "informational": "info",
+    "warning": "warn",
+    "err": "error",
+    "critical": "crit",
+}
+
+
 def _add_log_server_configuration(config, device):
     """Add SYSLOG_SERVER configuration to device config.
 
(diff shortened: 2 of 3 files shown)


berendt:
Depends on https://github.com/osism/python-osism/issues/2258

osism/tasks/conductor/sonic/config_generator.py
```diff
@@ -2074,6 +2074,16 @@ def _add_portchannel_configuration(config, portchannel_info):
             )
 
 
+# Map common syslog severity aliases to the SONiC-accepted set.
+# SONiC schema accepts: none, debug, info, notice, warn, error, crit.
+_SYSLOG_SEVERITY_ALIASES = {
+    "informational": "info",
+    "warning": "warn",
+    "err": "error",
+    "critical": "crit",
```
osfrickler, line 2083: why did you omit the other three? at least debug sounds like it could be useful in certain cases

osism/tasks/conductor/sonic/config_generator.py
```diff
@@ -2085,11 +2095,14 @@ def _add_log_server_configuration(config, device):
         proto = device.config_context.get("_segment_log_server_proto", "udp")
         severity = device.config_context.get("_segment_log_server_severity", "info")
         vrf = device.config_context.get("_segment_log_server_vrf", "mgmt")
+        proto = proto.strip().lower()
```
osfrickler, line 2098: this is uppercase in the generated config, I would prefer to stick to that

files/sonic/config_db.json
```diff
@@ -335,7 +335,7 @@
     "MGMT_PORT": {
         "eth0": {
             "admin_status": "up",
-            "autoneg": "true",
+            "autoneg": "on",
```
osfrickler, line 338: again, I'd prefer to stick to how the config is generated and amend the scheme if needed

PR #2261, opened 2026-05-05
sonic/config_generator: use "up" for BGP_NEIGHBOR_AF admin_status

berendt:
_add_bgp_configurations wrote "admin_status": "true" for every BGP_NEIGHBOR_AF row it generated: ipv4_unicast, ipv6_unicast and l2vpn_evpn keys for both interfaces and port channels, plus the VLAN peer ipv4_unicast row. The schema enforced by the new Pydantic validator (BgpNeighborAfListRow.admin_status) is Literal["up", "down"], so validation rejected every generated row with "Input should be 'up' or 'down'".

Replace the literal "true" with "up" at all seven call sites. Same fix pattern as #2260, which aligned other generated values with the schema.

AI-assisted: Claude Code

### DIFF
--- a/osism/tasks/conductor/sonic/config_generator.py
+++ b/osism/tasks/conductor/sonic/config_generator.py
@@ -38,7 +38,12 @@ from .connections import (
     get_connected_interface_ipv4_address,
 )
 from .cache import get_cached_device_interfaces
-from .constants import BGP_AF_L2VPN_EVPN_TAG, DEFAULT_SONIC_ROLES
+from .constants import (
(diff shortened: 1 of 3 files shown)


berendt:
Depends on https://github.com/osism/python-osism/issues/2258

osfrickler (commented):
I'm not convinced that this is a good idea, operationally it is better to stick to the values that a `config save` on the running switch generates, so that it is easy to spot actual config changes. If the scheme we use for verification doesn't match this, then IMO the scheme is wrong, not the generated config.

PR #2266, opened 2026-05-06
sonic: add --refresh-host-key option to refresh SSH host keys after r…

berendt:
…edeployment

Adds an optional --refresh-host-key flag to all SSH-using sonic commands (load, backup, ztp, reload, reboot, reset, show). When set, existing known_hosts entries for the target host are removed before connecting, allowing the new host key to be accepted after a switch has been redeployed and the host key changed.

AI-assisted: Claude Code


osfrickler (approved):
the code looks reasonable to me. I assume the unit tests are fine, but it would take a lot more time to verify that they really cover all reasonable scenarios, so I've skipped that part

PR #2275, opened 2026-05-13
vault view: show plain content for non-encrypted files

### LINKED ISSUE #1362: [bug] osism vault view fails with an error on non-encrypted files
### OSISM release version

latest

### What's the problem?

If one tries to view a file that is not encrypted or has been decrypted before, `osism vault view` exits with an error:

```
[ERROR]: Input is not vault encrypted data. for /opt/configuration/environments/openstack/secure.yml
```

I think it would be more useful if it would still show the file, maybe adding a warning that the file didn't need to be decrypted.

### References to existing reports

_No response_

### Severity

low

### Urgency

low

berendt:
Previously `osism vault view` invoked `ansible-vault view` directly,
which exits with an error when the file is not vault-encrypted. Now
the command reads the file once, checks for the `$ANSIBLE_VAULT`
header, and either delegates to `ansible-vault view` or prints the
plain content with a warning.

Missing files and permission errors are surfaced with a clear log
message and a non-zero exit code instead of silently falling through
to the plain-content path.

Adds unit tests for ``View.take_action`` covering the encrypted,
plain-file, missing-file, and permission-denied code paths, plus
parametrized coverage for multiple `$ANSIBLE_VAULT` header variants.

Closes osism/issues#1362

AI-assisted: Claude Code

### DIFF
--- a/osism/commands/vault.py
+++ b/osism/commands/vault.py
@@ -69,7 +69,26 @@ class View(Command):
         path = parsed_args.path
         if not os.path.isabs(path):
             path = os.path.join("/opt/configuration", path)
-        subprocess.call(f"/usr/local/bin/ansible-vault view {path}", shell=True)
+
+        try:
+            with open(path, "rb") as fp:
+                content = fp.read()
+        except FileNotFoundError:
+            logger.error(f"File not found: {path}")
+            return 1
+        except PermissionError:
+            logger.error(f"Permission denied reading file: {path}")
+            return 1
+        except OSError as exc:
+            logger.error(f"Cannot read file {path}: {exc}")
+            return 1
+
+        if content.startswith(b"$ANSIBLE_VAULT"):
+            subprocess.call(f"/usr/local/bin/ansible-vault view {path}", shell=True)
+            return
+
+        logger.warning(f"File is not vault-encrypted, showing plain content: {path}")
+        sys.stdout.write(content.decode("utf-8", errors="replace"))
 
 
 class Decrypt(Command):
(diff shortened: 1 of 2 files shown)


osism/commands/vault.py
```diff
@@ -14,6 +14,12 @@
 from osism import utils
 
 
+def _is_vault_encrypted(path):
```
osfrickler: what is this function good for? it doesn't seem to get used in the code below

osfrickler (changes requested):

osism/commands/vault.py
```diff
@@ -69,7 +69,26 @@ def take_action(self, parsed_args):
...
+        except PermissionError:
+            logger.error(f"Permission denied reading file: {path}")
+            return 1
+        except OSError as exc:
+            logger.error(f"Cannot read file {path}: {exc}")
+            return 1
+
+        if content.startswith(b"$ANSIBLE_VAULT"):
+            subprocess.call(f"/usr/local/bin/ansible-vault view {path}", shell=True)
+            return
```
ideaship:
```suggestion
            return subprocess.call(f"/usr/local/bin/ansible-vault view {path}", shell=True)
```

If we return 1 for all kinds of errors, then we should also pass on any error values returned by ansible-vault.

ideaship (changes requested):
The PR description talks about _is_vault_encrypted which does no longer exist in the commit.

osism/commands/vault.py
berendt: Good point — fixed in 7392ec1. The `ansible-vault view` exit code is now returned so vault errors are propagated consistently with the other error paths.

ideaship (commented):
As far as I am concerned, this can be merged. We should fix the shell=True pattern pointed out by sourceryAI in this PR or in a follow-up PR (there's a another instance of it in Decrypt.take_action).

PR #2279, opened 2026-05-18
Redistribute static + IPv6 routes

fzakfeld:


osism/tasks/conductor/sonic/config_generator.py
```diff
@@ -1979,9 +1979,15 @@ def _add_vrf_configuration(config, vrf_info, netbox_interfaces):
             # Add ROUTE_REDISTRIBUTE for VRF
             if "ROUTE_REDISTRIBUTE" not in config:
                 config["ROUTE_REDISTRIBUTE"] = {}
-            route_redistribute_key = f"{vrf_name}|connected|bgp|ipv4"
-            config["ROUTE_REDISTRIBUTE"][route_redistribute_key] = {}
-            logger.info(f"Added ROUTE_REDISTRIBUTE {route_redistribute_key}")
+            for proto, af in [
+                ("connected", "ipv4"),
+                ("connected", "ipv6"),
```
ideaship, line 1984: **Bug: IPv6 redistribution entries have no corresponding IPv6 unicast AF**

This loop emits `{vrf}|connected|bgp|ipv6` and `{vrf}|static|bgp|ipv6` into `ROUTE_REDISTRIBUTE`, but the `BGP_GLOBALS_AF` block above (lines 1952–1970) only creates `{vrf}|ipv4_unicast` and `{vrf}|l2vpn_evpn` for VNI-backed VRFs — it never creates `{vrf}|ipv6_unicast`. The IPv6 redistribution entries are therefore structurally orphaned for tenant VRFs: there is no IPv6 unicast AF to redistribute into. Either add a `{vrf}|ipv6_unicast` entry to `BGP_GLOBALS_AF` alongside the existing `ipv4_unicast` entry, or explain why the AF is not needed here.

osism/tasks/conductor/sonic/config_generator.py
```diff
@@ -1979,9 +1979,15 @@ def _add_vrf_configuration(config, vrf_info, netbox_interfaces):
             # Add ROUTE_REDISTRIBUTE for VRF
```
ideaship, line 1979: **Missing test coverage for `ROUTE_REDISTRIBUTE`**

There are currently no tests for emitted `ROUTE_REDISTRIBUTE` entries — `_add_vrf_configuration()` is only seen in the orchestrator test as a patch target, with no assertions on its output. This PR changes the emitted entry count from one to four, which is a behavioral change worth pinning. Please add a unit test (alongside the existing BGP globals tests in `_config_generator_helpers.py`) that calls `_add_vrf_configuration()` with a VNI-backed VRF fixture and asserts that all four expected keys are present in the resulting `ROUTE_REDISTRIBUTE` table.

osism/tasks/conductor/sonic/config_generator.py
```diff
@@ -1979,9 +1979,15 @@ def _add_vrf_configuration(config, vrf_info, netbox_interfaces):
...
             if "ROUTE_REDISTRIBUTE" not in config:
                 config["ROUTE_REDISTRIBUTE"] = {}
-            route_redistribute_key = f"{vrf_name}|connected|bgp|ipv4"
-            config["ROUTE_REDISTRIBUTE"][route_redistribute_key] = {}
-            logger.info(f"Added ROUTE_REDISTRIBUTE {route_redistribute_key}")
+            for proto, af in [
+                ("connected", "ipv4"),
+                ("connected", "ipv6"),
+                ("static", "ipv4"),
```
ideaship, line 1985: **Please confirm: all static routes from every VNI-backed VRF redistributed into BGP with no filtering**

With the addition of `("static", "ipv4")` and `("static", "ipv6")`, this change redistributes every static route from every VNI-backed VRF into BGP unconditionally — no route-map, no prefix-list, no opt-out per VRF. Static routes often include defaults, management escape routes, or other entries that should not leave the local VRF. Is it intentional that there is no filtering mechanism here? If selective export is expected in the future, this is worth noting now so the interface can be designed for it.

ideaship (changes requested):
Note: This comment has been resolved by https://github.com/osism/python-osism/pull/2297

~~Approving this change would be under the following assumption: entries whose key and value are fully derived from NetBox plus hardcoded SONiC policy are fully owned by the generator and should always be overwritten on regen.~~

~~Under that rule, the unconditional assignment at line 1989 (config["ROUTE_REDISTRIBUTE"][key] = {}) is correct. If the intent is that operators can layer route-maps onto generated VRFs, that is a feature request that requires changing the generator's ownership model — it is not a bug in the current code.~~

~~The reason for flagging this explicitly: another reviewer may dispute the assumption. generate_sonic_config() deep-copies /etc/sonic/config_db.json, and the PR expands the set of overwritten keys from one (connected|bgp|ipv4) to four, including the static variants where operator policy is most plausible. If the rule above is intentional and documented somewhere, a pointer here would close the question.~~

osism/tasks/conductor/sonic/config_generator.py
fzakfeld: I think it would be best if we wait for https://github.com/osism/python-osism/pull/2173 which will introduce a way of installing default routes per VRF with `default_route_ipv4` and `default_route_ipv6`. I only need `redistribute static` in VRFs where this is set anyways, so I think it makes total sense to change that.

Alternative would be a `default-originate` on the peering sessions to the servers, but this ignores the actual route's presence and would need to be passed to the other switches as well. I much prefer a `redistribute static`

fzakfeld:
Closing this as we don't need it anymore

PR #2298, opened 2026-05-20
Add unit tests for osism/utils task output and helpers

### LINKED ISSUE #2229: Unit tests for osism/utils/__init__.py — task output, revoke, ansible helpers
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul infrastructure). Part of **Tier 4** (#2199). Companion to the connection-init issue: covers task-output streaming, task revocation, the ansible-vault password helper, the ansible-facts freshness check, and the `first` iterator helper from `osism/utils/__init__.py`.

## Scope

Add `tests/unit/utils/test_init_task_output.py` covering the helpers below in `osism/utils/__init__.py`.

### `first(iterable, condition=lambda x: True)` — `__init__.py:348`

Pure function (already has doctests).

- `first((1, 2, 3), condition=lambda x: x % 2 == 0)` → `2`
- `first(range(3, 100))` → `3` (default condition)
- `first(())` → raises `StopIteration`
- No item satisfies condition → raises `StopIteration`
- Generator input (consumed lazily) → returns the first match without exhausting downstream

### `fetch_task_output(task_id, timeout=…, enable_play_recap=False)` — `__init__.py:371`

Patch `_init_redis` to return a `MagicMock` redis client. Capture stdout via `capsys`.

- One stdout message + `rc` message + `quit` action → prints stdout, returns the rc value (as `int`)
- Default rc when no `rc` message is sent → returns `0`
- `enable_play_recap=True` and stdout contains `"PLAY RECAP"` → log message about completion is emitted
- `enable_play_recap=False` and stdout contains `"PLAY RECAP"` → no extra log
- Each delivered message → `xdel(task_id, last_id)` called
- After processing one batch → `last_id` updated so the next `xread` uses it (`{str(task_id): last_id}`)
- `xread` returns `None` for the entire timeout window → `TimeoutError` raised
- `xread` returning data resets the stop-time deadline
- `quit` action → `r.close()` called and rc returned immediately (`return rc`)
- Custom timeout from env (`OSISM_TASK_TIMEOUT`) propagates (the module reads it at import; one happy-path test that the function honours an explicit `timeout=…` kwarg is enough)

### `push_task_output(task_id, line)` — `__init__.py:407`

Patch `_init_redis`.

- Calls `xadd(task_id, {"type": "stdout", "content": line})` exactly once

### `finish_task_output(task_id, rc=None)` — `__init__.py:411`

- `rc=None` → only the `quit` action is published (`{"type": "action", "content": "quit"}`)
- `rc=0` → only `quit` is published (the `if rc:` check is intentionally truthy)
- `rc=1` → both rc message and quit action published, in that order
- Verify the order via `xadd.call_args_list`

### `revoke_task(task_id)` — `__init__.py:418`

Patch `celery.Celery` and `osism.tasks.Config`.

- Happy path → instantiates `Celery("task")`, calls `app.config_from_object(Config)`, then `app.control.revoke(task_id, terminate=True)`, returns `True`
- `Celery(...)` raises → returns `False`, error logged
- `app.control.revoke` raises → returns `False`, error logged

### `get_ansible_vault_password()` — `__init__.py:318`

Patch `builtins.open` (`mock_open(read_data="<fernet-key>")`), `cryptography.fernet.Fernet`, and `_init_redis`.

- Key file readable, redis returns encrypted bytes, `Fernet().decrypt(...)` returns valid password → returns the decoded password (stripped? — the production code does **not** strip, only checks `password.strip() == ""`; verify the actual return value is the raw decoded text)
- Redis returns `None` for key `"ansible_vault_password"` → raises `ValueError("Ansible vault password is not set in Redis")`, error logged before re-raise
- Decryption returns empty string → `ValueError("...empty or contains only whitespace")`
- Decryption returns whitespace-only → same `ValueError`
- `open(keyfile)` raises `FileNotFoundError` → propagates after logging (test by patching `open` to raise)
- `Fernet(key).decrypt(...)` raises → propagates after logging

### `check_ansible_facts(max_age=None)` — `__init__.py:560`

Patch `_init_redis` and `osism.utils.settings.FACTS_MAX_AGE`. Use frozen-time helpers (or pass `max_age=...` explicitly).

- `r.scan` raises → warning logged, function returns early without further work
- No keys found → warning `"No Ansible facts found..."`, returns
- One stale host (`epoch` older than `max_age`) → warning logged listing the stale host with age in seconds
- One fresh host → no warning logged
- Mix of fresh and stale → only stale listed
- Hostname extraction: key `b"ansible_facts<hostname>"` → strips the `"ansible_facts"` prefix
- Bytes vs. str keys both handled
- `r.get(key)` returns `None` → host skipped (continue)
- `r.get` returns malformed JSON → caught (`JSONDecodeError`), debug log, host skipped
- `epoch` missing → debug log "facts missing ansible_date_time.epoch", host skipped
- `epoch` non-numeric (e.g. string `"foo"`) → caught (`ValueError`/`TypeError`), debug log, host skipped
- `r.scan` paginates correctly: first call returns `(cursor!=0, [k1])`, second `(0, [k2])` → both keys processed
- `max_age=None` → uses `settings.FACTS_MAX_AGE`
- Explicit `max_age=10` → overrides settings

## Mocking hints

- For `fetch_task_output`, build the `xread` return value as the production code expects:
  ```python
  data = [(b"task-id", [(b"123-0", {b"type": b"stdout", b"content": b"hello"})])]
  ```
  Use `side_effect` on the mock to return a sequence of payloads (one per loop iteration), ending with `None` only when you want the timeout path.
- For `check_ansible_facts`, prefer providing `max_age=10` and crafting `epoch` values relative to `time.time()` rather than freezing time.
- `_init_redis` is the single dependency most of these helpers share — patch it once per test or via a fixture returning a `MagicMock`.

## Definition of Done

- [ ] `tests/unit/utils/test_init_task_output.py` created
- [ ] `pytest --cov=osism.utils` for the targeted helpers ≥ 90 %
- [ ] `pipenv run pytest tests/unit/utils/test_init_task_output.py` passes locally

## Dependencies

berendt:
Covers fetch_task_output / push_task_output / finish_task_output, revoke_task, get_ansible_vault_password, check_ansible_facts and the first iterator helper from osism/utils/__init__.py. Companion to the connection-init suite added in

Closes #2229.

AI-assisted: Claude Code

### DIFF
--- a/tests/unit/utils/test_init_task_output.py
+++ b/tests/unit/utils/test_init_task_output.py
@@ -9,6 +9,7 @@ dependency most helpers share — it is patched per-test to return a
 ``MagicMock`` redis client.
 """
 
+import importlib
 from unittest.mock import call, mock_open
 
 import pytest
@@ -181,6 +182,8 @@ def test_fetch_task_output_timeout_when_xread_returns_none(mocker):
     with pytest.raises(TimeoutError):
         utils_pkg.fetch_task_output("task-1", timeout=10)
 
+    mock_r.close.assert_called_once_with()
+
 
 def test_fetch_task_output_data_resets_deadline(mocker):
     """When ``xread`` returns data the loop must reset its deadline.
@@ -239,6 +242,20 @@ def test_fetch_task_output_honours_explicit_timeout_kwarg(mocker):
     assert kwargs.get("block") == 42 * 1000
 
 
+def test_fetch_task_output_default_timeout_is_int_when_env_var_set(monkeypatch):
+    """``OSISM_TASK_TIMEOUT`` is read at import time; without an ``int`` cast
+    the default would be a ``str`` and ``time.time() + timeout`` would crash."""
+    monkeypatch.setenv("OSISM_TASK_TIMEOUT", "60")
+    try:
+        reloaded = importlib.reload(utils_pkg)
+        default_timeout = reloaded.fetch_task_output.__defaults__[0]
+        assert default_timeout == 60
+        assert isinstance(default_timeout, int)
+    finally:
+        monkeypatch.delenv("OSISM_TASK_TIMEOUT", raising=False)
+        importlib.reload(utils_pkg)
+
+
 # ---------------------------------------------------------------------------
 # push_task_output
 # ---------------------------------------------------------------------------
@@ -269,14 +286,17 @@ def test_finish_task_output_rc_none_publishes_only_quit(mocker):
     mock_r.xadd.assert_called_once_with("task-1", {"type": "action", "content": "quit"})
 
 
-def test_finish_task_output_rc_zero_publishes_only_quit(mocker):
-    """``if rc:`` is intentionally truthy — rc=0 must be treated like rc=None."""
+def test_finish_task_output_rc_zero_publishes_rc_then_quit(mocker):
+    """A successful exit (rc=0) must publish an rc message; only rc=None skips it."""
     mock_r = mocker.MagicMock()
     mocker.patch("osism.utils._init_redis", return_value=mock_r)
 
     utils_pkg.finish_task_output("task-1", rc=0)
 
-    mock_r.xadd.assert_called_once_with("task-1", {"type": "action", "content": "quit"})
+    assert mock_r.xadd.call_args_list == [
+        call("task-1", {"type": "rc", "content": 0}),
+        call("task-1", {"type": "action", "content": "quit"}),
+    ]
 
 
 def test_finish_task_output_nonzero_rc_publishes_rc_then_quit(mocker):
--- a/tests/unit/utils/test_init_task_output.py
+++ b/tests/unit/utils/test_init_task_output.py
@@ -1,0 +1,677 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for task-output streaming, task revocation, the ansible-vault
+password helper, the ansible-facts freshness check, and the ``first`` iterator
+helper from :mod:`osism.utils`.
+
+Companion to ``test_init_connections.py``. ``_init_redis`` is the single
+dependency most helpers share — it is patched per-test to return a
+``MagicMock`` redis client.
+"""
+
+from unittest.mock import call, mock_open
+
+import pytest
+
+import osism.utils as utils_pkg
+
+# ---------------------------------------------------------------------------
+# first
+# ---------------------------------------------------------------------------
+
+
+def test_first_returns_first_matching_item():
+    assert utils_pkg.first((1, 2, 3), condition=lambda x: x % 2 == 0) == 2
+
+
+def test_first_default_condition_returns_first_item():
+    assert utils_pkg.first(range(3, 100)) == 3
+
+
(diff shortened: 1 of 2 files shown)


tests/unit/utils/test_init_task_output.py
```diff
@@ -0,0 +1,677 @@
...
+    mock_r.xread.return_value = None
+    mocker.patch("osism.utils._init_redis", return_value=mock_r)
+    # time.time() is called once to compute ``stoptime`` and once per loop
+    # iteration. Returning a value past the deadline on the second poll exits
+    # the loop and raises ``TimeoutError``.
+    time_values = iter([0.0, 0.0, 100.0])
+    mocker.patch("osism.utils.time.time", side_effect=lambda: next(time_values))
+
+    with pytest.raises(TimeoutError):
+        utils_pkg.fetch_task_output("task-1", timeout=10)
```
ideaship, line 183: Design question: the normal exit path calls `r.close()` before returning; this path raises `TimeoutError` without closing `r`. Is that intentional? If closing on timeout is correct, this test should add `mock_r.close.assert_called_once()` to guard against it being dropped.

tests/unit/utils/test_init_task_output.py
```diff
@@ -0,0 +1,677 @@
...
+def test_finish_task_output_rc_none_publishes_only_quit(mocker):
+    mock_r = mocker.MagicMock()
+    mocker.patch("osism.utils._init_redis", return_value=mock_r)
+
+    utils_pkg.finish_task_output("task-1", rc=None)
+
+    mock_r.xadd.assert_called_once_with("task-1", {"type": "action", "content": "quit"})
+
+
+def test_finish_task_output_rc_zero_publishes_only_quit(mocker):
```
ideaship: \`if rc:\` silently drops \`rc=0\`, making a successful subprocess exit indistinguishable from "no rc message received" — \`fetch_task_output\` defaults \`rc = 0\` and returns it either way. The pre-refactor code (commit \`7257aa4\`) published the rc message unconditionally; the \`if rc:\` guard was introduced in that refactor without explanation, suggesting an oversight rather than intent. Consider \`if rc is not None:\`.

The docstring here ("intentionally truthy") appears to be a post-hoc interpretation of the existing code rather than documented design intent, and would block a correct fix.

ideaship (commented):
**Bug in production code not covered by new tests — `osism/utils/__init__.py:372`**

`timeout=os.environ.get("OSISM_TASK_TIMEOUT", 300)` is evaluated at import time. When `OSISM_TASK_TIMEOUT` is set in the environment the default becomes a `str`, and `time.time() + timeout` crashes with `TypeError`. The call site in `wait.py` passes no explicit timeout, so this is reachable in production.

Fix: `timeout=int(os.environ.get("OSISM_TASK_TIMEOUT", 300))`.

Note: all new tests pass an explicit `timeout=` kwarg, which is why this went undetected. No test exercises the default-parameter path.

ideaship (approved):

PR #2313, opened 2026-06-01
Return non-zero exit from get on query failure

ideaship:
## What

`osism get hostvars` and `osism get hosts` logged an error but then returned `None` when the underlying `ansible-inventory` subprocess failed. cliff turns a `take_action` return value of `None` into exit code 0 (`return_code = self.take_action(parsed_args) or 0`), so these commands exited **successfully** even when the inventory query could not be run at all — making them unsafe in `set -e` scripts and `&&` chains.

Both commands now `return 1` on `subprocess.CalledProcessError`, i.e. when the query itself could not be executed.

## Missing host

With the pinned `ansible-core` (2.18.9 — `release` repo `latest/base.yml`), `ansible-inventory --host <unknown>` does **not** return an empty result. It refuses the query with `ERROR! You must pass a single valid host to --host parameter` and exits 5. So `osism get hostvars <unknown-host>` now exits 1, matching its existing `Host not found in inventory.` error message.

## What stays exit 0 (empty results)

Empty results are deliberately left as success — a query that ran fine but had nothing to show:

- a host that exists but has no host vars (`ansible-inventory` prints `{}`, exits 0)
- a requested variable/fact absent from an otherwise successfully retrieved data set
- an inventory that lists no matching hosts

## A note on the convention

There is **no documented exit-code contract** for the `osism` CLI — the behaviour is emergent and inconsistent across commands. This PR does not attempt to define or enforce one. It only tries to make `osism get` match what looks like the prevailing convention elsewhere in the CLI: a non-zero exit signals an *operational failure* (the query could not be run), while an empty result set is treated as a valid, successful answer.

## Tests

Adds unit tests for both commands covering the failure path (raised `CalledProcessError` → exit 1) and the empty-result path (successful query with nothing to show → exit 0).

### DIFF
--- a/osism/commands/get.py
+++ b/osism/commands/get.py
@@ -139,7 +139,7 @@ class Hostvars(Command):
             )
         except subprocess.CalledProcessError:
             logger.error(f"Host {host} not found in inventory.")
-            return
+            return 1
 
         data = json.loads(result)
         table = []
@@ -249,7 +249,7 @@ class Hosts(Command):
             )
         except subprocess.CalledProcessError:
             logger.error("Error loading inventory.")
-            return
+            return 1
 
         data = json.loads(result)
         table = []
--- a/tests/unit/commands/test_get.py
+++ b/tests/unit/commands/test_get.py
@@ -1,0 +1,91 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for the ``osism get`` commands.
+
+These focus on the exit-code contract: a command must return a non-zero exit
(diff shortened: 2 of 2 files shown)


ideaship:
Follow-up: #2314 catalogs other `osism` subcommands with the same exit-0-on-failure pattern this PR fixes in `osism get` (timeouts, failed inventory queries, and not-found lookups across `reconciler`, `validate`, `wait`, `report`, `baremetal`, `compute`, `server`, `volume`, `netbox`).

PR #2328, opened 2026-06-03
sonic: Add support for BGP IPv6 neighbors on VLAN

### LINKED ISSUE #2312: sonic: Add support for BGP IPv6 neighbors on VLAN
The current code only checks for IPv4 addresses, we also need to support IPv6 here. 

https://github.com/osism/python-osism/blob/58ac59e912e0a8bffedba7805cc0aaed7076f06a/osism/tasks/conductor/sonic/config_generator.py#L1326-L1331

berendt:
## Summary

The VLAN BGP neighbor generation in `_add_bgp_configurations` only looked up the **IPv4** address of the connected peer interface, so IPv6 peers never received a `BGP_NEIGHBOR` entry — the case described in #2312.

This change adds IPv6 support for BGP neighbors on VLAN interfaces (SVIs).

## Changes

- **`connections.py`**: Factored the existing IPv4 lookup into a shared, IP-version-parameterized helper `_get_connected_interface_ip_address(..., ip_version)` (covers both direct addresses and FHRP VIPs). Added two thin wrappers:
  - `get_connected_interface_ipv4_address` (behaviour unchanged)
  - `get_connected_interface_ipv6_address` (new)
  - plus `_address_matches_ip_version` for the family check (`.` vs `:`).
- **`config_generator.py`**: The VLAN section now resolves both the IPv4 and IPv6 peer address. A dual-stack peer yields one numbered neighbor per address family, each with the matching `BGP_NEIGHBOR_AF` (`ipv4_unicast` / `ipv6_unicast`). Per-address dedup is preserved; `v6only=false` is set for both families in the default VRF (numbered global addresses, consistent with the prior IPv4 behaviour).

## Tests

- Extended the `patch_bgp` fixture with `peer_ipv6`.
- New cases: IPv6-only peer, dual-stack (both neighbors created), non-default VRF (no `v6only`), and IPv6 dedup.
- Direct unit tests in `test_connections.py` for the version detection and the IPv6 lookup (direct + FHRP VIP).

Closes #2312

### DIFF
--- a/osism/tasks/conductor/sonic/config_generator.py
+++ b/osism/tasks/conductor/sonic/config_generator.py
@@ -36,9 +36,16 @@ from .connections import (
     get_connected_interfaces,
     get_connected_device_for_sonic_interface,
     get_connected_interface_ipv4_address,
+    get_connected_interface_ip_addresses,
+    is_numbered_neighbor_address,
 )
 from .cache import get_cached_device_interfaces
-from .constants import BGP_AF_L2VPN_EVPN_TAG, DEFAULT_SONIC_ROLES
+from .constants import (
+    BGP_AF_IPV4_UNICAST,
+    BGP_AF_IPV6_UNICAST,
+    BGP_AF_L2VPN_EVPN_TAG,
+    DEFAULT_SONIC_ROLES,
+)
 
 # Global cache for metalbox IPs per device to avoid duplicate lookups
 _metalbox_ip_cache: dict[int, Optional[str]] = {}
@@ -1090,12 +1097,12 @@ def _add_bgp_configurations(
                 neighbor_id = port_name
                 vrf_name = get_vrf_for_interface(port_name)
 
-                ipv4_key = f"{vrf_name}|{neighbor_id}|ipv4_unicast"
+                ipv4_key = f"{vrf_name}|{neighbor_id}|{BGP_AF_IPV4_UNICAST}"
                 config["BGP_NEIGHBOR_AF"][ipv4_key] = {"admin_status": "true"}
 
                 # Only add ipv6_unicast if v6only would be true (no transfer role IPv4)
                 if not has_transfer_ipv4:
-                    ipv6_key = f"{vrf_name}|{neighbor_id}|ipv6_unicast"
+                    ipv6_key = f"{vrf_name}|{neighbor_id}|{BGP_AF_IPV6_UNICAST}"
                     config["BGP_NEIGHBOR_AF"][ipv6_key] = {"admin_status": "true"}
                     logger.debug(
                         f"Added BGP_NEIGHBOR_AF with ipv4_unicast and ipv6_unicast for interface {port_name} (no direct IPv4)"
@@ -1156,8 +1163,8 @@ def _add_bgp_configurations(
         neighbor_id = pc_name
         vrf_name = get_vrf_for_interface(pc_name)
 
-        ipv4_key = f"{vrf_name}|{neighbor_id}|ipv4_unicast"
-        ipv6_key = f"{vrf_name}|{neighbor_id}|ipv6_unicast"
+        ipv4_key = f"{vrf_name}|{neighbor_id}|{BGP_AF_IPV4_UNICAST}"
+        ipv6_key = f"{vrf_name}|{neighbor_id}|{BGP_AF_IPV6_UNICAST}"
         config["BGP_NEIGHBOR_AF"][ipv4_key] = {"admin_status": "true"}
         config["BGP_NEIGHBOR_AF"][ipv6_key] = {"admin_status": "true"}
 
@@ -1346,6 +1353,25 @@ def _add_bgp_configurations(
             if not addresses:
                 continue
 
+            # Address families the local SVI can source. A numbered BGP
+            # session only comes up when the SVI itself carries a routable
+            # address in that family, so an IPv4-only SVI must not emit an
+            # IPv6 numbered neighbor (and vice versa); link-local-only
+            # addresses do not count (see is_numbered_neighbor_address).
+            local_address_families = {
+                version
+                for version in (4, 6)
+                if any(
+                    is_numbered_neighbor_address(addr, version) for addr in addresses
+                )
+            }
+            if not local_address_families:
+                logger.debug(
+                    f"VLAN {vid} SVI has no routable IP address, "
+                    f"skipping BGP configuration"
+                )
+                continue
+
             # Find untagged member interfaces for this VLAN
             # Only untagged members are relevant for VLAN BGP neighbors
             if vid not in vlan_info["vlan_members"]:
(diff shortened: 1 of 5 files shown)


osism/tasks/conductor/sonic/config_generator.py
```diff
@@ -1387,35 +1393,53 @@ def get_vrf_for_interface(interface_name):
...
                         device, sonic_iface_name, netbox
                     )
+                    if peer_ipv4:
+                        peer_ips.append((peer_ipv4, BGP_AF_IPV4_UNICAST))
+
+                    peer_ipv6 = get_connected_interface_ipv6_address(
+                        device, sonic_iface_name, netbox
+                    )
+                    if peer_ipv6:
+                        peer_ips.append((peer_ipv6, BGP_AF_IPV6_UNICAST))
```
ideaship: **Finding 1 — VLAN BGP neighbor address-families are not gated by the local SVI's address families** (correctness, data-dependent)

The SVI's own address families aren't consulted before emitting peer neighbors — both `peer_ipv4` and `peer_ipv6` are looked up regardless of what `addresses` (line 1351) contains. An IPv4-only SVI here will still emit an IPv6 numbered neighbor (and vice versa), and a numbered session can't come up without a connected route in that AF on the SVI. This gap is pre-existing — the IPv4-only path never gated on the SVI having an IPv4 address either — so it's a widening to IPv6 rather than a regression. But the new `test_untagged_member_with_peer_ipv6` now codifies the bad case, which makes this PR the natural place to close it. Suggest parsing `vlan_interface_data["addresses"]` into the set of local AFs and only appending a peer of a family the SVI can source. Note `test_untagged_member_with_peer_ipv6` (test file line 370) currently asserts the IPv6 neighbor on the IPv4-only default `_vlan_info` — that test should set an IPv6 SVI address once the gate is in, otherwise it's locking in the unsourceable case.

osism/tasks/conductor/sonic/connections.py
```diff
@@ -423,18 +425,32 @@ def get_device_bgp_neighbors_via_loopback(
...
-def get_connected_interface_ipv4_address(device, sonic_port_name, netbox):
+def _address_matches_ip_version(address, ip_version):
+    """Return True if ``address`` is a valid IP address of ``ip_version``.
+
+    ``address`` is the bare address (network mask already stripped). Parsing
+    with :mod:`ipaddress` is more robust than inspecting the string for ``.``
+    or ``:``; an unparseable address simply does not match any version.
+    """
+    try:
+        return ipaddress.ip_address(address).version == ip_version
```
ideaship: **Finding 2 — Link-local IPv6 can be emitted as a numbered BGP neighbor** (correctness, data-dependent)

`_address_matches_ip_version` accepts link-local addresses (`fe80::/10`, and on the v4 side `169.254.0.0/16`). When the connected interface carries a link-local IPv6, this returns it and the VLAN path emits it as a numbered `BGP_NEIGHBOR[vrf|fe80::…]`, which is invalid — link-local neighbors need the unnumbered/`v6only` form. The IPv4 side has had the same latent gap (`169.254.0.0/16`), so this isn't a regression — but link-local is far more likely to actually appear on an IPv6 interface in practice, which is why it's worth filtering now. Suggest excluding `.is_link_local` (and arguably `.is_loopback`/`.is_multicast`) here, or filtering at the selection site, so only usable global addresses are returned.

osism/tasks/conductor/sonic/config_generator.py
```diff
@@ -1387,35 +1393,53 @@ def get_vrf_for_interface(interface_name):
...
+                # neighbor per address family.
+                peer_ips = []
                 if netbox:
                     peer_ipv4 = get_connected_interface_ipv4_address(
                         device, sonic_iface_name, netbox
                     )
+                    if peer_ipv4:
+                        peer_ips.append((peer_ipv4, BGP_AF_IPV4_UNICAST))
+
+                    peer_ipv6 = get_connected_interface_ipv6_address(
```
ideaship: **Finding 3 — Peer IP lookup doubles NetBox I/O per untagged member** (cleanup / efficiency)

These two helpers each independently re-fetch the connected interface (`connections.py:458`), re-filter its IP addresses (line 485), and re-query FHRP assignments (line 506) — so every untagged member pays two full round-trips where one would do. Consider a combined `_get_connected_interface_ip_addresses()` that fetches interface/IPs/FHRP once and returns both `(ipv4, ipv6)`; the two thin wrappers can remain for existing callers.

osism/tasks/conductor/sonic/connections.py
```diff
@@ -423,18 +425,32 @@ def get_device_bgp_neighbors_via_loopback(
     return bgp_neighbors
 
 
-def get_connected_interface_ipv4_address(device, sonic_port_name, netbox):
+def _address_matches_ip_version(address, ip_version):
```
ideaship: **Finding 4 — `_address_matches_ip_version` duplicates an existing version-detection idiom** (cleanup)

Note the project also detects IP version inline via `ipaddress.ip_interface(addr).version` (`config_generator.py:925-926`, `1839-1842`). Not worth a refactor on its own, but if you add link-local filtering here (see other comment), consider promoting this to a single shared helper so the two idioms can't drift on that rule.

osism/tasks/conductor/sonic/config_generator.py
```diff
@@ -1387,35 +1393,53 @@ def get_vrf_for_interface(interface_name):
                     )
                     continue
 
-                # Get peer IP address using the existing FHRP VIP detection logic
-                peer_ipv4 = None
+                # Get peer IP addresses (IPv4 and IPv6) using the existing
+                # FHRP VIP detection logic. A dual-stack peer yields one
+                # neighbor per address family.
+                peer_ips = []
```
ideaship: **Finding 5 — Confusable variable names in the VLAN BGP block** (cleanup)

`peer_ips` (list of `(addr, family)` tuples) and `peer_ips_found` (set of addr strings, line 1378) live in the same scope and read as siblings, but hold different shapes. Consider renaming this one to `peer_ip_afs` / `peer_neighbors` so a future `x in peer_ips` written by analogy with the dedup set can't silently compare against tuples.

osism/tasks/conductor/sonic/connections.py
```diff
@@ -465,25 +481,25 @@ def get_connected_interface_ipv4_address(device, sonic_port_name, netbox):
         if not connected_interface:
             return None
 
-        # First, try to get direct IPv4 addresses assigned to the connected interface
+        # First, try to get direct IP addresses assigned to the connected interface
         ip_addresses = netbox.ipam.ip_addresses.filter(
             assigned_object_id=connected_interface.id,
         )
 
         for ip_address in ip_addresses:
```
ideaship, line 508: **Finding 6 — IPv6 FHRP "direct address and VIP" precedence is untested** (missing/weak test)

The direct-address-before-VIP precedence here is inherited from the IPv4 helper, so it's intentional — but no test covers the "connected interface has both a direct IPv6 *and* an FHRP VIP" case (the VIP test only reaches the VIP path because there's no direct IPv6). Worth a one-line test asserting the direct address wins when both exist, plus a comment noting that's deliberate — so a future reader doesn't "fix" it toward VIP peering.

ideaship (changes requested):

ideaship (approved):

Issue #2199, opened 2026-04-24
Meta: Unit test coverage for osism/


berendt:
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul CI). Once the test infrastructure has landed, actual unit test coverage has to be added module by module. This meta issue tracks the overall rollout plan so nothing gets lost and links all sub-issues.

Approach: start with pure-logic modules (fast feedback, no mocking), then move to mock-heavy helpers, tasks, services, API, and commands. Each sub-issue is self-contained and reviewable on its own.

### Tier 1 — Pure logic (quick wins)

- [ ] #2194 — `osism/utils/inventory.py`
- [ ] #2195 — `osism/data/enums.py`
- [ ] #2196 — `osism/data/playbooks.py`
- [ ] #2197 — `osism/settings.py`
- [ ] #2198 — `osism/tasks/conductor/sonic/{constants,bgp}.py`

### Tier 2 — Conductor helpers (pure logic + lightweight NetBox stubs)

- [ ] #2202 — `osism/tasks/conductor/utils.py` — `deep_compare`, `deep_merge`, `deep_decrypt`, `_is_secret_key`, `load_yaml_file`
- [ ] #2203 — `osism/tasks/conductor/sonic/cache.py` — `InterfaceCache` (thread-lock, hit/miss)
- [ ] #2204 — `osism/tasks/conductor/sonic/device.py` — `get_device_platform`, `get_device_hostname`, `get_device_mac_address`
- [ ] #2205 — `osism/tasks/conductor/sonic/connections.py` — `find_interconnected_devices` and helpers
- [ ] #2206 — `osism/tasks/conductor/config.py` + `osism/tasks/conductor/netbox.py` — `get_configuration`, `get_nb_device_query_list_*`, `get_device_oob_ip`

### Tier 3 — Conductor core (large modules, may split)

- [ ] #2219 — `osism/tasks/conductor/sonic/interface.py` — conversion helpers (`get_speed_from_port_type`, `convert_*_to_*`, alias mapping, `get_port_config`)
- [ ] #2220 — `osism/tasks/conductor/sonic/interface.py` — `detect_breakout_ports`, `detect_port_channels`
- [ ] #2221 — `osism/tasks/conductor/sonic/config_generator.py` — `generate_sonic_config`, metalbox/NTP/DNS caches, log-server, SNMP
[...]

PR #2193, opened 2026-04-24
Introduce foundation for unit tests (pytest + Zuul CI)

berendt:
Set up the minimal infrastructure so unit tests can be added in follow-up issues. Covers #2192.

- Add [dev-packages] in Pipfile: pytest, pytest-cov, pytest-mock
- Create tests/ layout with a smoke test asserting osism is importable and exposes __version__
- Configure pytest via [tool:pytest] in setup.cfg
- Add playbooks/test-unit.yml and a new python-osism-unit-tests Zuul job wired into the check and periodic-daily pipelines; the existing python-osism-test-setup job is kept in parallel (replace decision left to review per the issue)
- Document local test execution in README.md

Closes osism/python-osism#2192

AI-assisted: Claude Code

Issue #2192, opened 2026-04-24
Introduce foundation for unit tests (pytest + Zuul CI integration)


berendt:
## Background

The `osism/python-osism` repository currently has **no unit tests at all**:

- No `tests/` directory exists
- No pytest configuration (neither in `setup.cfg`, `pyproject.toml`, nor as `pytest.ini`)
- No test dependencies in `Pipfile` or `requirements.txt`
- The Zuul job `python-osism-test-setup` (see `playbooks/test-setup.yml`) only installs the package via `pipenv install && pipenv run pip install .` but does not run any tests

The goal of this issue is to establish the **foundation** so that unit tests can be written and automatically executed in Zuul CI. Actual test coverage of the modules (`osism/commands/`, `osism/utils/`, `osism/tasks/`, `osism/services/`, `osism/api.py`, …) will be addressed in follow-up issues.

## Scope

**In scope (foundation only):**

- Set up the test framework (pytest)
- Establish directory structure and configuration
- One or two minimal smoke tests so the CI pipeline passes and the infrastructure is validated
- Zuul integration: new job `python-osism-unit-tests` running in the `check` pipeline
- Documentation on how to run tests locally

**Out of scope:**

- Full test coverage of existing modules
- Integration / end-to-end tests (the existing `pytest-testinfra`-based scenarios remain untouched)
- Coverage gates / thresholds (can be added later)

### 1. Test dependencies

Add a new `[dev-packages]` section in `Pipfile` (or alternatively move them into a separate `requirements.test.txt`):

- `pytest`
- `pytest-cov`
- `pytest-mock`

Versions should be pinned, consistent with the existing style in `Pipfile`.

### 2. Directory structure

```
tests/
├── __init__.py
├── conftest.py
└── unit/
    ├── __init__.py
    └── test_smoke.py
```

- `conftest.py` initially only contains a placeholder for shared fixtures to be added later.
- `tests/unit/test_smoke.py` contains a trivial test (e.g. `import osism` / version check) so that a failure in the test infrastructure is immediately visible.

### 3. pytest configuration

Add a `[tool:pytest]` section to `setup.cfg` (consistent with the existing setup):

```ini
[tool:pytest]
testpaths = tests/unit
python_files = test_*.py
python_classes = Test*
python_functions = test_*
addopts = -ra --strict-markers
```

Do not set a coverage threshold — only enable optional reporting via `--cov=osism`.

### 4. Zuul CI integration

In `.zuul.yaml`:

[...]

PR #2333, opened 2026-06-09
Add unit tests for SONiC sync and exporter modules and fix issues found in review

### LINKED ISSUE #2224: Unit tests for osism/tasks/conductor/sonic/{sync,exporter}.py
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul infrastructure). Part of **Tier 3** (#2199). Closes the SONiC sync pipeline tests:

- `osism/tasks/conductor/sonic/sync.py` (`sync_sonic`) is the entry-point Celery task that orchestrates SONiC config generation, NetBox export, file export, and cache management.
- `osism/tasks/conductor/sonic/exporter.py` (`save_config_to_netbox`, `export_config_to_file`) writes the generated configuration to NetBox local context and to the SONiC export directory on disk.

These two modules share enough state (caches, device, generated config) that one combined test module is the smallest reasonable unit.

## Scope

Add two test modules:

- `tests/unit/tasks/conductor/sonic/test_sync.py` — for `osism/tasks/conductor/sonic/sync.py`
- `tests/unit/tasks/conductor/sonic/test_exporter.py` — for `osism/tasks/conductor/sonic/exporter.py`

### `sync_sonic(device_name=None, task_id=None, show_diff=True)` — `sync.py:25`

Patch the heavy external dependencies at their import site inside `sync.py`:

- `osism.tasks.conductor.sonic.sync.utils.nb`
- `...get_nb_device_query_list_sonic`
- `...find_interconnected_devices`
- `...calculate_minimum_as_for_group`
- `...generate_sonic_config`
- `...save_config_to_netbox`
- `...export_config_to_file`
- `...clear_interface_cache`, `...clear_all_caches`, `...clear_vip_addresses_cache`
- `..._load_metalbox_devices_cache`, `...load_vip_addresses_cache`
- `...get_interface_cache_stats`
- `...utils.push_task_output`, `...utils.finish_task_output`

Do **not** test `find_interconnected_devices` or `generate_sonic_config` here — they have their own issues. Stub them.

**Cache lifecycle**

- Caches cleared at start (`clear_interface_cache`, `clear_all_caches`) and at end (`clear_interface_cache`, `clear_all_caches`, `clear_vip_addresses_cache`) — verify the call order
- `_load_metalbox_devices_cache` and `load_vip_addresses_cache` called once per invocation

**Single-device path (`device_name="sw-1"`)**

- Device exists, role slug ∈ `DEFAULT_SONIC_ROLES` → device added, processed
- Device exists, role slug ∉ allowed list → warning logged, returns `{}`
- Device exists, `role` is `None` → warning logged, returns `{}`
- `nb.dcim.devices.get` returns `None` → error logged, returns `{}`
- `nb.dcim.devices.get` raises → error logged, returns `{}`
- Single device is a spine → all spine/superspine devices fetched for group detection (verify `nb.dcim.devices.filter` called via `nb_device_query_list`)
- Single device is a leaf → uses devices list as-is (no extra fetch)

**Multi-device path (`device_name=None`)**

- `nb_device_query_list` produces devices; only those with allowed roles are kept
- Devices without `role` skipped silently

**Per-device processing**

- HWSKU missing → device skipped, debug log
- HWSKU not in `SUPPORTED_HWSKUS` → warning logged, device skipped
- HWSKU valid → `generate_sonic_config(device, hwsku, device_as_mapping, config_version)` called and result stored under `device.name`
- `config_version` taken from `device.custom_fields["sonic_parameters"]["config_version"]` if present
- AS mapping built from `calculate_minimum_as_for_group` for each spine group
- AS mapping passed into `generate_sonic_config`
- Group with `min_as=None` → not propagated to `device_as_mapping`

**Diff handling (`show_diff=True`)**

- `save_config_to_netbox(..., return_diff=True)` returns `(True, diff_text)` and `task_id` set → diff streamed via `push_task_output`
- Returns `(True, None)` (first-time configuration) → "First-time configuration created" message pushed
- Returns `(False, None)` → no diff output, but `export_config_to_file` still called
- `show_diff=False` → `save_config_to_netbox(device, sonic_config)` called without `return_diff`

**Task output**

- `task_id` set → `push_task_output("Processing device: ...")` per device, plus `finish_task_output(task_id, rc=0)` at end
- `task_id` `None` → no `push_task_output` calls

**Cache stats**

- `get_interface_cache_stats` non-empty → debug log
- Empty/`None` stats → no log

### `save_config_to_netbox(device, config, return_diff=False)` — `exporter.py:15`

Patch `osism.tasks.conductor.sonic.exporter.utils.nb`.

- `device.local_context_data` `None` → first-time path, `device.save()` called, returns `True` (or `(True, None)` with `return_diff`)
- `device.local_context_data` `{"sonic_config": same}` → `DeepDiff` empty, returns `False` / `(False, None)`, `device.save()` not called
- `device.local_context_data` differs → diff generated via `difflib.unified_diff`, journal entry created via `nb.extras.journal_entries.create`, `device.save()` called, returns `True` / `(True, diff_text)`
- Journal create raises → error logged, but device save still happens, return value still indicates change
- `device.save()` (or any other op) raises → error logged, returns `False` / `(False, None)`
- Diff returned in tuple form when `return_diff=True`; bool form otherwise

### `export_config_to_file(device, config)` — `exporter.py:106`

Patch `osism.tasks.conductor.sonic.exporter.settings`, `os.makedirs`, `os.path.exists`, `os.path.islink`, `os.symlink`, `os.remove`, and `builtins.open` (`mock_open`). Patch `get_device_hostname`.

**Filename selection**

- `SONIC_EXPORT_IDENTIFIER="hostname"` → uses `get_device_hostname(device)`
- `SONIC_EXPORT_IDENTIFIER="serial-number"` and `device.serial="ABC123"` → uses serial
- `serial-number` mode but `device.serial` empty/missing → warning logged, falls back to hostname
- Filename is `f"{prefix}{identifier}{suffix}"`, joined under `export_dir`

**Diff handling**

- File missing → write happens, returns `True`
- File exists, JSON identical → no write, returns `False`
- File exists, JSON differs → write happens, returns `True`
- File exists but unreadable / invalid JSON → warning logged, write happens, returns `True`

**Symlink handling (only when `serial-number` + `device.serial` truthy)**

- Hostname filepath does not exist → `os.symlink` called
- Hostname filepath exists as a regular file → `os.remove` called first, then `os.symlink`
- Hostname filepath is a dangling symlink (`os.path.exists=False, os.path.islink=True`) → `os.remove` called first
- `os.symlink` raises → error logged, but function still returns `True` (the config was written successfully)
- `identifier_type="hostname"` → no symlink attempt
- `identifier_type="serial-number"` but `device.serial` falsy → no symlink attempt, debug log

**Error handling**

- Outer exception (e.g. `os.makedirs` fails) → error logged, returns `False`

## Mocking hints

- Build a small reusable device factory: `SimpleNamespace(name="sw-1", id=1, serial="ABC", role=SimpleNamespace(slug="leaf"), local_context_data=None, custom_fields={"sonic_parameters": {"hwsku": "Accton-AS7326-56X"}}, config_context={})`.
- Stub `generate_sonic_config` to return a minimal config like `{"PORT": {"Ethernet0": {}}}` so the `len(sonic_config["PORT"])` log doesn't crash.
- For `_load_metalbox_devices_cache` / `clear_all_caches` patches: just verify call counts. Don't try to exercise the real cache.
- For `export_config_to_file`, prefer a `tmp_path` fixture for real filesystem if it simplifies the diff/symlink tests; otherwise patch `os.*` and `open`.
- For `save_config_to_netbox`, the `DeepDiff` call works on plain dicts — no extra mocking needed for the equality path.

## Definition of Done

- [ ] `tests/unit/tasks/conductor/sonic/test_sync.py` and `tests/unit/tasks/conductor/sonic/test_exporter.py` created
- [ ] `pytest --cov=osism.tasks.conductor.sonic.sync` and `pytest --cov=osism.tasks.conductor.sonic.exporter` ≥ 90 %
- [ ] `pipenv run pytest tests/unit/tasks/conductor/sonic/test_sync.py tests/unit/tasks/conductor/sonic/test_exporter.py` passes locally

## Dependencies

berendt:
Closes #2224. Part of the Tier 3 SONiC unit-test effort (#2199), follow-up
to #2192 / #2193.

Adds two test modules covering the SONiC sync pipeline. Both mirror the
existing `tests/unit/tasks/conductor/sonic/` conventions: the shared
`mock_nb` / `loguru_logs` fixtures, `SimpleNamespace` device stubs, and a
per-module fixture that stubs collaborators at their import site (the
`patch_orchestrator_helpers` pattern).

The review surfaced seven real defects in `exporter.py` / `sync.py` that the
initial tests would otherwise have pinned as intended behavior. Since the
requested regression tests only pass against fixed sources, the fixes are
included here as separate commits (see below), so the PR is no longer
tests-only.

## Change set (commit order)

1. **Add unit tests for SONiC exporter module** — `test_exporter.py`
   - `save_config_to_netbox`: first-time create, no-change (`DeepDiff` empty,
     no save), changed config (unified diff + journal entry + save), journal
     failure still saves. Both the tuple (`return_diff=True`) and bool return
     forms are asserted.
   - `export_config_to_file`: filename selection (hostname vs. serial-number,
     empty/missing serial falls back to hostname with a warning), diff
     handling (missing / identical / changed / unreadable file), symlink
     handling (absent target, existing regular file, dangling symlink,
     `os.symlink` failure, hostname mode / falsy serial skip), and the outer
     `os.makedirs` failure path.
   - File-export tests use a real `tmp_path` filesystem where that reads more
     clearly (the symlink end-state proves remove-then-symlink ordering), and
     patch `os.*` only to inject the symlink / makedirs failures.

2. **Add unit tests for SONiC sync orchestrator** — `test_sync.py`
   - Cache lifecycle: exact start/end clear ordering via `attach_mock`, and
     the metalbox / VIP loads called once per invocation.
   - Single-device path: allowed role processed; disallowed role, `role=None`,
     not-found, and lookup-raises each return `{}`; single spine triggers the
     full spine fetch (`nb.dcim.devices.filter`), single leaf does not.
   - Multi-device path: only allowed roles kept; role-less devices skipped.
   - Per-device: missing / unsupported / valid HWSKU; `config_version` read
     from custom fields.
   - AS mapping: computed per spine group, propagated into
     `generate_sonic_config`, and a `min_as=None` group is not propagated.
   - Diff handling: diff streamed to task output, first-time message,
     no-change still exports, `show_diff=False` calls save without
     `return_diff`.
   - Task output gated on `task_id`; cache-stats logged only when non-empty.

   `generate_sonic_config` and `find_interconnected_devices` are stubbed (they
   have their own issues under #2199), per the issue's instructions.

## Review feedback fixes (each with regression tests)

3. **Own only the sonic_config key in NetBox local context** (review finding 1)
   — `save_config_to_netbox` no longer replaces the device's entire
   `local_context_data` dict; sibling keys such as `frr_parameters` /
   `netplan_parameters` survive, and diffing covers only the owned
   `sonic_config` key (mirrors the #2297 partitioned-ownership model).

4. **Create SONiC journal entry only after successful save** (finding 7)
   — a failed `device.save()` can no longer leave an orphaned journal entry
   claiming the update succeeded.

5. **Write SONiC config exports atomically** (finding 3)
   — temp file + `os.replace()` so a mid-write failure (ENOSPC, killed
   process) cannot truncate the previous export.

6. **Reconcile hostname symlink independently of config changes**
   (findings 2 + 4) — a missing/stale link is repaired even when the config
   content is unchanged, the reconciliation is idempotent, and a
   `hostname == serial` device no longer destroys its own config via a
   self-referential symlink.

7. **Always clean up caches and finish task output in sync_sonic** (finding 5)
   — `try/finally` around the sync body; early returns and per-device
   failures no longer leak the module-level caches into the next run, finish
   the task with `rc=1`, and a failing device no longer aborts the remaining
   devices. The summary log no longer raises on `PORT`-less configs.

8. **Raise on SONiC persistence failures instead of reporting no change**
   (finding 6) — failed NetBox saves and file exports raise instead of
   returning the same value as "no changes"; `sync_sonic` reports them via
   `rc=1` instead of finishing failed syncs with `rc=0`.

## Local verification

- `black --check` — pass
- `flake8` — pass
- `pytest` — not run locally; the unit-test suite runs in the
  `python-osism-unit-tests` Zuul job. The coverage gate
  (`--cov=...sync`/`...exporter` >= 90 %) is verified there.

### DIFF
--- a/tests/unit/tasks/conductor/sonic/test_exporter.py
+++ b/tests/unit/tasks/conductor/sonic/test_exporter.py
@@ -202,6 +202,25 @@ def test_save_config_save_failure_returns_false(mock_nb, loguru_logs):
     assert _has_log(loguru_logs, "ERROR", "Failed to save local context")
 
 
+def test_save_config_changed_path_save_failure_creates_no_journal(
+    mock_nb, loguru_logs
+):
+    """A failed save on the changed path must not leave an orphaned journal
+    entry claiming the update succeeded — journal creation follows the save."""
+    device = _make_save_device(
+        local_context_data={"sonic_config": {"PORT": {"Ethernet0": {}}}}
+    )
+    device.save.side_effect = RuntimeError("netbox write failed")
+
+    result = save_config_to_netbox(
+        device, {"PORT": {"Ethernet1": {}}}, return_diff=True
+    )
+
+    assert result == (False, None)
+    mock_nb.extras.journal_entries.create.assert_not_called()
+    assert _has_log(loguru_logs, "ERROR", "Failed to save local context")
+
+
 def test_save_config_save_failure_bool_form(mock_nb):
     device = _make_save_device(local_context_data=None)
     device.save.side_effect = RuntimeError("boom")
--- a/tests/unit/tasks/conductor/sonic/test_exporter.py
+++ b/tests/unit/tasks/conductor/sonic/test_exporter.py
@@ -361,6 +361,28 @@ def test_export_unreadable_file_is_overwritten(
     assert _has_log(loguru_logs, "WARNING", "Could not read existing config file")
 
 
+def test_export_write_failure_preserves_previous_export(
+    tmp_path, export_settings, patch_hostname, mocker, loguru_logs
+):
+    """The export is written atomically: a mid-write failure must leave the
+    previous export intact instead of truncating it, and not leak a temp file."""
+    export_settings(tmp_path, identifier="hostname")
+    device = SimpleNamespace(name="sw-1", serial="ABC123")
+    target = tmp_path / "osism_sw-1_config_db.json"
+    old_config = {"PORT": {"Ethernet9": {}}}
+    target.write_text(json.dumps(old_config))
+    mocker.patch(
+        "osism.tasks.conductor.sonic.exporter.json.dump",
+        side_effect=OSError("no space left on device"),
+    )
+
+    assert export_config_to_file(device, {"PORT": {"Ethernet0": {}}}) is False
+
+    assert json.loads(target.read_text()) == old_config
+    assert not (tmp_path / "osism_sw-1_config_db.json.tmp").exists()
+    assert _has_log(loguru_logs, "ERROR", "Failed to export config")
+
+
 # --- symlink handling -------------------------------------------------------
 
 
--- a/tests/unit/tasks/conductor/sonic/test_exporter.py
+++ b/tests/unit/tasks/conductor/sonic/test_exporter.py
@@ -121,6 +121,59 @@ def test_save_config_change_bool_form(mock_nb):
     assert result is True
 
 
+def test_save_config_preserves_sibling_context_keys(mock_nb):
+    """Only the ``sonic_config`` key is owned — sibling keys like
+    ``frr_parameters`` must survive a config update untouched."""
+    frr = {"asn": 4200000001, "loopback": "10.0.0.1"}
+    device = _make_save_device(
+        local_context_data={
+            "frr_parameters": frr,
+            "sonic_config": {"PORT": {"Ethernet0": {}}},
+        }
+    )
+
+    result = save_config_to_netbox(device, {"PORT": {"Ethernet1": {}}})
+
+    assert result is True
+    device.save.assert_called_once_with()
+    assert device.local_context_data == {
+        "frr_parameters": frr,
+        "sonic_config": {"PORT": {"Ethernet1": {}}},
+    }
+
+
+def test_save_config_sibling_only_context_is_first_time(mock_nb):
+    """A context holding only sibling keys has no ``sonic_config`` to diff:
+    the first-time path adds the key and keeps the siblings."""
+    frr = {"asn": 4200000001}
+    device = _make_save_device(local_context_data={"frr_parameters": frr})
+
+    result = save_config_to_netbox(device, {"PORT": {}}, return_diff=True)
+
+    assert result == (True, None)
+    device.save.assert_called_once_with()
+    assert device.local_context_data == {
+        "frr_parameters": frr,
+        "sonic_config": {"PORT": {}},
+    }
+    # First-time configuration creates no journal entry.
+    mock_nb.extras.journal_entries.create.assert_not_called()
+
+
+def test_save_config_sibling_keys_do_not_trigger_change(mock_nb):
+    """Diffing covers only ``sonic_config`` — sibling keys must not register
+    as removals and force a save when the SONiC config itself is unchanged."""
+    config = {"PORT": {"Ethernet0": {}}}
+    device = _make_save_device(
+        local_context_data={"frr_parameters": {"asn": 1}, "sonic_config": config}
+    )
+
+    result = save_config_to_netbox(device, config)
+
+    assert result is False
+    device.save.assert_not_called()
+
+
 def test_save_config_journal_failure_still_saves(mock_nb, loguru_logs):
     """A failing journal create is logged but must not block the save."""
     mock_nb.extras.journal_entries.create.side_effect = RuntimeError("journal down")
--- a/tests/unit/tasks/conductor/sonic/test_exporter.py
+++ b/tests/unit/tasks/conductor/sonic/test_exporter.py
@@ -430,6 +430,78 @@ def test_export_replaces_dangling_symlink(tmp_path, export_settings, patch_hostn
     assert os.readlink(link) == "osism_ABC123_config_db.json"
 
 
+def test_export_unchanged_config_repairs_missing_symlink(
+    tmp_path, export_settings, patch_hostname
+):
+    """Symlink reconciliation is independent of a content change: an unchanged
+    config with a missing hostname link still gets the link created."""
+    export_settings(tmp_path, identifier="serial-number")
+    device = SimpleNamespace(name="sw-1", serial="ABC123")
+    config = {"PORT": {"Ethernet0": {}}}
+    (tmp_path / "osism_ABC123_config_db.json").write_text(json.dumps(config))
+
+    assert export_config_to_file(device, config) is False
+
+    link = tmp_path / "osism_sw-1_config_db.json"
+    assert link.is_symlink()
+    assert os.readlink(link) == "osism_ABC123_config_db.json"
+
+
+def test_export_unchanged_config_repoints_stale_symlink(
+    tmp_path, export_settings, patch_hostname
+):
+    """A hostname link pointing at the wrong target is repointed even when
+    the exported content is unchanged."""
+    export_settings(tmp_path, identifier="serial-number")
+    device = SimpleNamespace(name="sw-1", serial="ABC123")
+    config = {"PORT": {"Ethernet0": {}}}
+    (tmp_path / "osism_ABC123_config_db.json").write_text(json.dumps(config))
+    link = tmp_path / "osism_sw-1_config_db.json"
+    link.symlink_to("osism_OTHER_config_db.json")
+
+    assert export_config_to_file(device, config) is False
+
+    assert os.readlink(link) == "osism_ABC123_config_db.json"
+
+
+def test_export_existing_correct_symlink_left_untouched(
+    tmp_path, export_settings, patch_hostname, mocker
+):
+    """A link that already points at the config file is not removed and
+    recreated on every run."""
+    export_settings(tmp_path, identifier="serial-number")
+    device = SimpleNamespace(name="sw-1", serial="ABC123")
+    config = {"PORT": {"Ethernet0": {}}}
+    (tmp_path / "osism_ABC123_config_db.json").write_text(json.dumps(config))
+    link = tmp_path / "osism_sw-1_config_db.json"
+    link.symlink_to("osism_ABC123_config_db.json")
+    remove_spy = mocker.spy(os, "remove")
+    symlink_spy = mocker.spy(os, "symlink")
+
+    assert export_config_to_file(device, config) is False
+
+    remove_spy.assert_not_called()
+    symlink_spy.assert_not_called()
+    assert os.readlink(link) == "osism_ABC123_config_db.json"
+
+
+def test_export_hostname_equals_serial_keeps_config_file(
+    tmp_path, export_settings, patch_hostname
+):
+    """When the serial equals the hostname the config file already lives at
+    the hostname path — no self-referential symlink may replace it."""
+    export_settings(tmp_path, identifier="serial-number")
+    device = SimpleNamespace(name="sw-1", serial="sw-1")
+    config = {"PORT": {"Ethernet0": {}}}
+
+    assert export_config_to_file(device, config) is True
+
+    target = tmp_path / "osism_sw-1_config_db.json"
+    assert target.exists()
+    assert not target.is_symlink()
+    assert json.loads(target.read_text()) == config
+
+
 def test_export_symlink_failure_still_returns_true(
     tmp_path, export_settings, patch_hostname, mocker, loguru_logs
 ):
--- a/tests/unit/tasks/conductor/sonic/test_exporter.py
+++ b/tests/unit/tasks/conductor/sonic/test_exporter.py
@@ -191,20 +191,19 @@ def test_save_config_journal_failure_still_saves(mock_nb, loguru_logs):
     assert _has_log(loguru_logs, "ERROR", "Failed to save diff to journal")
 
 
-def test_save_config_save_failure_returns_false(mock_nb, loguru_logs):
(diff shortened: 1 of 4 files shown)


tests/unit/tasks/conductor/sonic/test_exporter.py
```diff
@@ -0,0 +1,385 @@
...
+def test_save_config_no_change_return_diff_tuple(mock_nb):
+    config = {"PORT": {"Ethernet0": {}}}
+    device = _make_save_device(local_context_data={"sonic_config": config})
+
+    result = save_config_to_netbox(device, config, return_diff=True)
+
+    assert result == (False, None)
+
+
+def test_save_config_change_writes_journal_and_saves(mock_nb):
```
ideaship, line 90: **Finding 1 — `local_context_data` wholesale-replace clobbers sibling keys (data loss)**

These changed-context save tests only build `local_context_data={"sonic_config": ...}`, which hides that `save_config_to_netbox` replaces the device's *entire* dict (`exporter.py:88`/`:94`) and diffs the full existing context against `{"sonic_config": ...}` (`exporter.py:42`). Please add a case with a sibling key, e.g. `local_context_data={"frr_parameters": {...}, "sonic_config": old}`, and assert the sibling survives.

It won't today — verified by running the real exporter against a `leaf`-role device:
```
BEFORE: ['frr_parameters', 'sonic_config']
AFTER : ['sonic_config']
```
The exporter deletes `frr_parameters`, reports it as a removal in its diff log, and writes a journal entry attributing the deletion to the SONiC update. `frr_parameters` is not hypothetical — the inventory-reconciler reads it (and `netplan_parameters`) out of `local_context_data` via `config_context`.

Tests-only PR, so the source fix is a separate change: mirror the #2297 partitioned-ownership model — own and replace only the `sonic_config` key (`device.local_context_data = {**(device.local_context_data or {}), "sonic_config": config}`) and diff only that key. This test should pin sibling-key survival once that lands.

tests/unit/tasks/conductor/sonic/test_exporter.py
```diff
@@ -0,0 +1,385 @@
...
+        device, {"PORT": {"Ethernet1": {}}}, return_diff=True
+    )
+
+    assert changed is True
+    assert diff_text is not None
+    device.save.assert_called_once_with()
+    assert _has_log(loguru_logs, "ERROR", "Failed to save diff to journal")
+
+
+def test_save_config_save_failure_returns_false(mock_nb, loguru_logs):
```
ideaship: **Finding 6 — failed writes reported to the task layer as `rc=0`**

`assert result == (False, None)` pins an ambiguity: `save_config_to_netbox` returns the same `(False, None)` for a *failed* save (`exporter.py:103`) as for *no change* (`:48`), and `export_config_to_file` does the same (`:226` vs `:222`). In `sync.py` (not in this diff), `:219` then treats a failed write as "no changes", logs accordingly, and still finishes `rc=0` (`:244`) — failed syncs report success. Source fix: failures should raise or return a distinct sentinel so task status reflects them; this test should assert the distinction once that lands.

tests/unit/tasks/conductor/sonic/test_exporter.py
```diff
@@ -0,0 +1,385 @@
...
+    device = _make_save_device(local_context_data=None)
+    device.save.side_effect = RuntimeError("netbox write failed")
+
+    result = save_config_to_netbox(device, {"PORT": {}}, return_diff=True)
+
+    assert result == (False, None)
+    assert _has_log(loguru_logs, "ERROR", "Failed to save local context")
+
+
+def test_save_config_save_failure_bool_form(mock_nb):
```
ideaship: **Finding 7 — journal entry created before `device.save()`**

Both save-failure tests pass `local_context_data=None`, hitting only the first-time branch. Please add a *changed-path* save failure (existing `sonic_config` + raising `device.save()`). In `exporter.py` (not in this diff) the journal entry is created at `:71` *before* `save()` at `:89`, so a failed save returns `(False, None)` while an orphaned journal entry already claims the update succeeded (and the in-memory device stays mutated). Source fix: move journal creation after a successful save.

tests/unit/tasks/conductor/sonic/test_exporter.py
```diff
@@ -0,0 +1,385 @@
...
+    assert target.exists()
+    assert not target.is_symlink()
+    assert _has_log(loguru_logs, "WARNING", "Serial number not found")
+    assert _has_log(loguru_logs, "DEBUG", "Symlink conditions not met")
+
+
+# --- diff handling ----------------------------------------------------------
+
+
+def test_export_no_change_returns_false(tmp_path, export_settings, patch_hostname):
```
ideaship, line 318: **Finding 4 — symlink repair gated on `config_changed`**

Symlink reconciliation should be tested independently of a content change. In `exporter.py` (not in this diff) the whole symlink block is gated by `if config_changed:` (`:177`), so an unchanged serial-mode config with a missing/stale hostname link is never repaired — and after a symlink failure, a later unchanged run never retries. Please add a case: existing identical config + missing hostname link, and assert the link gets created. It currently won't.

tests/unit/tasks/conductor/sonic/test_exporter.py
```diff
@@ -0,0 +1,385 @@
...
+
+    assert export_config_to_file(device, new_config) is True
+    assert json.loads(target.read_text()) == new_config
+    assert _has_log(loguru_logs, "WARNING", "Could not read existing config file")
+
+
+# --- symlink handling -------------------------------------------------------
+
+
+def test_export_creates_symlink_when_hostname_absent(
```
ideaship, line 380: **Finding 2 — hostname == serial → self-referential symlink (low priority)**

Low-priority defensive case. Every symlink test uses distinct hostname (`sw-1`) and serial (`ABC123`); add one where they're equal. In `exporter.py` (not in this diff), when `serial == hostname` the just-written config file *is* the hostname path, so `:207` `os.remove`s it and `:209` symlinks it to itself — config destroyed, yet the function returns `True`. One-line source fix: skip the symlink when `filepath == hostname_filepath`.

Unlikely in practice — needs the serial set equal to the device name/inventory_hostname, which doesn't happen with real manufacturer serials or on the seeded testbed (serials unset → safe fallback). But the guard is trivial against silent data loss that reports success.

tests/unit/tasks/conductor/sonic/test_exporter.py
```diff
@@ -0,0 +1,385 @@
...
+
+    target = tmp_path / "osism_sw-1_config_db.json"
+    assert target.exists()
+    assert not target.is_symlink()
+
+
+# --- error handling ---------------------------------------------------------
+
+
+def test_export_makedirs_failure_returns_false(
```
ideaship: **Finding 3 — non-atomic write truncates the previous export (cheap hygiene)**

Cheap-hygiene note, not a blocker. Error coverage only injects a `makedirs` failure; a write/serialize-failure case (patch `json.dump` to raise, assert the *previous* export is still intact) would be a nice addition. In `exporter.py` (not in this diff) the file is opened `"w"` at `:179`, truncating before serialization, so a mid-write crash (ENOSPC / killed process) leaves the prior config truncated and returns `False`. Source fix: temp file + `os.replace()`.

Low priority here — the export self-heals on the next sync and the in-repo reader rejects bad JSON — but it would rank higher if we expected attackers or external consumers of the export dir.

tests/unit/tasks/conductor/sonic/test_sync.py
```diff
@@ -0,0 +1,474 @@
...
+def test_single_device_allowed_role_is_processed(mock_nb, patch_sync_deps):
+    device = make_device(name="sw-1", role_slug="leaf")
+    mock_nb.dcim.devices.get.return_value = device
+
+    result = sync_sonic(device_name="sw-1")
+
+    assert result == {"sw-1": {"PORT": {"Ethernet0": {}}}}
+
+
+def test_single_device_disallowed_role_returns_empty(
```
ideaship, line 149: **Finding 5 — cleanup / `finish_task_output` skipped on early returns and mid-loop exceptions (would block)**

These early-return tests check the return value and log message but not that caches were cleared or `finish_task_output` was called. In `sync.py` (not in this diff), the early returns (`:77`/`:80`/`:83`) and any exception in the per-device loop — e.g. the unguarded `len(sonic_config['PORT'])` at `:225` on a `PORT`-less config — return *after* caches load (`:42–52`) but *before* cleanup (`:237–239`) and `finish_task_output` (`:244`). Caches are module-level, so the leak corrupts the next run. Please assert cleanup + finalization on these paths and add a mid-loop raising-collaborator case; the source fix is a `try/finally`.

ideaship (changes requested):
Nothing  major, just some minor issues.

ideaship (approved):

Issue #2330, opened 2026-06-08
Ensure that the SONiC SNMP/GNMI services are accessible only on the management network


berendt:
## Problem

The generated SONiC ConfigDB does not restrict which networks can reach the
switch's SNMP and gNMI/telemetry services:

- **SNMP** (UDP/161): `_add_snmp_configuration` already binds the agent to the
  OOB IP in the management VRF (`SNMP_AGENT_ADDRESS_CONFIG` →
  `<oob_ip>|161|mgmt`, `config_generator.py:2168`), but that is only the listen
  address — there is no source-network restriction.
- **gNMI / telemetry** (`TELEMETRY` table, `config_db.json:458`): shipped as a
  static `"gnmi": {}` with no generator code and no binding, so the gNMI server
  listens on all interfaces by default.

Both services must be reachable **only** from the out-of-band management
network, consistent with #2329 (SSH).

## Goal

Generate, per switch, SONiC control-plane ACLs that permit SNMP and
gNMI/telemetry only from the device's management/OOB network and implicitly
drop them from everywhere else.

## Proposed approach

Use SONiC control-plane ACLs handled by `caclmgrd`: `ACL_TABLE` entries of
`type: CTRLPLANE` bound to the respective services, plus `ACL_RULE`s that
ACCEPT the management subnet. When a CTRLPLANE table binds a service, caclmgrd
installs an implicit default-drop for that service, so any source outside the
permitted rules can no longer reach it.

The management subnet is already available in the generator:
`get_device_oob_ip(device)` returns `(oob_ip, prefix_len)`; the network is
`IPv4Network(f"{oob_ip}/{prefix_len}", strict=False)` — the same data used for
`MGMT_INTERFACE` (`config_generator.py:363-369`).

Only the device's own OOB management subnet is permitted; SNMP polling stations
and telemetry collectors are assumed to live on the management network. No other
sources are added in this iteration.

Example (per device, subnet derived from the OOB IP):

```json
"ACL_TABLE": {
    "SNMP_ONLY": {
        "policy_desc": "SNMP_ONLY",
        "type": "CTRLPLANE",
        "services": ["SNMP"]
    },
    "GNMI_ONLY": {
        "policy_desc": "GNMI_ONLY",
        "type": "CTRLPLANE",
        "services": ["EXTERNAL_CLIENT"]
    }
},
"ACL_RULE": {
    "SNMP_ONLY|RULE_1": {
        "PRIORITY": "9999",
[...]

PR #2337, opened 2026-06-10
sonic: Restrict SSH to the OOB management network

### LINKED ISSUE #2329: Ensure that the SONiC SSH service is accessible only on the management network
## Problem

The generated SONiC ConfigDB does not restrict which networks can reach the
switch's SSH service (TCP/22). The management VRF is enabled and `eth0` carries
the OOB management IP, but the control-plane SSH service is not locked down to
the management network — front-panel / in-band interfaces with IPs in the
default VRF can still reach it. SSH must be reachable **only** from the
out-of-band management network.

## Goal

Generate, per switch, a SONiC control-plane ACL that permits SSH only from the
device's management/OOB network and implicitly drops SSH from everywhere else.

## Proposed approach

Use SONiC control-plane ACLs handled by `caclmgrd`: an `ACL_TABLE` of
`type: CTRLPLANE` bound to the `SSH` service, plus an `ACL_RULE` that ACCEPTs
the management subnet. When a CTRLPLANE table binds a service, caclmgrd installs
an implicit default-drop for that service, so any source outside the permitted
rules can no longer reach SSH.

The management subnet is already available in the generator:
`get_device_oob_ip(device)` returns `(oob_ip, prefix_len)`; the network is
`IPv4Network(f"{oob_ip}/{prefix_len}", strict=False)` — the same data used for
`MGMT_INTERFACE` (`config_generator.py:363-369`).

Only the device's own OOB management subnet is permitted; no other sources
(metalbox/gateway, central admin networks, etc.) are added in this iteration.

Example (per device, subnet derived from the OOB IP):

```json
"ACL_TABLE": {
    "SSH_ONLY": {
        "policy_desc": "SSH_ONLY",
        "type": "CTRLPLANE",
        "services": ["SSH"]
    }
},
"ACL_RULE": {
    "SSH_ONLY|RULE_1": {
        "PRIORITY": "9999",
        "PACKET_ACTION": "ACCEPT",
        "SRC_IP": "<oob_subnet>/<prefix>",
        "IP_TYPE": "IP"
    }
}
```

## Implementation notes

- New helper `_add_ssh_acl_configuration(config, device, oob_ip_result)` in
  `osism/tasks/conductor/sonic/config_generator.py`, called from
  `generate_sonic_config` near the management-interface block, only when an OOB
  IP exists. When no OOB IP is present, no ACL is emitted (and SSH is not
  locked down — to be revisited if needed).
- Register `ACL_TABLE` and `ACL_RULE` as generator-owned tables
  (`TOP_LEVEL_SCAFFOLD_KEYS` / `OWNED_TABLE_KEYS`) so they are rebuilt from
  scratch on every regen and never carry stale entries.
- Permit only the local OOB subnet (`oob_ip/prefix_len`, normalised to the
  network address). This matches "only the management network".
- Validation: SONiC `ACL_TABLE` / `ACL_RULE` are not yet in the generated
  pydantic schema set (`_generated/_schemas.py` only models `DASH_ACL_*`), so
  the validator emits warnings, not errors, for these tables. Consider adding
  the `sonic-acl` YANG model and regenerating (relates to #2258).
- Add unit tests under `tests/unit/tasks/conductor/sonic/`: ACL generated when
  OOB IP present, correct (network-normalised) subnet in the rule, table absent
  when no OOB IP.

## Related

- #2330 (SNMP/GNMI) — same control-plane-ACL mechanism; the helper and
  owned-table wiring should be shared.

berendt:
Closes #2329

## What this does

Generates, per switch, a SONiC control-plane ACL (handled by `caclmgrd`) that permits SSH only from the device's OOB management subnet. Once a `CTRLPLANE` table binds the `SSH` service, caclmgrd installs an implicit default-drop for it, so front-panel / in-band interfaces with IPs in the default VRF can no longer reach TCP/22.

## Walkthrough (commit order)

1. **`sonic: Take ownership of ACL_TABLE and ACL_RULE on regen`** — registers both tables in `ON_DEMAND_OWNED_TABLE_KEYS` (feeding `OWNED_TABLE_KEYS`), so base-config content is dropped up front and the tables are rebuilt from scratch on every regen. The on-demand registry was chosen over `TOP_LEVEL_SCAFFOLD_KEYS` so the tables are entirely **absent** when nothing emits them (no empty `{}` scaffold entries). Includes an orchestrator test pinning that stale ACL entries carried over from `config_db.json` do not survive a regen without an OOB IP. The existing exhaustive ownership test (`test_generate_sonic_config_every_owned_table_drops_stale_entries`) picks the new keys up automatically.

2. **`sonic: Restrict SSH to the OOB management network`** — adds `_add_ssh_acl_configuration(config, device, oob_ip_result)`, called from `generate_sonic_config` in the management-interface block, only when an OOB IP exists. The permitted subnet is derived from the same `get_device_oob_ip` result used for `MGMT_INTERFACE`, normalised to the network address via `IPv4Network(..., strict=False)`. Emits exactly the shape from the issue:
   - `ACL_TABLE["SSH_ONLY"]` — `type: CTRLPLANE`, `services: ["SSH"]`
   - `ACL_RULE["SSH_ONLY|RULE_1"]` — `PRIORITY 9999`, `PACKET_ACTION ACCEPT`, `SRC_IP <oob_network>/<prefix>`, `IP_TYPE IP`

   Tests: helper-level (table/rule shape, network normalisation incl. non-octet boundary, wholesale replacement of pre-existing entries) plus orchestrator glue (called with the full OOB result when present, not called and tables absent when not).

## Notes for the reviewer

- Only the device's own OOB subnet is permitted — no metalbox/gateway or central admin networks (per the issue, this iteration).
- When no OOB IP is present, no ACL is emitted and SSH is not locked down (per the issue, to be revisited if needed).
- `ACL_TABLE`/`ACL_RULE` are not in the generated pydantic schema set (`_generated/_schemas.py` only models `DASH_ACL_*`), so the validator reports them as warnings, not errors. Adding the `sonic-acl` YANG model is deferred (relates to #2258).
- SNMP/GNMI control-plane ACLs are #2330; the helper and owned-table wiring added here are the shared mechanism it can build on.
- No CHANGELOG edit — this repo batches changelog entries in release PRs.

AI-assisted: Claude Code

### DIFF
--- a/osism/tasks/conductor/sonic/config_generator.py
+++ b/osism/tasks/conductor/sonic/config_generator.py
@@ -431,7 +431,7 @@ def generate_sonic_config(device, hwsku, device_as_mapping=None, config_version=
         config["MGMT_INTERFACE"][f"eth0|{oob_ip}/{prefix_len}"] = {}
         metalbox_ip = _get_metalbox_ip_for_device(device)
         config["STATIC_ROUTE"]["mgmt|0.0.0.0/0"] = {"nexthop": metalbox_ip}
-        # Restrict control-plane services (SNMP, gNMI) to the OOB network
+        # Restrict control-plane services (SSH, SNMP, gNMI) to the OOB network
         _add_ctrlplane_acls(config, oob_ip, prefix_len)
     else:
         oob_ip = None
@@ -2369,15 +2369,15 @@ def _get_gnmi_port(config):
 
 
 def _add_ctrlplane_acls(config, oob_ip, prefix_len):
-    """Add control-plane ACLs restricting SNMP and gNMI to the OOB network.
+    """Add control-plane ACLs restricting SSH, SNMP and gNMI to the OOB network.
 
-    Emits ACL_TABLE entries of type CTRLPLANE bound to the SNMP and
+    Emits ACL_TABLE entries of type CTRLPLANE bound to the SSH, SNMP and
     EXTERNAL_CLIENT (gNMI/telemetry) caclmgrd services, plus one ACL_RULE
     per table accepting only the device's OOB management subnet (the
     network-normalised oob_ip/prefix_len). caclmgrd installs an implicit
     default-drop for every service bound in a CTRLPLANE table, so sources
-    outside the OOB subnet can no longer reach SNMP or gNMI. The SSH_ONLY
-    table (#2329) belongs here as well once implemented.
+    outside the OOB subnet -- e.g. front-panel interfaces with IPs in the
+    default VRF -- can no longer reach SSH, SNMP or gNMI.
 
     caclmgrd's EXTERNAL_CLIENT service has no built-in destination port
     (verified against sonic-host-services 202211 through master): the rule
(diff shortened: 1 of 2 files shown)


berendt:
Has to be revisited after merge of https://github.com/osism/python-osism/pull/2338.

ideaship (approved):

PR #2349, opened 2026-06-12
Add unit tests for osism/utils semaphore, redlock and task locks

### LINKED ISSUE #2230: Unit tests for osism/utils/__init__.py — semaphore, redlock, task locks
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul infrastructure). Part of **Tier 4** (#2199). Companion to the connection-init and task-output issues: covers the concurrency primitives (`RedisSemaphore`, redlock, NetBox semaphore) and the global task-lock helpers from `osism/utils/__init__.py`.

## Scope

Add `tests/unit/utils/test_init_locks.py` covering the helpers below in `osism/utils/__init__.py`.

### `RedisSemaphore` — `__init__.py:109`

Use a `MagicMock` for the redis client. Drive `time.time()` and `time.sleep()` either via `freezegun` or `mocker.patch` so the loop exits deterministically.

**`__init__`**

- Stores `redis`, `key=f"semaphore:{key}"`, `maxsize`, `timeout`; `identifier=None`
- `key` prefixing is applied even when the input already starts with `semaphore:` (no double-strip — confirm production behaviour)

**`acquire(timeout=None)`**

- `zcard < maxsize` on first try → `zadd` called with `{identifier: now}`, `self.identifier` set, returns `True`
- `zcard >= maxsize` for entire window → returns `False`
- `zremrangebyscore(self.key, 0, now-60)` called every iteration (cleanup)
- Default timeout fallback: `acquire(None)` with `self.timeout=None` → uses `10` seconds
- Explicit `acquire(timeout=5)` → overrides instance timeout
- Identifier is a fresh `uuid.uuid4()` per call (verify by patching `uuid.uuid4`)
- `time.sleep(0.01)` called between retries when full

**`release()`**

- Sets identifier → `zrem(self.key, self.identifier)` called and `identifier=None`
- Called twice → second call is a no-op (no extra `zrem`)
- Called without acquire → no-op

**Context manager (`__enter__` / `__exit__`)**

- `acquire` returns `True` → `__enter__` returns `self`
- `acquire` returns `False` → `__enter__` raises `TimeoutError` containing the key
- `__exit__` always calls `release()` and returns `False` (don't swallow exceptions)

### `create_redlock(key, auto_release_time=3600)` — `__init__.py:441`

Patch `pottery.Redlock`, `_init_redis`, `logging.getLogger`.

- Returns the `Redlock` instance built with `key`, `masters={redis}`, `auto_release_time`
- Default `auto_release_time=3600`
- Custom `auto_release_time=600` propagated
- Pottery logger is set to `CRITICAL` level (verify `setLevel(logging.CRITICAL)` call on the `"pottery"` logger)
- stdout/stderr suppressed during construction (the production code uses `redirect_stdout(devnull)`/`redirect_stderr(devnull)` — assert no construction-time noise leaks; one assertion via `capsys` is enough)

### `create_netbox_semaphore(netbox_url, max_connections=None)` — `__init__.py:469`

Patch `_init_redis`, `osism.utils.settings.NETBOX_MAX_CONNECTIONS`.

- `max_connections=None` → uses `settings.NETBOX_MAX_CONNECTIONS`
- Explicit `max_connections=20` → propagated
- Returns a `RedisSemaphore` with `key=f"netbox_semaphore_{md5_hash[:8]}"`, `timeout=30`, `redis_client=_init_redis()`
- Two different URLs → two different keys (verify the hash is per-URL)
- Same URL twice → identical key

### `set_task_lock(user=None, reason=None)` — `__init__.py:497`

Patch `_init_redis`, `osism.utils.settings.OPERATOR_USER`. Capture the JSON written via `redis.set`.

- `user=None` → falls back to `settings.OPERATOR_USER`
- Explicit `user="alice"` → used directly
- `reason=None` → stored as `null` in JSON
- Captured JSON contains `locked=True`, ISO-formatted `timestamp`, `user`, `reason`
- `redis.set` raises → returns `False`, error logged

### `remove_task_lock()` — `__init__.py:526`

- Calls `delete("osism:task_lock")` → returns `True`
- `delete` raises → returns `False`, error logged

### `is_task_locked()` — `__init__.py:541`

- Redis returns `None` → returns `None`
- Redis returns valid JSON bytes → returns the parsed dict (verify `.decode("utf-8")` is applied)
- `redis.get` raises → returns `None`, error logged
- JSON-decode error → returns `None`, error logged

### `check_task_lock_and_exit()` — `__init__.py:641`

Patch `is_task_locked` and `builtins.exit`.

- No lock → no exit, returns `None`
- Locked, all fields present → logs the user/timestamp/reason, then calls `exit(1)`
- Locked, `reason=None` → `Reason:` line not logged
- Locked, missing `user`/`timestamp` keys → defaults `"unknown"` used in log

## Mocking hints

- For `RedisSemaphore`, patch `time.time` to return a controlled sequence so the `while time.time() < end_time` loop terminates predictably:
  ```python
  mocker.patch("osism.utils.time.time", side_effect=[0, 0.001, 11])
  mocker.patch("osism.utils.time.sleep")
  ```
- For `create_redlock`, the production code does `from pottery import Redlock` inside the function, so patch `pottery.Redlock`.
- `set_task_lock` JSON payload includes a `datetime.now().isoformat()` value — match it with a regex (`\d{4}-\d{2}-\d{2}T...`) instead of an exact string, or freeze time.
- The pottery logger setting is mutable global state — verify the call but reset the level in a fixture if you care about test isolation.

## Definition of Done

- [ ] `tests/unit/utils/test_init_locks.py` created
- [ ] `pytest --cov=osism.utils` for the targeted helpers ≥ 90 %
- [ ] `pipenv run pytest tests/unit/utils/test_init_locks.py` passes locally

## Dependencies

berendt:
Implements #2230.

Adds `tests/unit/utils/test_init_locks.py`, the third companion to
`test_init_connections.py` and `test_init_task_output.py`, covering the
concurrency primitives and global task-lock helpers in
`osism/utils/__init__.py`. No production code is changed — this is a
pure test addition. 40 test functions (41 cases) following the
established `mocker` / `loguru_logs` conventions of the existing suite.

## Change set (commit order)

1. **Add RedisSemaphore unit tests** — `__init__` attribute storage and
   the unconditional `semaphore:` key prefix (including the
   already-prefixed input → double prefix); `acquire()` first-try
   success, full-window timeout returning `False`, per-iteration
   `zremrangebyscore` cleanup, the default-10s and explicit-timeout
   fallbacks, and a fresh `uuid.uuid4()` per call; `release()`
   idempotency; and the `__enter__`/`__exit__` context-manager protocol
   (returns `self`, raises `TimeoutError` containing the key, always
   releases, never swallows exceptions). `time.time`/`time.sleep`/
   `uuid.uuid4` are patched so the acquisition loop is deterministic.
2. **Add create_redlock and create_netbox_semaphore unit tests** —
   `create_redlock` returns the `pottery.Redlock` built with
   `key`/`masters={redis}`/`auto_release_time` (default `3600` and a
   custom value), forces the `"pottery"` logger to `CRITICAL`, and
   suppresses construction-time stdout/stderr (asserted via `capsys`).
   `create_netbox_semaphore` falls back to
   `settings.NETBOX_MAX_CONNECTIONS` or takes an explicit value, builds
   a per-URL md5-derived key with `timeout=30` and the shared redis
   client, and produces a stable key per URL.
3. **Add task-lock helper unit tests** — `set_task_lock`,
   `remove_task_lock`, `is_task_locked` and `check_task_lock_and_exit`,
   including the operator-user/reason fallbacks, the JSON payload shape
   (regex-matched ISO timestamp), the `.decode("utf-8")` path, and the
   error-logging branches. `builtins.exit` and `is_task_locked` are
   patched for `check_task_lock_and_exit`.

## Notes for the reviewer

- `pottery.Redlock` is patched on the `pottery` module (not imported
  into `osism.utils`) because production imports it lazily inside
  `create_redlock` — matching the issue's mocking hint.
- `logging.getLogger` is patched in the `create_redlock` tests so they
  never mutate the process-global `pottery` logger level (test
  isolation), addressing the issue's note about that mutable state.
- `freezegun` is not a dev dependency, so the `set_task_lock` timestamp
  is matched with a regex (`^\d{4}-\d{2}-\d{2}T...`) rather than frozen,
  as the issue allows.

## Local verification

Per repository practice the test suite, `flake8`, `mypy` and
`python-black` run only in the pipenv venv / Zuul CI (the dev tools and
`pottery` are not installed in the base interpreter), so they were not
run locally. `python -m py_compile` passes. The Zuul
`python-osism-unit-tests` job will run the suite. The tests exercise
every branch of the targeted helpers, satisfying the ≥90 % coverage
criterion for them.

Closes #2230

Assisted-by: Claude:claude-opus-4-8

### DIFF
--- a/tests/unit/utils/test_init_locks.py
+++ b/tests/unit/utils/test_init_locks.py
@@ -406,3 +406,223 @@ def test_create_netbox_semaphore_same_url_identical_key(mocker):
     sem_b = utils_pkg.create_netbox_semaphore("https://same.example")
 
     assert sem_a.key == sem_b.key
+
+
+# ---------------------------------------------------------------------------
+# set_task_lock
+# ---------------------------------------------------------------------------
+
+_TASK_LOCK_KEY = "osism:task_lock"
+_ISO_TIMESTAMP_RE = re.compile(r"^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}")
+
+
+def _captured_lock_payload(mock_r):
+    """Return the JSON payload ``set_task_lock`` wrote to redis as a dict."""
+    mock_r.set.assert_called_once()
+    args = mock_r.set.call_args.args
+    assert args[0] == _TASK_LOCK_KEY
+    return json.loads(args[1])
+
+
+def test_set_task_lock_user_none_falls_back_to_operator_user(mocker):
+    mock_r = mocker.MagicMock()
+    mocker.patch("osism.utils._init_redis", return_value=mock_r)
+    mocker.patch("osism.utils.settings.OPERATOR_USER", "operator-x")
+
+    assert utils_pkg.set_task_lock(user=None) is True
+
+    assert _captured_lock_payload(mock_r)["user"] == "operator-x"
+
+
+def test_set_task_lock_explicit_user_used_directly(mocker):
+    mock_r = mocker.MagicMock()
+    mocker.patch("osism.utils._init_redis", return_value=mock_r)
+    mocker.patch("osism.utils.settings.OPERATOR_USER", "operator-x")
+
+    assert utils_pkg.set_task_lock(user="alice") is True
+
+    assert _captured_lock_payload(mock_r)["user"] == "alice"
+
+
+def test_set_task_lock_reason_none_stored_as_null(mocker):
+    mock_r = mocker.MagicMock()
+    mocker.patch("osism.utils._init_redis", return_value=mock_r)
+    mocker.patch("osism.utils.settings.OPERATOR_USER", "operator-x")
+
+    utils_pkg.set_task_lock(reason=None)
+
+    assert _captured_lock_payload(mock_r)["reason"] is None
+
+
+def test_set_task_lock_payload_contents(mocker):
+    mock_r = mocker.MagicMock()
+    mocker.patch("osism.utils._init_redis", return_value=mock_r)
+    mocker.patch("osism.utils.settings.OPERATOR_USER", "operator-x")
+
+    utils_pkg.set_task_lock(user="alice", reason="maintenance")
+
+    payload = _captured_lock_payload(mock_r)
+    assert payload["locked"] is True
+    assert payload["user"] == "alice"
+    assert payload["reason"] == "maintenance"
+    assert _ISO_TIMESTAMP_RE.match(payload["timestamp"])
+
+
+def test_set_task_lock_redis_failure_returns_false(mocker, loguru_logs):
+    mock_r = mocker.MagicMock()
+    mock_r.set.side_effect = RuntimeError("redis down")
+    mocker.patch("osism.utils._init_redis", return_value=mock_r)
+    mocker.patch("osism.utils.settings.OPERATOR_USER", "operator-x")
+
+    assert utils_pkg.set_task_lock(user="alice") is False
+
+    errors = [r["message"] for r in loguru_logs if r["level"] == "ERROR"]
+    assert any("Failed to set task lock" in m for m in errors)
+    assert any("redis down" in m for m in errors)
+
+
+# ---------------------------------------------------------------------------
+# remove_task_lock
+# ---------------------------------------------------------------------------
+
+
+def test_remove_task_lock_deletes_key_returns_true(mocker):
+    mock_r = mocker.MagicMock()
+    mocker.patch("osism.utils._init_redis", return_value=mock_r)
+
+    assert utils_pkg.remove_task_lock() is True
+
+    mock_r.delete.assert_called_once_with(_TASK_LOCK_KEY)
+
+
+def test_remove_task_lock_failure_returns_false(mocker, loguru_logs):
+    mock_r = mocker.MagicMock()
+    mock_r.delete.side_effect = RuntimeError("redis down")
+    mocker.patch("osism.utils._init_redis", return_value=mock_r)
+
+    assert utils_pkg.remove_task_lock() is False
+
+    errors = [r["message"] for r in loguru_logs if r["level"] == "ERROR"]
+    assert any("Failed to remove task lock" in m for m in errors)
+
+
+# ---------------------------------------------------------------------------
+# is_task_locked
+# ---------------------------------------------------------------------------
+
+
+def test_is_task_locked_returns_none_when_unset(mocker):
+    mock_r = mocker.MagicMock()
+    mock_r.get.return_value = None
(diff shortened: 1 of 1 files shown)


tests/unit/utils/test_init_locks.py
```diff
@@ -0,0 +1,606 @@
...
+    redis.zcard.return_value = 5
+    sem = utils_pkg.RedisSemaphore(redis, "job", 5, timeout=10)
+    # now=0; loop checks 1 and 2 enter (two iterations); 11 exits.
+    mocker.patch("osism.utils.time.time", side_effect=[0, 1, 2, 11])
+    mocker.patch("osism.utils.time.sleep")
+
+    assert sem.acquire() is False
+
+    assert redis.zremrangebyscore.call_count == 2
+    redis.zremrangebyscore.assert_called_with("semaphore:job", 0, -60)
```
ideaship: This assertion pins the eviction threshold at a constant `-60` across both loop iterations, which locks in a production quirk: `acquire()` captures `now` once before the retry loop (`utils/__init__.py:142`), so the cleanup boundary `now - 60` (line 147) never advances while a caller waits.

The boundary only governs reaping of *ghost* slots — a worker that acquired the semaphore and died before `release()` ran (OOM, container restart), leaving its score stuck in the sorted set. With the production config (maxsize=`NETBOX_MAX_CONNECTIONS` default 5, timeout=30s, 60s eviction window), the bug surfaces when: NetBox writes are sustained at full concurrency, one ghost slot is silently holding capacity, and a new `_update_netbox_device_field` call (`tasks/netbox.py:31`) starts waiting while that ghost is in its 30–60s age window. A per-iteration recompute would evict the ghost the moment it ages past 60s and let the waiter take the freed slot; with the frozen boundary the waiter's threshold never reaches it, so `acquire()` times out and `__enter__` raises `TimeoutError` (`utils/__init__.py:167`) — before the `try/except` inside the `with`, so the device update fails spuriously even though capacity had freed up.

The correct fix is to recompute `now` inside the loop — at which point this test's `assert_called_with(..., 0, -60)` would fail (it would see `1-60`, `2-60`, …). As written, the test actively resists that fix. Either assert the advancing threshold, or add an explicit comment that the frozen boundary is a known limitation being codified rather than endorsed.

ideaship (changes requested):

ideaship (approved):

PR #2352, opened 2026-06-12
Add unit tests for osism/utils/rabbitmq.py

### LINKED ISSUE #2232: Unit tests for osism/utils/rabbitmq.py
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul infrastructure). Part of **Tier 4** (#2199). `osism/utils/rabbitmq.py` (173 LOC) resolves RabbitMQ node IPs by combining `ansible-inventory` calls, ansible-facts cached in Redis, and Jinja2 template resolution from `internal_interface`. Includes a small helper for loading the encrypted RabbitMQ password.

## Scope

Add `tests/unit/utils/test_rabbitmq.py` covering both functions in `osism/utils/rabbitmq.py`.

### `get_rabbitmq_node_addresses()` — `rabbitmq.py:13`

Patch:

- `osism.utils.rabbitmq.subprocess.check_output`
- `osism.utils.rabbitmq.get_inventory_path`
- `osism.utils.rabbitmq.get_hosts_from_inventory`
- `osism.utils.rabbitmq.utils.redis` (the lazy attribute on `osism.utils`) — easiest via `mocker.patch("osism.utils.rabbitmq.utils.redis", ...)` to a `MagicMock`

**Inventory & host discovery**

- Inventory call returns valid JSON, two hosts → both processed in alphabetical order (the function calls `.sort()` on the host list)
- `get_hosts_from_inventory` returns `[]` → returns `None`, error logged "No hosts found in rabbitmq group"
- `subprocess.check_output` (first call, group listing) raises `CalledProcessError` → returns `None`, error logged
- First call returns invalid JSON → returns `None` (`json.JSONDecodeError`), error logged
- Outer generic exception → returns `None`, error logged

**Per-host resolution**

- `redis.get("ansible_facts<host>")` returns `None` → host skipped, error logged, processing continues to next host
- Hostvars query returns no `internal_interface` → host skipped, error logged
- `internal_interface` is a literal string (`"eth0"`) → used directly without templating
- `internal_interface` is a Jinja template (`"{{ ansible_local.testbed_network_devices.management }}"`) → resolved against the facts dict via dotted path traversal
- Template path resolves to a non-string (`None`, dict, int) → host skipped, error logged
- Template path traversal hits a non-dict mid-walk (e.g. `ansible_local.foo` is a string) → resolves to `None`, host skipped
- Interface name normalization: `"eth0.100"` → `"ansible_eth0_100"`, `"eth-0"` → `"ansible_eth_0"`
- Facts missing the normalized interface key → host skipped, error logged
- Interface facts present but no `ipv4` key → host skipped, error logged
- `ipv4` present but `address` empty/missing → host skipped, error logged
- Happy path: returns `[("10.0.0.5", "host1"), ("10.0.0.6", "host2")]` in alphabetical-host order

**Aggregate results**

- All hosts skipped → returns `None`, error logged "Could not retrieve address for any RabbitMQ node"
- At least one address → returns the list (no error)

### `load_rabbitmq_password()` — `rabbitmq.py:139`

Patch `os.path.exists` and `osism.tasks.conductor.utils.load_yaml_file` (or import path used by the production code: `from osism.tasks.conductor.utils import load_yaml_file` — patch at the call site).

- File missing → returns `None`, error logged
- `load_yaml_file` returns `None` → returns `None`, error logged "Empty or invalid secrets file"
- `load_yaml_file` returns a non-dict (e.g. a list) → returns `None`, error logged
- Dict without `rabbitmq_password` key → returns `None`, error logged
- Dict with `rabbitmq_password="  hunter2  "` → returns `"hunter2"` (stripped)
- `rabbitmq_password=42` (int) → coerced via `str(...).strip()` → `"42"`
- `load_yaml_file` raises → returns `None`, error logged

## Mocking hints

- The function makes **two** `subprocess.check_output` calls per host (one for the group, one for the host vars). Use `side_effect` with a list of payloads:
  ```python
  mocker.patch("osism.utils.rabbitmq.subprocess.check_output", side_effect=[
      json.dumps({"rabbitmq": {"hosts": ["host1", "host2"]}, ...}).encode(),
      json.dumps({"internal_interface": "eth0"}).encode(),
      json.dumps({"internal_interface": "eth0"}).encode(),
  ])
  ```
- `redis.get` is a `MagicMock`; set its `side_effect` per host or use `.return_value=json.dumps(facts).encode()`.
- For the Jinja template path, build facts as nested dicts:
  ```python
  facts = {"ansible_local": {"testbed_network_devices": {"management": "eth1"}}}
  ```
- `get_inventory_path` returns a string path; the function is covered in #2194, so just stub it here.
- `RABBITMQ_USER` is a module constant — verify its value (`"openstack"`) in one quick test.

## Definition of Done

- [ ] `tests/unit/utils/test_rabbitmq.py` created
- [ ] `pytest --cov=osism.utils.rabbitmq` shows ≥ 95 %
- [ ] `pipenv run pytest tests/unit/utils/test_rabbitmq.py` passes locally

## Dependencies

berendt:
Closes #2232

Adds `tests/unit/utils/test_rabbitmq.py` covering both functions in
`osism/utils/rabbitmq.py`, plus one small robustness fix in
`get_rabbitmq_node_addresses` that the tests pin down.

## Production change

**Skip RabbitMQ hosts that fail address resolution.** Previously a
failure while resolving a single host (corrupt facts in the Redis
cache, a failing `ansible-inventory --host` call, non-JSON hostvars or
an `internal_interface` value that is not a string) raised out of the
per-host loop and made the whole function return `None`, discarding the
addresses already collected for other hosts — while the expected data
problems right next to these spots (missing facts, missing interface,
missing IPv4 address) only skipped the affected host. Per-host failures
are now caught, logged and skipped like the other skip branches; the
function still returns `None` if no host resolves at all.

## What is covered

Tests are grouped into one class per function
(`TestGetRabbitmqNodeAddresses`, `TestLoadRabbitmqPassword`), with
factory fixtures (`setup_addresses`, `setup_password`) wiring up the
collaborators.

**`get_rabbitmq_node_addresses()`**

- *Inventory & host discovery*: two hosts processed in alphabetical
  order (happy path returning `[("10.0.0.5","host1"),("10.0.0.6","host2")]`);
  empty rabbitmq group → `None`; first `subprocess.check_output` raising
  `CalledProcessError`; invalid JSON from the group listing
  (`JSONDecodeError`); outer generic exception. The inventory
  invocations are pinned: `--limit rabbitmq` is the only thing scoping
  the listing to the group, and the hostvars lookup must use the
  non-minified inventory (`prefer_minified=False`) because the minified
  variant omits hostvars.
- *Per-host resolution*: cache miss skips the host and continues to the
  next; partial-success tests (host1 resolves, host2 fails with corrupt
  cached facts / failing hostvars query / non-JSON hostvars / non-string
  `internal_interface`) pin that only the failing host is dropped;
  missing `internal_interface`; literal interface name used directly;
  Jinja2 template resolved via dotted-path traversal of the facts;
  template resolving to a non-string (`None` / dict / int); traversal
  hitting a non-dict mid-walk; interface-name normalization
  (`eth0.100` → `ansible_eth0_100`, `eth-0` → `ansible_eth_0`); missing
  normalized interface key; interface without `ipv4`; `ipv4` without an
  `address`.
- *Aggregate*: all hosts skipped → `None`; at least one address → the
  list is returned with no error logged.

**`load_rabbitmq_password()`** — missing file; empty / invalid / non-dict
secrets; missing `rabbitmq_password` key; whitespace stripping; int
coercion; loader raising.

Plus a check that the `RABBITMQ_USER` module constant is `"openstack"`.

## Notes for the reviewer

- **Redis mocking deviates from the issue's hint.** The issue suggested
  `mocker.patch("osism.utils.rabbitmq.utils.redis", ...)`, but that target
  does not resolve: `redis` is a lazy `__getattr__` attribute on the
  `osism.utils` *package*, and the function reads it via a function-local
  `from osism import utils`, so the `rabbitmq` module never gains a `utils`
  global (`AttributeError: module 'osism.utils.rabbitmq' has no attribute
  'utils'`). Instead the cached attribute is seeded on the package
  namespace with `mocker.patch.dict(osism.utils.__dict__, {"redis": ...})`,
  which also bypasses the lazy initializer that would otherwise open a real
  Redis connection. Reading the cache key the function builds is asserted in
  the happy-path test.
- `load_yaml_file` and `os.path.exists` are patched at the call site, as the
  issue suggests.
- Tests follow the existing `mocker` + `loguru_logs` conventions
  (see `tests/unit/utils/test_init_connections.py`).

## Verification

- `black --check` and `flake8` clean on both files; unit tests run in CI.

### DIFF
--- a/osism/utils/rabbitmq.py
+++ b/osism/utils/rabbitmq.py
@@ -40,84 +40,92 @@ def get_rabbitmq_node_addresses():
 
         node_addresses = []
         for host in rabbitmq_hosts:
-            # Get ansible facts from Redis cache
-            facts_data = utils.redis.get(f"ansible_facts{host}")
-            if not facts_data:
-                logger.error(f"No ansible facts found in cache for {host}")
-                continue
+            # A failure on one host must not discard the addresses already
+            # collected for the other nodes.
+            try:
+                # Get ansible facts from Redis cache
+                facts_data = utils.redis.get(f"ansible_facts{host}")
+                if not facts_data:
+                    logger.error(f"No ansible facts found in cache for {host}")
+                    continue
 
-            facts = json.loads(facts_data)
+                facts = json.loads(facts_data)
 
-            # Get hostvars for this host to find internal_interface
-            hostvar_inventory_path = get_inventory_path(
-                "/ansible/inventory/hosts.yml", prefer_minified=False
-            )
-            result = subprocess.check_output(
-                f"ansible-inventory -i {hostvar_inventory_path} --host {host}",
-                shell=True,
-                stderr=subprocess.DEVNULL,
-            )
-            hostvars = json.loads(result)
+                # Get hostvars for this host to find internal_interface
+                hostvar_inventory_path = get_inventory_path(
+                    "/ansible/inventory/hosts.yml", prefer_minified=False
+                )
+                result = subprocess.check_output(
+                    f"ansible-inventory -i {hostvar_inventory_path} --host {host}",
+                    shell=True,
+                    stderr=subprocess.DEVNULL,
+                )
+                hostvars = json.loads(result)
 
-            internal_interface_raw = hostvars.get("internal_interface")
-            if not internal_interface_raw:
-                logger.error(f"internal_interface not found in hostvars for {host}")
-                continue
+                internal_interface_raw = hostvars.get("internal_interface")
+                if not internal_interface_raw:
+                    logger.error(f"internal_interface not found in hostvars for {host}")
+                    continue
 
-            # Resolve Jinja2 template if present (e.g., "{{ ansible_local.testbed_network_devices.management }}")
-            internal_interface = internal_interface_raw
-            template_match = re.match(r"\{\{\s*(.+?)\s*\}\}", internal_interface_raw)
-            if template_match:
-                path = template_match.group(1).strip()
-                parts = path.split(".")
-                value = facts
-                for part in parts:
-                    if isinstance(value, dict):
-                        value = value.get(part)
+                # Resolve Jinja2 template if present (e.g., "{{ ansible_local.testbed_network_devices.management }}")
+                internal_interface = internal_interface_raw
+                template_match = re.match(
+                    r"\{\{\s*(.+?)\s*\}\}", internal_interface_raw
+                )
+                if template_match:
+                    path = template_match.group(1).strip()
+                    parts = path.split(".")
+                    value = facts
+                    for part in parts:
+                        if isinstance(value, dict):
+                            value = value.get(part)
+                        else:
+                            value = None
+                            break
+                    if value and isinstance(value, str):
+                        internal_interface = value
                     else:
-                        value = None
-                        break
-                if value and isinstance(value, str):
-                    internal_interface = value
-                else:
+                        logger.error(
+                            f"Could not resolve template '{internal_interface_raw}' from facts for {host}"
+                        )
+                        continue
+
+                logger.debug(f"Internal interface for {host}: {internal_interface}")
+
+                # Look for the interface in ansible facts
+                # Interface names with special chars are normalized (e.g., eth0.100 -> ansible_eth0_100)
+                normalized_interface = internal_interface.replace(".", "_").replace(
+                    "-", "_"
+                )
+                interface_key = f"ansible_{normalized_interface}"
+
+                interface_facts = facts.get(interface_key)
+                if not interface_facts:
                     logger.error(
-                        f"Could not resolve template '{internal_interface_raw}' from facts for {host}"
+                        f"Interface {internal_interface} ({interface_key}) not found in ansible facts for {host}"
                     )
                     continue
 
-            logger.debug(f"Internal interface for {host}: {internal_interface}")
+                # Get IPv4 address
+                ipv4_info = interface_facts.get("ipv4")
+                if not ipv4_info:
+                    logger.error(
+                        f"No IPv4 address found for interface {internal_interface} on {host}"
+                    )
+                    continue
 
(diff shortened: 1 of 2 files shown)


ideaship (commented):
**Per-host failures abort the whole discovery and discard partial results — is that intended?**

Inside the per-host loop, `json.loads(facts_data)` (corrupt cache), the `--host` `subprocess.check_output`, `json.loads(result)` (non-JSON), and `re.match(...)` on a truthy non-string `internal_interface` are all uncaught locally. Any one of them propagates to the function-level handlers and returns `None`, throwing away every address already collected for earlier hosts. That's a different blast radius than the `continue` skip-branches right next to them, which only drop the offending host. The existing `CalledProcessError`/`JSONDecodeError` tests only hit the *first* (group-listing) call, so this asymmetry isn't exercised — which also means the commit message's "every error and skip branch" / 100% coverage framing is a bit optimistic here. Mostly I'd want to know: is abort-all-after-partial-success the behavior you want, or should these `continue` like the skips? Whatever the answer, a partial-success test (host1 resolves, host2 fails) would pin it.

**Optional:** two production arguments carry correctness weight but aren't asserted, because their collaborators are mocked — so a future regression flipping either would pass silently:

- the hostvars lookup deliberately uses `prefer_minified=False` (the minified inventory omits hostvars, so `internal_interface` wouldn't resolve)
- `--limit rabbitmq` is the *only* thing scoping results to the group (`get_hosts_from_inventory` does no group filtering)

Each is a one-line `call_args` assertion if you think they're worth pinning. Not blocking.

ideaship (approved):

PR #2369, opened 2026-06-12
Add lightweight integration tests for Celery/Redis basics in CI

### LINKED ISSUE #2368: Add lightweight integration tests for Celery/Redis basics in CI
## Motivation

The task processing core of python-osism (Celery broker, queue routing, workers, result backend, Redis streams for task output, distributed locks) is currently not covered by any test in CI. The `check` pipeline only runs linters, unit tests (`tests/unit/`), and container image builds. A lightweight integration test job can cover these paths end-to-end with very little infrastructure: a single Redis container and a Celery worker started from the same venv as the tests.

## Prerequisite: make the broker URL configurable

`broker_url` and `result_backend` are hardcoded to `redis://redis` in `osism/tasks/__init__.py` (also duplicated in `osism/commands/service.py` for `beat` and `flower`). In contrast, the direct Redis access in `osism.utils._init_redis()` already honors `REDIS_HOST` / `REDIS_PORT` / `REDIS_DB` from `osism/settings.py`.

The Celery `Config` should derive its URLs the same way, e.g.:

```python
broker_url = os.environ.get(
    "CELERY_BROKER_URL",
    f"redis://{settings.REDIS_HOST}:{settings.REDIS_PORT}/{settings.REDIS_DB}",
)
result_backend = os.environ.get("CELERY_RESULT_BACKEND", broker_url)
```

This is backwards compatible (the default of `REDIS_HOST` is `redis`) and removes the duplicated literals in `service.py`.

Side finding to fix along the way: `task_track_started = (True,)` is a tuple instead of a bool — it only works because a non-empty tuple is truthy.

## Proposed setup

New Zuul job `python-osism-integration-tests` (analogous to the existing jobs, `pre-run: playbooks/pre.yml`, new `playbooks/test-integration.yml`), added to the `check` and `periodic-daily` pipelines:

1. Start a Redis container on the Zuul node (`ensure-docker` role from zuul-jobs, then e.g. `docker run -d -p 6379:6379 redis:7-alpine`).
2. Install the package with pipenv exactly like the unit test job does.
3. Run `pytest tests/integration` with `REDIS_HOST=localhost`.

The Celery worker is started as a session-scoped pytest fixture from the same venv:

```python
@pytest.fixture(scope="session")
def celery_worker():
    proc = subprocess.Popen(
        ["celery", "-A", "osism.tasks.ansible", "worker",
         "-n", "ci-worker", "-Q", "osism-ansible", "-c", "1"],
        env={**os.environ, "GATHER_FACTS_SCHEDULE": "0"},
    )
    # wait until app.control.inspect().ping() succeeds, yield, then terminate()
```

`GATHER_FACTS_SCHEDULE=0` prevents registration of the periodic `gather_facts` task (which would try to run `/run.sh` in containers that do not exist in CI). Deliberately only the `ansible` Celery app is used as worker: it has no import-time dependency on NetBox/OpenStack/ansible-core, unlike `conductor` and friends, which would require extras and external services.

## Test cases (`tests/integration/`)

1. **Celery round-trip** (the core): dispatch `osism.tasks.ansible.noop.delay()` and assert `result.get(timeout=60) is True`. This validates broker, queue routing (`osism-ansible`), worker, and result backend in a single test.
2. **Worker visibility**: `app.control.inspect().ping()` — the same mechanism as `osism get status workers`; optionally invoke the CLI directly for end-to-end coverage.
3. **Redis streams path**: write via `push_task_output` / `finish_task_output` with a test task ID, assert that `fetch_task_output` returns the output and return code. This is the mechanism `osism apply` uses to stream logs and is testable with Redis alone.
4. **Locking primitives**: `create_redlock` acquire/release and `set_task_lock` / `is_task_locked` / `remove_task_lock`.

## Notes

- The unit test job currently runs a bare `pytest`, which would also collect `tests/integration/`. The integration tests need an `integration` marker (skipped when Redis is not reachable) or the unit job must be restricted to `pytest tests/unit`.
- Expected job runtime: ~2-3 minutes.

## Possible follow-ups (out of scope here)

- Compose-based variant using the container image already built in `check` (services `redis`, worker, optionally `osism service api`), which would additionally cover the image and entrypoint.
- API tests via `uvicorn osism.api:app` + httpx, beat scheduling, `revoke_task`.

berendt:
Closes #2368.

Adds a lightweight integration-test job that covers the Celery/Redis
task-processing core (broker, queue routing, worker, result backend, Redis
streams for task output, distributed locks) end-to-end in CI, plus the
prerequisite that makes the broker URL configurable.

### 1. Derive Celery broker URL from settings and environment
`osism/tasks/__init__.py`, `osism/commands/service.py`

- `Config.broker_url` / `Config.result_backend` were hardcoded to
  `redis://redis`, with the same literals duplicated in the `beat` and
  `flower` service commands. They now derive from `REDIS_HOST` /
  `REDIS_PORT` / `REDIS_DB` (the same settings `osism.utils._init_redis`
  already honors), overridable via `CELERY_BROKER_URL` /
  `CELERY_RESULT_BACKEND`. The default of `REDIS_HOST` is `redis`, so this
  is backwards compatible.
- `service.py` now references `Config.broker_url` (single source of truth)
  for `beat`/`flower`, removing the duplicated literals.
- Fixed `task_track_started`, which was the tuple `(True,)` instead of the
  intended bool (it only worked because a non-empty tuple is truthy).

This is what lets the tests point Celery at a local Redis via
`REDIS_HOST=localhost`.

### 2. Add Celery/Redis integration tests
`tests/integration/`, `setup.cfg`, `README.md`

- `test_celery.py` — **round-trip**: `ansible.noop.delay()` →
  `result.get(timeout=60) is True` (broker + `osism-ansible` routing +
  worker + result backend); **worker visibility**: a fresh `Celery`
  client configured from `Config` sees the worker via
  `inspect().ping()` (the mechanism behind `osism get status workers`).
- `test_redis_streams.py` — `push_task_output` / `finish_task_output` /
  `fetch_task_output` round-trip (the path `osism apply` uses to stream
  logs).
- `test_locking.py` — `create_redlock` acquire/release and
  `set_task_lock` / `is_task_locked` / `remove_task_lock`.
- `conftest.py` starts the worker as a session-scoped fixture from the
  same venv, running only the `ansible` Celery app (no import-time
  dependency on NetBox/OpenStack/ansible-core) with
  `GATHER_FACTS_SCHEDULE=0` so the periodic `gather_facts` task is not
  registered. It polls `inspect().ping()` until the worker is ready and
  terminates it on teardown.
- Tests carry an `integration` marker (registered in `setup.cfg` for the
  `--strict-markers` run) and are **skipped when Redis is not reachable**,
  so a local `pytest` is unaffected.

### 3. Add Zuul integration-test job
`.zuul.yaml`, `playbooks/test-integration.yml`

- New `python-osism-integration-tests` job added to the `check` and
  `periodic-daily` pipelines. It reuses `playbooks/pre.yml` (which already
  sets up Docker via `ensure-docker`), starts a `redis:7-alpine`
  container, installs the package with pipenv exactly like the unit-test
  job, and runs `pytest tests/integration` with `REDIS_HOST=localhost`.

## Notes / deviations

- The unit-test job is already restricted to `tests/unit` via
  `testpaths = tests/unit` in `setup.cfg`, so a bare `pytest` never
  collects `tests/integration` — the issue's first concern is already
  satisfied. The `integration` marker + Redis-reachability skip are still
  added (the marker is required by `--strict-markers`, and the skip keeps
  `pytest tests/integration` green locally without Redis).
- No `CHANGELOG.md` entry: this repo maintains the changelog in batched,
  post-release commits that map merged PR numbers to release versions, so
  per-PR entries are out of convention here.
- Follow-ups listed in the issue (compose-based variant, API/beat/revoke
  coverage) are intentionally out of scope.

## Local verification

- `black --check` — clean (6 files)
- `flake8` — clean
- `pytest tests/integration` — not run locally (no Redis); exercised by
  the new CI job.

### DIFF
--- a/osism/tasks/__init__.py
+++ b/osism/tasks/__init__.py
@@ -13,7 +13,7 @@ from pathlib import Path
 
 from loguru import logger
 
-from osism import utils
+from osism import settings, utils
 
 # Regex pattern for extracting hosts from Ansible output
 HOST_PATTERN = re.compile(r"^(ok|changed|failed|skipping|unreachable):\s+\[([^\]]+)\]")
@@ -23,11 +23,14 @@ class Config:
     broker_connection_retry_on_startup = True
     enable_utc = True
     enable_ironic = os.environ.get("ENABLE_IRONIC", "True")
-    broker_url = "redis://redis"
-    result_backend = "redis://redis"
+    broker_url = os.environ.get(
+        "CELERY_BROKER_URL",
+        f"redis://{settings.REDIS_HOST}:{settings.REDIS_PORT}/{settings.REDIS_DB}",
+    )
+    result_backend = os.environ.get("CELERY_RESULT_BACKEND", broker_url)
     task_create_missing_queues = True
     task_default_queue = "default"
-    task_track_started = (True,)
+    task_track_started = True
     task_routes = {
         "osism.tasks.ansible.*": {"queue": "osism-ansible"},
         "osism.tasks.ceph.*": {"queue": "ceph-ansible"},
--- a/playbooks/test-integration.yml
+++ b/playbooks/test-integration.yml
@@ -1,0 +1,43 @@
+---
+- name: Run integration tests
+  hosts: all
+
+  vars:
+    python_venv_dir: /tmp/venv
+
+  tasks:
+    - name: Start Redis container
+      ansible.builtin.shell:
+        executable: /bin/bash
+        cmd: |
+          set -e
+          set -o pipefail
+          set -x
+
+          docker run -d --name redis -p 6379:6379 redis:7-alpine
+      changed_when: true
+
+    - name: Install dependencies
+      ansible.builtin.shell:
+        executable: /bin/bash
+        chdir: "{{ zuul.project.src_dir }}"
+        cmd: |
+          set -e
+          set -o pipefail
+          set -x
+
+          {{ python_venv_dir }}/bin/pipenv install --dev --deploy
+          {{ python_venv_dir }}/bin/pipenv run pip install .
+
+    - name: Run pytest
+      ansible.builtin.shell:
+        executable: /bin/bash
+        chdir: "{{ zuul.project.src_dir }}"
+        cmd: |
+          set -e
+          set -o pipefail
+          set -x
+
+          {{ python_venv_dir }}/bin/pipenv run pytest tests/integration
+      environment:
+        REDIS_HOST: localhost
--- a/playbooks/test-integration.yml
+++ b/playbooks/test-integration.yml
@@ -17,6 +17,24 @@
           docker run -d --name redis -p 6379:6379 redis:7-alpine
       changed_when: true
 
+    - name: Wait for Redis to be ready
+      ansible.builtin.shell:
+        executable: /bin/bash
+        cmd: |
+          set -e
+          set -o pipefail
+          set -x
+
+          for i in $(seq 1 30); do
+            if docker exec redis redis-cli ping | grep -q PONG; then
+              exit 0
+            fi
+            sleep 1
+          done
+          echo "Redis did not become ready within 30s" >&2
+          exit 1
+      changed_when: false
+
     - name: Install dependencies
       ansible.builtin.shell:
         executable: /bin/bash
@@ -41,3 +59,6 @@
           {{ python_venv_dir }}/bin/pipenv run pytest tests/integration
       environment:
         REDIS_HOST: localhost
+        # Fail the session if Redis is unreachable instead of skipping every
+        # test, so a Redis-startup failure turns the job red rather than green.
+        OSISM_REQUIRE_REDIS: "1"
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -25,7 +25,26 @@ WORKER_QUEUE = "osism-ansible"
 # Celery treats a ``-n`` value without ``@`` as the host part (yielding
 # ``celery@ci-worker``), so the node name must be given explicitly.
 WORKER_NAME = "ci-worker@%h"
+# Node-name prefix the worker registers under (``ci-worker@<hostname>``). Used
+# to gate readiness on *this* worker rather than any worker on the broker.
+WORKER_NODE_PREFIX = WORKER_NAME.split("@", 1)[0] + "@"
 WORKER_BOOT_TIMEOUT = 60
+
+
+def _require_redis():
+    """Return ``True`` when an unreachable Redis must fail the session.
+
+    Set ``OSISM_REQUIRE_REDIS=1`` (as the CI job does) to turn the
+    skip-when-unreachable behaviour into a hard failure, so a Redis-startup
+    problem surfaces as a red job instead of an all-skipped green one. Left
+    unset locally, the suite keeps skipping when Redis is not running.
+    """
+    return os.environ.get("OSISM_REQUIRE_REDIS", "").lower() not in (
+        "",
+        "0",
+        "false",
+        "no",
+    )
 
 
 def _redis_reachable():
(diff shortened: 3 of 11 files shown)


playbooks/test-integration.yml
```diff
@@ -0,0 +1,43 @@
...
+    - name: Run pytest
+      ansible.builtin.shell:
+        executable: /bin/bash
+        chdir: "{{ zuul.project.src_dir }}"
+        cmd: |
+          set -e
+          set -o pipefail
+          set -x
+
+          {{ python_venv_dir }}/bin/pipenv run pytest tests/integration
```
ideaship, line 59: This job can pass green with zero tests executed. `tests/integration/conftest.py` skips all `integration`-marked tests when Redis is unreachable, and pytest exits 0 on all-skipped (reproduced locally as `5 skipped, exit 0`). With no readiness gate after `docker run -d`, any Redis-startup failure here surfaces as a passing job instead of a red one, defeating the gate. Please add (a) a Redis-readiness wait before pytest, and (b) a required-Redis mode for CI, e.g. an `OSISM_REQUIRE_REDIS=1` env lever that makes `conftest.py` fail the session instead of skipping when Redis is unreachable. Set it only in this job so local skip behavior is preserved.

osism/tasks/__init__.py
```diff
@@ -23,11 +23,14 @@ class Config:
     broker_connection_retry_on_startup = True
     enable_utc = True
     enable_ironic = os.environ.get("ENABLE_IRONIC", "True")
-    broker_url = "redis://redis"
-    result_backend = "redis://redis"
+    broker_url = os.environ.get(
```
ideaship, line 26: This is a production-path change with no unit test. It replaced two hardcoded constants with real precedence logic (`CELERY_BROKER_URL` override, and `result_backend` defaulting to `broker_url`), but there is no reference to either env var anywhere under `tests/`; the integration suite only exercises the `REDIS_HOST`-derived fallback. Please add a unit test asserting: (1) `CELERY_BROKER_URL` wins over the derived URL, (2) `CELERY_RESULT_BACKEND` overrides independently, (3) absent `CELERY_RESULT_BACKEND`, `result_backend == broker_url`. Note these are class attributes resolved at import time from import-frozen `settings.REDIS_*`, so the test must `importlib.reload(settings)` then `importlib.reload(osism.tasks)` after setting env; `monkeypatch.setenv` alone will assert the stale cached value and pass for the wrong reason.

osism/tasks/__init__.py
```diff
@@ -23,11 +23,14 @@ class Config:
...
+        "CELERY_BROKER_URL",
+        f"redis://{settings.REDIS_HOST}:{settings.REDIS_PORT}/{settings.REDIS_DB}",
+    )
+    result_backend = os.environ.get("CELERY_RESULT_BACKEND", broker_url)
     task_create_missing_queues = True
     task_default_queue = "default"
-    task_track_started = (True,)
+    task_track_started = True
```
ideaship, line 33: This `(True,)` to `True` fix is correct (the 1-tuple only worked by being truthy), but it is an unrelated bug fix folded into a commit about broker-URL derivation; it should have been its own commit so it is traceable independently. Since the commit's suite advertises end-to-end "task core" coverage, also consider asserting the corrected value: either a direct `Config.task_track_started is True` check, or a STARTED-state transition in the Celery round-trip test.

tests/integration/test_locking.py
```diff
@@ -0,0 +1,41 @@
...
+
+def test_redlock_acquire_and_release():
+    """A Redlock can be acquired and released against the live Redis."""
+    lock = utils.create_redlock(key=f"itest-lock-{uuid.uuid4()}")
+
+    assert lock.acquire(timeout=10)
+    lock.release()
+
+
+def test_task_lock_set_check_remove():
```
ideaship, line 26: This test reads/writes/deletes the fixed global key `osism:task_lock` on db 0 (see `osism/utils/__init__.py:500-554`); it cannot use a per-run UUID like the redlock/stream tests do. Against a disposable CI Redis that is fine, but the README example points `REDIS_HOST=localhost` at a local box, where this would clobber a real operator lock. Please either run the integration suite against a dedicated `REDIS_DB`, or add an explicit note in the README that the suite mutates live state on db 0 and must target a disposable Redis.

tests/integration/conftest.py
```diff
@@ -0,0 +1,117 @@
...
+    )
+
+    try:
+        deadline = time.time() + WORKER_BOOT_TIMEOUT
+        while time.time() < deadline:
+            if proc.poll() is not None:
+                raise RuntimeError(
+                    f"Celery worker exited early with code {proc.returncode}"
+                )
+            if celery_app.control.inspect().ping():
```
ideaship: Readiness breaks on *any* worker pinging the broker, not `ci-worker` specifically. On an isolated CI Redis this is fine, but on a shared local broker it can pass before this worker is up and hide a startup failure. Consider gating on the expected node, e.g. `any(n.startswith("ci-worker@") for n in (celery_app.control.inspect().ping() or {}))`, mirroring what `test_worker_is_visible` already does.

ideaship (changes requested):

ideaship (approved):

PR #2389, opened 2026-06-17
Add 'osism reset facts' command to clear the Ansible fact cache

### LINKED ISSUE #2388: Add a command to reset the Ansible fact cache

## Description

Add an `osism` command that clears the Ansible fact cache so operators can remove stale or incorrect facts. By default the command flushes the entire cache. A parameter restricts the operation to selected hosts or groups when an operator wants to reset only part of it.

The command only clears the cache. It does not start a new fact-gathering run; the cache rebuilds on the next Ansible run that collects facts.

No confirmation prompt is needed. The operation is destructive but low-risk, since the cache is rebuilt automatically.

## Motivation

Operators troubleshooting a deployment need a clean way to drop cached facts. Stale or wrong facts can make Ansible act on outdated information about a host, producing confusing results that are hard to diagnose. There is no direct command to clear them today, so an operator has to locate and remove the cache by hand.

A dedicated command turns this into a single, predictable step. Restricting the reset to specific hosts or groups keeps the rest of the cache intact when only one host misbehaves.

---

_Drafted by [planwerk-review](https://github.com/planwerk/planwerk-review) with Claude_

berendt:
## Summary

Adds `osism reset facts`, a command that clears the Ansible fact cache so
operators can drop stale or incorrect facts during troubleshooting. The
fact cache lives in Redis under `ansible_facts<host>` keys (read by
`osism get facts`, `osism/api.py`, and `check_ansible_facts`); until now
there was no command to clear it.

- By default the command flushes the **entire** cache.
- `-l/--limit <pattern>` restricts the reset to selected hosts or groups,
  resolving the pattern through `ansible-inventory --list --limit`.
- It only clears the cache — it does **not** start a fact-gathering run and
  asks for **no** confirmation. The cache rebuilds on the next Ansible run
  that gathers facts.

Closes #2388

## Changes (in commit order)

1. **`Add 'osism reset facts' command to clear the fact cache`**
   - New `osism/commands/reset.py` with `class Facts(Command)`:
     - No limit → SCAN `ansible_facts*` and batch-delete every key
       (mirrors the scan loop in `osism/utils/__init__.py:check_ansible_facts`).
     - `-l/--limit` → expand the host/group pattern via `ansible-inventory`
       and delete only the matching `ansible_facts<host>` keys
       (mirrors the limit→inventory idiom in `osism/commands/report.py`).
     - Redis key deletion reuses the `utils.redis.delete(...)` idiom from
       `osism/commands/vault.py`.
   - Registers the entry point `reset facts = osism.commands.reset:Facts`
     in `setup.cfg` (no `reset` namespace existed before; no collision).

2. **`Add unit tests for 'osism reset facts'`**
   - New `tests/unit/commands/test_reset.py` covering flush-all,
     empty-cache no-op, Redis error, `--limit` host expansion, and the
     inventory-load failure and timeout paths.

## Notes for reviewers

- **Command name** (`reset facts`) is a design choice the issue did not
  fix. It avoids overloading `sync facts` (which *does* gather) and there
  is no existing `reset` namespace. Renaming is a one-line `setup.cfg`
  change if you prefer e.g. `facts reset`.
- **Redis cache only.** A second, file-based fact cache at
  `/cache/facts/<host>` is read only by `osism/api.py:search_inventory`.
  This command clears the canonical Redis cache that the operator-facing
  CLI uses; clearing the API's file cache is out of scope absent a
  confirmed need.
- The external CLI reference (osism.tech) lives in a separate repo and is
  not updated here; the in-repo doc surface is the command docstring and
  `--limit` help text rendered by `osism reset facts --help`.

## Local verification

- `flake8 osism/commands/reset.py tests/unit/commands/test_reset.py` — pass
- Unit suite (`pytest tests/unit`) runs in CI/Zuul per project convention.

---

_Implemented by [planwerk-review](https://github.com/planwerk/planwerk-review) with Claude:claude-opus-4-8_

### DIFF
--- a/osism/commands/reset.py
+++ b/osism/commands/reset.py
@@ -1,0 +1,104 @@
+# SPDX-License-Identifier: Apache-2.0
+
+import json
+import subprocess
+
+from cliff.command import Command
+from loguru import logger
+from redis.exceptions import RedisError
+
+from osism import utils
+from osism.utils.inventory import get_hosts_from_inventory, get_inventory_path
+
+
+class Facts(Command):
+    """Reset (clear) the cached Ansible facts.
+
+    By default the whole fact cache is flushed. Use ``--limit`` to clear
+    only the facts of selected hosts or groups. The command does not
+    gather new facts; the cache is rebuilt on the next Ansible run that
+    collects facts.
+    """
+
+    def get_parser(self, prog_name):
+        parser = super(Facts, self).get_parser(prog_name)
+        parser.add_argument(
+            "-l",
+            "--limit",
+            type=str,
+            help="Limit the reset to selected hosts or groups (Ansible host pattern)",
+        )
+        return parser
+
+    def take_action(self, parsed_args):
+        if parsed_args.limit:
+            return self._reset_limited(parsed_args.limit)
+        return self._reset_all()
+
+    def _reset_all(self):
+        removed = 0
+        try:
+            cursor = 0
+            while True:
+                cursor, batch = utils.redis.scan(
+                    cursor, match="ansible_facts*", count=100
+                )
+                if batch:
+                    utils.redis.delete(*batch)
+                    removed += len(batch)
+                if cursor == 0:
+                    break
+        except RedisError as exc:
+            logger.error(f"Failed to reset Ansible fact cache: {exc}")
+            return 1
+
+        logger.info(f"Removed cached facts for {removed} host(s)")
+        return 0
+
+    def _reset_limited(self, limit):
+        try:
+            result = subprocess.run(
+                [
+                    "ansible-inventory",
+                    "-i",
+                    get_inventory_path("/ansible/inventory/hosts.yml"),
+                    "--list",
+                    "--limit",
+                    limit,
+                ],
+                capture_output=True,
+                text=True,
+                timeout=30,
+            )
+
+            if result.returncode != 0:
+                logger.error(
+                    f"Error loading inventory (rc={result.returncode}): "
+                    f"{result.stderr}"
+                )
+                return 1
+        except subprocess.TimeoutExpired:
+            logger.error("Timeout loading inventory.")
+            return 1
+
+        try:
+            inventory = json.loads(result.stdout)
+        except json.JSONDecodeError as exc:
+            logger.error(f"Failed to parse inventory output: {exc}")
+            return 1
+
+        hosts = get_hosts_from_inventory(inventory)
+
+        if not hosts:
+            logger.warning("No hosts matched the given limit.")
+            return 0
+
+        keys = [f"ansible_facts{host}" for host in hosts]
+        try:
+            deleted = utils.redis.delete(*keys)
+        except RedisError as exc:
+            logger.error(f"Failed to reset Ansible fact cache: {exc}")
+            return 1
+
+        logger.info(f"Removed cached facts for {deleted} host(s)")
+        return 0
--- a/osism/commands/reset.py
+++ b/osism/commands/reset.py
@@ -31,6 +31,9 @@ class Facts(Command):
         return parser
 
     def take_action(self, parsed_args):
+        if parsed_args.limit is not None and not parsed_args.limit.strip():
+            logger.error("--limit must not be empty.")
+            return 1
         if parsed_args.limit:
             return self._reset_limited(parsed_args.limit)
         return self._reset_all()
--- a/tests/unit/commands/test_reset.py
+++ b/tests/unit/commands/test_reset.py
@@ -84,6 +84,19 @@ def test_facts_returns_nonzero_on_redis_error(mock_redis, loguru_logs):
 # --- limited path ---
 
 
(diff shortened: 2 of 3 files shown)


osism/commands/reset.py
```diff
@@ -0,0 +1,104 @@
...
+        parser.add_argument(
+            "-l",
+            "--limit",
+            type=str,
+            help="Limit the reset to selected hosts or groups (Ansible host pattern)",
+        )
+        return parser
+
+    def take_action(self, parsed_args):
+        if parsed_args.limit:
```
ideaship, line 37: `if parsed_args.limit:` treats `-l ""` as "no limit" and falls through to `_reset_all()`, flushing the entire fact cache.

The realistic trigger is a script calling `osism reset facts -l "$PATTERN"` where `$PATTERN` expands empty — the operator means to scope the reset but wipes every host instead.

Reject an empty/whitespace-only limit explicitly before branching, e.g.:

```python
def take_action(self, parsed_args):
    if parsed_args.limit is not None and not parsed_args.limit.strip():
        logger.error("--limit must not be empty.")
        return 1
    if parsed_args.limit:
        return self._reset_limited(parsed_args.limit)
    return self._reset_all()
```

Don't fix this by passing `""` through to `ansible-inventory` instead: I verified against the pinned `ansible-core==2.19.3` that `ansible-inventory --limit ""` returns **all** hosts (rc=0).

Please add a test asserting `-l ""` returns non-zero and never calls `redis.delete`/`scan`.

ideaship (changes requested):

tests/unit/commands/test_reset.py
```diff
@@ -0,0 +1,158 @@
...
+    assert rc == 1
+    mock_redis.delete.assert_not_called()
+    errors = [r for r in loguru_logs if r["level"] == "ERROR"]
+    assert any("Failed to reset Ansible fact cache" in r["message"] for r in errors)
+
+
+# --- limited path ---
+
+
+def test_facts_limit_deletes_only_selected_hosts(mock_redis):
```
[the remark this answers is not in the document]
berendt: Added the missing `assert mock_redis.delete.call_count == 1` to the existing limited-path RedisError test so it verifies exactly one batched delete attempt is made before the error surfaces, matching the reviewer's request that 'only one delete is attempted'.

Files changed: `tests/unit/commands/test_reset.py`

---

_Posted by [planwerk-agent](https://github.com/planwerk/planwerk-agent) 394730b address with Claude_

osism/commands/reset.py
berendt: Reject an empty/whitespace-only --limit with a non-zero exit before branching in take_action, and added a parametrized test asserting -l "" and -l "   " return 1 and never call redis.scan/delete.

Files changed: `osism/commands/reset.py`, `tests/unit/commands/test_reset.py`

---

_Posted by [planwerk-agent](https://github.com/planwerk/planwerk-agent) 394730b address with Claude_

berendt:
## Address Report

> Addressed by [planwerk-agent](https://github.com/planwerk/planwerk-agent) 394730b with Claude



### Threads

- **PRRT_kwDOGYFtAs6KRZRH** — DONE: Added the missing `assert mock_redis.delete.call_count == 1` to the existing limited-path RedisError test so it verifies exactly one batched delete attempt is made before the error surfaces, matching the reviewer's request that 'only one delete is attempted'.
  - Files: `tests/unit/commands/test_reset.py`
- **PRRT_kwDOGYFtAs6LSStb** — DONE: Reject an empty/whitespace-only --limit with a non-zero exit before branching in take_action, and added a parametrized test asserting -l "" and -l "   " return 1 and never call redis.scan/delete.
  - Files: `osism/commands/reset.py`, `tests/unit/commands/test_reset.py`

### Summary

The branch had already grown (via autosquash) since the review to include `test_facts_limit_returns_nonzero_on_redis_error`, which sets `mock_redis.delete.side_effect = RedisError(...)`, asserts `rc == 1`, and asserts the same 'Failed to reset Ansible fact cache' error as the flush-all path. The only remaining gap versus the reviewer's explicit ask was the 'only one delete is attempted' assertion; since `_reset_limited` deletes all matched keys in a single batched `redis.delete(*keys)` call, I added `assert mock_redis.delete.call_count == 1` to verify the error surfaces after exactly one attempt without retrying. I adapted the reviewer's per-key `assert_has_calls`/`ansible_facts:host` suggestion to the actual single-batch implementation and `ansible_facts<host>` key format, preserving the requested semantics.

Addressed the reviewer's exact ask: an empty or whitespace-only --limit was falsy and fell through to _reset_all(), flushing the whole fact cache. take_action now rejects such a limit (logger.error + return 1) before any branching, so no Redis key is touched. A parametrized unit test covers both "" and "   ", asserting rc==1 and that neither redis.scan nor redis.delete is called. Tests were not run locally per project convention; CI will execute them.

---

_Address report generated by [planwerk-agent](https://github.com/planwerk/planwerk-agent) 394730b address with Claude_

ideaship (approved):

PR #2392, opened 2026-06-17
Add an osism openstack command that wraps the OpenStack CLI

### LINKED ISSUE #2391: Add an `osism openstack` command that runs the OpenStack CLI against the configured clouds
**Category**: feature | **Scope**: Medium

## Description

Add an `osism openstack` command that runs arbitrary OpenStack CLI commands from inside the osism container image, using the clouds defined in the configuration repository's `clouds.yaml`. The command should behave like a transparent wrapper: anything that works with the plain `openstack` client should work through `osism openstack`, passing all subcommands and parameters straight through. By default it targets the `admin` cloud, with the cloud selectable explicitly when needed.

The command should pick up cloud credentials the same way the existing flavor-manager and image-manager (for example the Octavia image) already do, rather than relying on a separate openstack container. This reuses the established mechanism for loading clouds and secrets from the configuration repository.

The author asks that the existing implementation of the flavor-manager and image-manager be used as the reference for how credentials and clouds are loaded.

## Motivation

Operators today reach for a separate openstack container to talk to a cloud, but that container cannot read the encrypted `secure.yml` in the configuration repository, so authenticating is awkward. Running the OpenStack CLI directly through osism removes that extra container and the access problem in one step, while reusing the same credential handling operators already trust from the flavor-manager and image-manager.

Without this, operators keep juggling a second container and working around its inability to decrypt `secure.yml`. With it, a single `osism openstack ...` invocation gives them the full OpenStack CLI against the right cloud straight from the osism container.

---

_Drafted by [planwerk-review](https://github.com/planwerk/planwerk-review) e69a237 with Claude_

berendt:
## Summary

Adds an `osism openstack` command that transparently wraps the OpenStack
CLI from inside the osism container image. It passes all subcommands and
parameters straight through to the `openstack` client, defaults to the
`admin` cloud, and lets the cloud be selected with `--cloud`.

Credentials are loaded the same way the flavor-manager and image-manager
do — via `setup_cloud_environment`/`get_cloud_helpers` from
`osism/tasks/openstack.py`, which decrypts `secrets.yml`/`secure.yml`
from the configuration repository. This removes the need for a separate
openstack container that cannot read the encrypted `secure.yml`.

The command runs the client in-process (`subprocess.run`), structurally
mirroring the existing `osism openstack stress` command, so stdout/stderr
and the exit code pass through unmodified — the behaviour an operator
expects from a transparent wrapper.

Closes #2391

## Changes (in commit order)

1. **Declare `python-openstackclient` runtime dependency** —
   `requirements.txt`. The `openstack` binary the command shells out to
   ships with `python-openstackclient`, which was not a declared
   dependency (only the `openstacksdk` library was). Pinned to `9.0.0`,
   the highest release compatible with the pinned `openstacksdk==4.10.0`
   (it requires `openstacksdk>=4.6.0`; 10.x requires `>=4.12.0`/`>=4.14.0`
   and would conflict).
2. **Add and register `osism openstack` passthrough command** —
   `osism/commands/openstack.py` + `setup.cfg`. The command and its entry
   point land together so the CLI is never half-registered. `--os-cloud`
   is injected so the client selects the profile whose password
   `setup_cloud_environment` wrote into the temporary `clouds.yaml`.
   Both `openstack` and the existing `openstack stress` entry points are
   registered; cliff resolves the longest matching command prefix, so
   `openstack stress` keeps routing to `OpenStackStress` while
   `openstack server list` routes to the new command.
3. **Add unit tests** — `tests/unit/commands/test_openstack.py`. Cover
   argument passthrough with the default `admin` cloud, `--cloud`
   selection, the setup-failure short circuit, exit-code propagation, a
   missing `openstack` binary, and the cliff routing coexistence with
   `openstack stress`.
4. **Document `osism openstack` in CHANGELOG** — `CHANGELOG.md`.

## Notes for reviewers

- **Execution model.** The issue says load credentials "the same way the
  flavor-manager and image-manager do," which dispatch a Celery task on
  the `openstack` worker queue. This change reuses the *same credential
  helpers* but runs the client **in-process** (like `osism openstack
  stress`) rather than via a Celery task. Only the in-process model
  delivers a true transparent wrapper — real exit codes and stdio/TTY
  passthrough; a Celery round-trip would stream output through Redis and
  lose exit-code/interactivity transparency.
- **`stress` is now a reserved word under `osism openstack`.** Plain
  `openstack` has no `stress` subcommand, so nothing real is lost.
- **Working directory.** `setup_cloud_environment` `os.chdir("/tmp")`, so
  relative paths in passed-through commands resolve against `/tmp`. This
  is pre-existing behaviour shared with `stress`/the managers, not
  introduced here.

---

_Drafted by [planwerk-review](https://github.com/planwerk/planwerk-review) e69a237 with Claude:claude-opus-4-8_

### DIFF
--- a/osism/commands/openstack.py
+++ b/osism/commands/openstack.py
@@ -1,0 +1,128 @@
+# SPDX-License-Identifier: Apache-2.0
+
+import argparse
+import subprocess
+
+from cliff.command import Command
+from loguru import logger
+
+
+class _PassthroughParser(argparse.ArgumentParser):
+    """Argument parser for a transparent OpenStack CLI passthrough.
+
+    ``argparse.REMAINDER`` only starts capturing at the first *positional*
+    token, so any OpenStack global option placed before the subcommand
+    (``--os-interface``, ``--os-region-name``, ...) would be matched against
+    this wrapper's own options and rejected. cliff calls ``parse_args()`` in
+    ``App.run_subcommand``; delegating that to ``parse_known_args()`` lets the
+    wrapper consume only its own options (``--cloud``) and forward everything
+    else verbatim, preserving order.
+
+    Note: cliff's *app-level* parser still consumes its own global options
+    (``--debug``, ``-v``/``-q``, ``--log-file``) before this parser ever runs,
+    so those cannot be forwarded to the OpenStack CLI.
+    """
+
+    def parse_args(self, args=None, namespace=None):
+        namespace, arguments = self.parse_known_args(args, namespace)
+        namespace.arguments = arguments
+        return namespace
+
+
+class Run(Command):
+    """Run the OpenStack CLI against a configured cloud.
+
+    Everything other than the ``--cloud`` selector is forwarded verbatim to
+    the ``openstack`` CLI, including global options that precede the
+    subcommand, e.g. ``osism openstack --os-region-name RegionOne server
+    list``. Select the cloud profile with ``--cloud`` rather than
+    ``--os-cloud``; the latter is injected automatically and rejected when
+    passed through.
+    """
+
+    def get_parser(self, prog_name):
+        # Build our own parser instead of super().get_parser(): we need the
+        # _PassthroughParser class, allow_abbrev=False so a forwarded token
+        # like "--clou" is not silently expanded to this wrapper's --cloud,
+        # and add_help=False so -h/--help is forwarded to the OpenStack CLI.
+        parser = _PassthroughParser(
+            prog=prog_name,
+            description=self.get_description(),
+            add_help=False,
+            allow_abbrev=False,
+        )
+        parser.add_argument(
+            "--cloud",
+            type=str,
+            default="admin",
+            help="Cloud name in clouds.yaml (default: %(default)s)",
+        )
+        return parser
+
+    def take_action(self, parsed_args):
+        from osism.tasks.openstack import get_cloud_helpers
+
+        # The cloud profile is selected with --cloud; injecting it as
+        # --os-cloud as well would be ambiguous and order-dependent. This is a
+        # best-effort guard: it matches "--os-cloud"/"--os-cloud=" exactly and
+        # does not catch an argparse-style abbreviation (e.g. "--os-clou") that
+        # openstackclient might still expand downstream.
+        if any(
+            arg == "--os-cloud" or arg.startswith("--os-cloud=")
+            for arg in parsed_args.arguments
+        ):
+            logger.error(
+                "Do not pass '--os-cloud' through to the OpenStack CLI; "
+                "use the '--cloud' option to select the cloud profile."
+            )
+            return 1
+
+        setup_cloud_environment, _, cleanup_cloud_environment = get_cloud_helpers()
+
+        _, temp_files, original_cwd, success = setup_cloud_environment(
+            parsed_args.cloud
+        )
+        # setup_cloud_environment may have copied temp files and/or changed the
+        # working directory before failing, so always clean up once it ran.
+        try:
+            if not success:
+                logger.error(
+                    f"Failed to set up cloud environment for '{parsed_args.cloud}'"
+                )
+                return 1
+
+            command = ["openstack", "--os-cloud", parsed_args.cloud] + list(
+                parsed_args.arguments
+            )
+            # Log only the leading subcommand tokens and an argument count: the
+            # forwarded vector routinely carries secrets (e.g. 'user create
+            # --password ...', '--secret ...') that must not end up in durable
+            # debug logs.
+            subcommand = []
+            for arg in parsed_args.arguments:
+                if arg.startswith("-"):
+                    break
+                subcommand.append(arg)
+            logger.debug(
+                "Running 'openstack {}' against cloud '{}' ({} argument(s))".format(
+                    " ".join(subcommand) or "<no subcommand>",
+                    parsed_args.cloud,
+                    len(parsed_args.arguments),
+                )
+            )
+
+            try:
+                # No shell=True and command is a list, so the passed-through
+                # arguments are exec'd directly without shell interpretation;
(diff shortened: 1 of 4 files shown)


osism/commands/openstack.py
```diff
@@ -0,0 +1,77 @@
...
+        parser = super(Run, self).get_parser(prog_name)
+        parser.add_argument(
+            "--cloud",
+            type=str,
+            default="admin",
+            help="Cloud name in clouds.yaml (default: %(default)s)",
+        )
+        parser.add_argument(
+            "arguments",
+            nargs=argparse.REMAINDER,
```
ideaship: `arguments` uses `argparse.REMAINDER`, which only starts capturing at the first *positional* token. Any OpenStack global option placed *before* the subcommand is therefore matched against this wrapper's own optionals and rejected before `take_action` runs. I checked:

- `osism openstack --debug server list` → `unrecognized arguments: --debug`
- `osism openstack --os-interface public server list` → rejected
- `osism openstack --os-region-name RegionOne server list` → rejected
- `osism openstack server list --cloud foo` → `--cloud` is *not* consumed here; it stays in `arguments` and is forwarded to the CLI while the wrapper still runs against the default `admin` cloud
- `osism openstack -- server list` → the `--` delimiter is retained in `arguments` and forwarded to `openstack`

Since openstackclient is cliff-based, its global options must precede the subcommand, so moving them after the subcommand is not equivalent.

The `--os-cloud` handling just below shows you already considered global options and deliberately route cloud selection through `--cloud` only. So: is this constrained passthrough (subcommand first, no leading global options) intentional? If so, a note in the help text / docstring would make that clear to users. If the intent is a transparent passthrough, then `--debug` / `--os-interface` / `--os-region-name` being rejected is an unintended side effect worth fixing.

If you do fix it, note it is **not** a one-line `parse_known_args` swap: cliff calls `cmd_parser.parse_args(sub_argv)` in `App.run_subcommand`, so you would need a custom parser whose `parse_args` delegates to `parse_known_args` and stashes the remainder (or another routing approach), plus a test exercising a leading global option.

osism/commands/openstack.py
```diff
@@ -0,0 +1,77 @@
...
+            help="Subcommand and parameters passed to the OpenStack CLI",
+        )
+        return parser
+
+    def take_action(self, parsed_args):
+        from osism.tasks.openstack import get_cloud_helpers
+
+        # The cloud profile is selected with --cloud; injecting it as
+        # --os-cloud as well would be ambiguous and order-dependent.
+        if any(
```
ideaship, line 70: Good guard — routing cloud selection through `--cloud` only is the right call. One caveat: it matches `--os-cloud` / `--os-cloud=` exactly and won't catch an argparse-style abbreviation (e.g. `--os-clou`) that openstackclient could still expand to `--os-cloud` downstream. Today the leading case is already rejected by the parser, so this is minor — a one-line comment noting the guard is best-effort would suffice. If you later relax the parser to admit leading global options, revisit this guard's robustness at the same time.

osism/commands/openstack.py
```diff
@@ -0,0 +1,77 @@
...
+            if not success:
+                logger.error(
+                    f"Failed to set up cloud environment for '{parsed_args.cloud}'"
+                )
+                return 1
+
+            command = ["openstack", "--os-cloud", parsed_args.cloud] + list(
+                parsed_args.arguments
+            )
+            logger.debug(f"Executing OpenStack CLI: {' '.join(command)}")
```
ideaship: Because this is a raw passthrough, the forwarded arguments frequently contain secrets — `user create --password …`, `application credential create --secret …`, `--property secret=…`, inline tokens. Dumping the full `command` here writes those values into debug logs, which (unlike a transient `ps` listing) are durable and often shipped to centralized logging. Consider logging only the subcommand plus an argument count, or redacting known-sensitive flags, rather than the whole argument vector.

tests/unit/commands/test_openstack.py
```diff
@@ -0,0 +1,149 @@
...
+    with patch(
+        "osism.tasks.openstack.get_cloud_helpers",
+        return_value=(setup, MagicMock(), cleanup),
+    ), patch("osism.commands.openstack.subprocess.run", run):
+        result = cmd.take_action(parsed_args)
+
+    return result, setup, cleanup, run
+
+
+def test_passes_arguments_through_with_default_cloud():
```
ideaship, line 44: Every passthrough test here puts the subcommand first, so none exercises a *leading* OpenStack global option (`--debug`, `--os-interface`, `--os-region-name`) — exactly the shape that REMAINDER rejects (see the comment on `osism/commands/openstack.py`). If the passthrough is meant to be transparent, a regression test like `osism openstack --debug server list` asserting the forwarded vector is `["openstack", "--os-cloud", "admin", "--debug", "server", "list"]` would catch it. As written, the suite is green while that invocation shape does not work.

tests/unit/commands/test_openstack.py
```diff
@@ -0,0 +1,149 @@
...
+        result = cmd.take_action(parsed_args)
+
+    assert result == 1
+    cleanup.assert_called_once_with([], "/orig")
+
+
+def test_openstack_subcommands_route_correctly():
+    from cliff.commandmanager import CommandManager
+
+    manager = CommandManager("osism.commands")
```
ideaship, line 180: `CommandManager("osism.commands")` reads *installed* entry-point metadata, not the source tree, so this test fails with a misleading `Unknown command …` when `egg-info` is stale (e.g. `setup.cfg` edited without a reinstall). It is green in CI (fresh install) but confusing locally. Worth a comment noting the reinstall dependency, or asserting against the source registration instead of installed metadata.

ideaship (changes requested):
Several issues/questions regarding the handling of global openstack options. Needs clarification.

ideaship (approved):

PR #2427, opened 2026-07-02
Create per-device ZTP firmware symlink from sonic_parameters.version

berendt:
## Summary

Adds support for per-device SONiC firmware selection to the `sync sonic` command, matching the ZTP `dynamic-url` switch in [osism/ansible-collection-services#2131](https://github.com/osism/ansible-collection-services/pull/2131).

That PR changed the ZTP firmware install to a `dynamic-url` built from `<prefix><identifier><suffix>` (identifier `serial-number`), so each switch fetches `sonic-broadcom-enterprise-base_<serial>.bin` during ZTP. For that name to resolve to the right image, a per-serial link to the version-specific image must exist. The serial→version mapping lives in NetBox (`sonic_parameters.version`), and `sync sonic` already writes the per-serial artefacts into the export directory.

## Changes

- **Settings** (`settings.py`): new `SONIC_FIRMWARE_DIR` / `_PREFIX` / `_SUFFIX` / `_IDENTIFIER`, mirroring the ansible httpd role defaults. `SONIC_FIRMWARE_DIR` defaults to the same httpd-served directory as the config exports (`/etc/sonic/export`).
- **`export_firmware_link()`** (`exporter.py`): creates `<prefix><serial><suffix>` as a *relative* symlink to `<prefix><version><suffix>`. Reconciled on every run (missing/stale/wrongly-pointed links are repaired), skips devices without a configured version, and raises on real failures so they surface as a non-zero task rc. The target image is provided out of band (e.g. downloaded by metalbox), so a dangling link until it arrives is expected.
- **`sync_sonic`** (`sync.py`): reads `sonic_parameters.version` per device (like `hwsku` / `config_version`) and passes it to the exporter. `config_db.json` and the CONFIG DB VERSION are untouched.

For the example device (serial `732656X2419308`, `version: 4.4.2`) this produces:

```
/etc/sonic/export/sonic-broadcom-enterprise-base_732656X2419308.bin
    -> sonic-broadcom-enterprise-base_4.4.2.bin
```

## Tests

- 13 tests in `test_exporter.py` (identifier/target naming, serial fallback, reconciliation, self-reference guard, error paths).
- 3 tests in `test_sync.py` (version → firmware link, `None` when absent, rc=1 on failure); `patch_sync_deps` and `make_device` extended accordingly.

### DIFF
--- a/osism/tasks/conductor/sonic/sync.py
+++ b/osism/tasks/conductor/sonic/sync.py
@@ -18,8 +18,25 @@ from .config_generator import (
     _load_metalbox_devices_cache,
 )
 from .constants import DEFAULT_SONIC_ROLES, SUPPORTED_HWSKUS
-from .exporter import save_config_to_netbox, export_config_to_file
+from .exporter import (
+    save_config_to_netbox,
+    export_config_to_file,
+    export_firmware_link,
+)
 from .cache import clear_interface_cache, get_interface_cache_stats
+
+
+def _get_sonic_parameter(device, key):
+    """Return ``sonic_parameters[key]`` from a device's custom fields, or None.
+
+    Centralises the nested ``custom_fields -> sonic_parameters -> <key>``
+    lookup so the per-field reads in ``sync_sonic`` (hwsku, config_version,
+    version) stay short and consistent. Missing custom fields, a missing or
+    empty ``sonic_parameters``, or a missing key all yield None.
+    """
+    custom_fields = getattr(device, "custom_fields", None) or {}
+    sonic_parameters = custom_fields.get("sonic_parameters") or {}
+    return sonic_parameters.get(key)
 
 
 def sync_sonic(device_name=None, task_id=None, show_diff=True):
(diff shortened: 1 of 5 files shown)


osism/tasks/conductor/sonic/sync.py
```diff
@@ -237,7 +248,11 @@ def sync_sonic(device_name=None, task_id=None, show_diff=True):
                 # Export the generated configuration to local file (only if changed)
                 file_changed = export_config_to_file(device, sonic_config)
 
-                if netbox_changed or file_changed:
+                # Reconcile the per-device ZTP firmware symlink (only when a
+                # sonic_parameters.version is configured)
+                firmware_changed = export_firmware_link(device, version)
```
ideaship: **The firmware link is not actually reconciled to NetBox state — it's created opportunistically, which defeats the point of the feature.** The docstring calls this a reconcile "on every call," but it isn't one. The link is silently skipped whenever a real SONiC switch has `version` set but (1) has no/unsupported `hwsku` yet, or (2) throws in any earlier config step — and once a link exists, clearing `version` in NetBox never removes it, so ZTP keeps serving a withdrawn image.

Mechanism: the call sits after `if not hwsku: continue` (line 196), after the unsupported-HWSKU `continue` (207-211), and inside the per-device `try` after config gen/save/export — while `export_firmware_link` early-returns on `if not version` (exporter.py:321) before touching the filesystem. The result is an ordering inversion: firmware is needed *early* in ZTP bring-up, but here it's gated behind config eligibility that only matters *later*. A freshly-registered switch with `serial`+`version` but no `hwsku` yet — the exact case this feature exists to serve — 404s on its firmware fetch. (The line-187 comment claiming independence from `config_version` is technically true but masks the real coupling to `hwsku`.)

This should be fixed rather than documented away:
- Move the `export_firmware_link` call before the HWSKU gate / outside the config `try`, so firmware is reconciled for every in-scope device regardless of config eligibility.
- Remove a managed link when `version` is absent — guarded to only unlink a symlink whose target matches `<prefix><version><suffix>`, so an operator-placed real image is never clobbered.

If you believe the create-only behavior is genuinely safe, that claim rests on an invariant — `hwsku` and `version` are always set together at registration and `version` is never cleared — that isn't enforced or tested anywhere in the repo. Please don't just add a comment asserting it; either enforce it in code (fail loudly when `version` is set without `hwsku`) or make the reconcile complete as above.

Regression tests to add (the current suite defaults a supported `hwsku` on every device, so none of these paths are exercised):
- `version` set + no/unsupported `hwsku` → link is still created
- earlier config step raises + `version` set → asserts intended link behavior
- pre-existing link, then `version` cleared → link is removed

ideaship (changes requested):

ideaship (approved):

PR #2434, opened 2026-07-05
Add unit tests for osism/tasks/__init__.py

### LINKED ISSUE #2353: Unit tests for osism/tasks/__init__.py
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul infrastructure). Part of **Tier 5** (#2199). `osism/tasks/__init__.py` (489 LOC) is the shared foundation of all Celery workers: the Celery `Config` class, container-version lookup, the central `run_ansible_in_environment` / `run_command` subprocess runners (incl. host extraction from Ansible output via `HOST_PATTERN`), play-execution logging, and the CLI-side `handle_task` wait/revoke helper.

## Scope

Add `tests/unit/tasks/test_init.py` covering the module-level helpers in `osism/tasks/__init__.py`. There is no existing coverage: `osism.tasks.handle_task` is only ever *mocked* in `tests/unit/commands/test_manage_wiring.py`, never tested directly. `tests/unit/tasks/__init__.py` already exists, so only the test module itself needs to be created. None of the targets are Celery tasks (plain functions/classes), so no broker is needed. If the `run_ansible_in_environment` matrix grows too large, the issue may be split during implementation (subprocess runners vs. pure helpers), as done for Tier 3.

### `HOST_PATTERN` — `__init__.py:19`

Pure regex, no mocking. Parametrize:

- `"ok: [node-1]"`, `"changed: [host.example.com]"`, `"failed: [x]"`, `"skipping: [x]"`, `"unreachable: [x]"` → match, `group(2)` is the hostname
- Delegation lines: `"ok: [node-1 -> 192.168.0.5]"` → `[^\]]+` captures `"node-1 -> 192.168.0.5"` (pin current behavior)
- `"fatal: [node-1]: FAILED!"` → no match (Ansible reports failures as `fatal:`, which the pattern does not list — document this gap with a test)
- `"TASK [Gathering Facts]"`, `"ok:"`, empty string → no match
- Leading whitespace → no match against the raw line (the caller strips first; the pattern is anchored with `^`)

### `Config` — `__init__.py:22`

No mocking; plain attribute assertions:

- `broker_url == "redis://redis"`, `result_backend == "redis://redis"`, `task_default_queue == "default"`, `task_create_missing_queues is True`, `broker_connection_retry_on_startup is True`, `enable_utc is True`
- `task_routes` maps all eight `osism.tasks.<sub>.*` patterns to their queues (`osism-ansible`, `ceph-ansible`, `conductor`, `kolla-ansible`, `kubernetes`, `netbox`, `openstack`, `reconciler`)
- `enable_ironic` defaults to the string `"True"` (read from `ENABLE_IRONIC` at import time — only the already-imported value can be asserted; do not reload the module)
- `task_track_started == (True,)` — note: this is a 1-tuple due to a trailing comma; pin current behavior and flag the likely typo in the PR description (fixing it is out of scope here)

### `get_container_version(worker)` — `__init__.py:43`

Patch `osism.tasks.Path` (the file path `/interface/versions/{worker}.yml` is hardcoded) with a callable redirecting into `tmp_path`, then write real YAML files.

- Version file missing → returns `"unknown"`, debug logged
- `osism-ansible` file with `osism_ansible_version: "7.0.5a"` → returns `"7.0.5a"` (key derived via `worker.replace("-", "_") + "_version"`)
- `kolla-ansible` / `ceph-ansible` / `osism-kubernetes` key derivation (parametrize)
- Version value `""` → returns `"latest"`
- YAML present but key missing → returns `"unknown"`
- Empty YAML file (`yaml.safe_load` → `None`) → `None.get` raises `AttributeError`, caught by the broad `except` → returns `"unknown"`, warning logged
- Invalid YAML (`yaml.YAMLError`) → returns `"unknown"`, warning logged

### `log_play_execution(...)` — `__init__.py:95`

Patch `osism.tasks.Path` (hardcoded `/share/ansible-execution-history.json`) to a `tmp_path` file and `osism.tasks.get_container_version` (covered above). `fcntl.flock` works on real temp files — no mock needed.

- Happy path: one JSON line appended; record contains `request_id`, `worker`, `worker_version` (from the patched `get_container_version`), `environment`, `role`, `hosts`, `arguments`, `result`; `timestamp` ends with `"Z"`
- `hosts=None` → `[]` in the record; `arguments=None` and `arguments=""` → `""` (falsy check)
- Two calls append two lines (mode `"a"`)
- Parent directory does not exist → created via `mkdir(parents=True, exist_ok=True)`
- Write failure (e.g. patched `Path` pointing into a read-only dir, or `builtins.open` raising) → warning logged, **no exception propagates**

### `run_ansible_in_environment(...)` — `__init__.py:146`

Patch:

- `osism.tasks.subprocess.Popen`
- `osism.tasks.utils.redis` — lazy attribute on `osism.utils`; patch with `mocker.patch("osism.tasks.utils.redis", new=MagicMock(), create=True)` **before** anything touches it, otherwise `osism.utils.__getattr__` opens a real Redis connection
- `osism.tasks.utils.create_redlock`, `osism.tasks.utils.push_task_output`, `osism.tasks.utils.finish_task_output`
- `osism.tasks.log_play_execution`
- `osism.tasks.tempfile.mkdtemp` (return a `str(tmp_path / "ssh")` dir) and optionally `osism.tasks.shutil.rmtree` to assert cleanup

**Argument handling**

- `arguments` as `list` → joined with spaces; as `str` → passed through (`type(...) == list` is an exact-type check — a tuple is *not* joined; pin that)
- `worker="kolla-ansible"` + `"-e kolla_action=stop"` in arguments → `-e kolla_action_stop_ignore_missing=true` appended; already present → not duplicated; non-kolla worker with stop action → not appended

**Environment construction** (assert via `Popen.call_args.kwargs["env"]`)

- `ANSIBLE_FORCE_COLOR == "1"`, `PY_COLORS == "1"`, `PYTHONUNBUFFERED == "1"`
- `ANSIBLE_SSH_CONTROL_PATH_DIR` set to the `mkdtemp` result (prefix contains the `request_id`)
- `ANSIBLE_SSH_RETRIES` defaults to `"3"` (`str(ssh_retries)`); pre-set via `monkeypatch.setenv("ANSIBLE_SSH_RETRIES", "7")` → not overridden
- Sub-environment `environment="manager.sub1"` → `env["SUB"] == "manager.sub1"`, `env["ENVIRONMENT"] == "manager"`, default branch builds `/run-manager.sh`
- `utils.redis.get("ansible_vault_password")` truthy → `env["VAULT"] == "/ansible-vault.py"`; `None` → key absent

**Worker dispatch** (assert the `command` string passed to `Popen`, incl. `stdbuf -oL ` prefix and `shell=True`)

- `kolla-ansible` → `stdbuf -oL /run.sh deploy <role> <args>`
- `kolla-ansible` + role `mariadb-backup` (and alias `mariadb_backup`) → action `backup`, role rewritten to `mariadb`, and `-e kolla_action=deploy` (or any of the listed actions) regex-rewritten to `-e kolla_action=backup`
- `osism-kubernetes` and `ceph-ansible` → `stdbuf -oL /run.sh <role> <args>`
- Any other worker → `stdbuf -oL /run-<environment>.sh <role> <args>`

**Output loop, host extraction, logging**

- Fake process: `poll` `side_effect=[None, None, 0]`, `stdout.readline` returning e.g. `b"ok: [node-2]\n"`, `b"ok: [node-1]\n"`, `wait` → `0`; returned `result` is the concatenated output; `push_task_output` called once per line; `finish_task_output` called with `rc=0`
- `log_play_execution` called twice: first with `result="started"`, `hosts=None`; then with `result="success"` and `hosts=["node-1", "node-2"]` (deduplicated via set, **sorted**)
- `wait` returns non-zero → second call has `result="failure"`
- Duplicate host lines (`ok:` + `changed:` for the same host) → host appears once
- `publish=False` → neither `push_task_output` nor `finish_task_output` called

**Locking & cleanup**

- `locking=True` → `create_redlock(key="lock-ansible-<environment>-<role>", auto_release_time=...)`; `lock.acquire()` called; `lock.release()` called in the `finally` even when `wait` raises
- `lock.release()` raising → warning logged, no exception propagates
- `locking=False` → `create_redlock` never called
- SSH control dir: `shutil.rmtree` called on the `mkdtemp` dir even when `Popen` raises; `rmtree` raising → warning logged, original return/exception flow unaffected

### `run_command(...)` — `__init__.py:371`

Patch `osism.tasks.subprocess.Popen`, `osism.tasks.utils.push_task_output`, `osism.tasks.utils.finish_task_output`, `osism.tasks.utils.create_redlock`.

- `Popen` invoked with the argv list `[command, arg1, arg2]` (no `shell=True`) and `stderr=subprocess.STDOUT`
- `ignore_env=True` → `env` passed verbatim; `ignore_env=False` → `os.environ` copy updated with `env` (assert an injected key plus an inherited one via `monkeypatch.setenv`)
- Output loop: lines accumulated into the return value; `push_task_output` per line; `finish_task_output(request_id, rc=rc)`
- `publish=False` → no push/finish calls
- `locking=True` → `create_redlock(key="lock-<command>", ...)` and `lock.release()` called — note: `acquire()` is **never** called in the current implementation; pin that behavior and flag it in the PR description
- `locking=False` → `create_redlock` not called

### `handle_task(t, wait, format, timeout)` — `__init__.py:418`

Patch `osism.tasks.utils.fetch_task_output`, `osism.tasks.utils.revoke_task`, and `prompt_toolkit.prompt` (function-local `from prompt_toolkit import prompt` → patch in the `prompt_toolkit` module). Use a `MagicMock` task: the code reads **both** `t.id` (for `fetch_task_output`) and `t.task_id` (for all log messages).

- `wait=True`, fetch succeeds → returns the fetch result; called with `(t.id, timeout=timeout)`
- `fetch_task_output` raises `TimeoutError` → returns `1`, info logs mention the timeout and the `osism wait` follow-up command (use the `loguru_logs` fixture from `tests/conftest.py`)
- `KeyboardInterrupt` + prompt answers `"y"` / `"yes"`: `revoke_task(t.task_id)` called; returns `True` → "has been revoked" logged; returns `False` → error logged; function returns `1`
- `KeyboardInterrupt` + prompt answer `"n"` (or default) → no revoke, "continues running" logged, returns `1`
- `KeyboardInterrupt` + prompt itself raises `KeyboardInterrupt` → handled, returns `1`
- `KeyboardInterrupt` + prompt raises `EOFError` → handled, returns `1`
- `wait=False`, `format="log"` → returns `0`, info logged, `fetch_task_output` never called
- `wait=False`, `format="script"` → prints exactly `t.task_id` (assert via `capsys`), returns `0`

## Mocking hints

- All patch targets live at `osism.tasks.*` — the module imports `os`, `subprocess`, `tempfile`, `shutil`, `yaml` and `utils` directly, so patch e.g. `osism.tasks.subprocess.Popen`, not `subprocess.Popen`.
- `utils.redis` is created lazily by `osism.utils.__getattr__` on first access and would open a real connection — always patch with `create=True` before calling `run_ansible_in_environment`.
- Build a small fake-process helper/fixture for both runners:
  ```python
  def make_process(lines, rc=0):
      p = MagicMock()
      p.poll.side_effect = [None] * len(lines) + [rc]
      p.stdout.readline.side_effect = [line.encode() for line in lines]
      p.wait.return_value = rc
      return p
  ```
- For `get_container_version` / `log_play_execution` the file paths are hardcoded absolute paths; patching `osism.tasks.Path` with `lambda p: tmp_path / Path(p).name` keeps everything inside `tmp_path` and lets the real YAML/JSON and `fcntl` code run.
- `log_play_execution` is called from inside `run_ansible_in_environment` — patch it there so the runner tests don't touch the history file, and test it in isolation separately.
- Loguru output is invisible to `caplog`; use the existing `loguru_logs` fixture from `tests/conftest.py` for all "warning/error logged" assertions.
- None of these functions are Celery tasks, so they can be called directly; the Celery `app` lives in `osism/tasks/ansible.py` etc. and is not touched by this issue.

## Definition of Done

- [ ] `tests/unit/tasks/test_init.py` created
- [ ] `pytest --cov=osism.tasks` shows ≥ 90 % for `osism/tasks/__init__.py`
- [ ] `pipenv run pytest tests/unit/tasks/test_init.py` passes locally

## Dependencies

berendt:
Closes #2353

Adds unit test coverage for the shared Celery-worker helpers in `osism/tasks/__init__.py`, which previously had no dedicated test module.

### Change set (single commit)

**876ab59 — Add unit tests for osism/tasks/__init__.py**

Introduces `tests/unit/tasks/test_init.py` (775 lines) covering the module-level helpers that back the Celery workers — none of which are Celery tasks themselves, so they are exercised directly without a broker:

- **`HOST_PATTERN`** — the host-extraction regex: parametrized match/no-match cases, including the pinned `ok: [src -> dst]` delegation capture and documented gaps (`fatal:` lines, leading whitespace).
- **`Config`** — the static flags (`task_default_queue`, `task_create_missing_queues`, `broker_connection_retry_on_startup`, `enable_utc`) and the complete `task_routes` queue mapping.
- **`get_container_version`**
- **`log_play_execution`**
- **`run_ansible_in_environment`** and **`run_command`** — both subprocess runners, driven by a `make_process` helper that fakes `Popen.poll`/`stdout.readline`/`wait`.
- **`handle_task`** — the CLI-side wait/revoke helper.

### Notes for the reviewer

Two current behaviors are pinned as-is (characterization tests, not endorsements):
- `run_command` with `locking=True` creates and releases a redlock but never calls `acquire()`.
- Tuple arguments are not joined, because `run_ansible_in_environment` uses an exact-type list check.

`Config.broker_url` / `Config.result_backend` precedence and `task_track_started` are already pinned by `tests/unit/tasks/test_config.py` (which reloads `osism.settings` / `osism.tasks` to exercise import-time resolution); this module deliberately does not duplicate them. Because `test_config.py` reloads `osism.tasks`, everything under test is accessed as a module attribute (`tasks.Config`, `tasks.get_container_version`, …) rather than imported by name, so a binding cannot go stale depending on test order. Loguru output is invisible to pytest's `caplog`, so the `loguru_logs` fixture from `tests/conftest.py` is used for every "warning/error logged" assertion.

No divergence from the issue: the change is test-only and adds no production code.

---

_Implemented by [planwerk-agent](https://github.com/planwerk/planwerk-agent) 4c14036 with Claude:claude-opus-4-8_

### DIFF
--- a/tests/unit/tasks/test_init.py
+++ b/tests/unit/tasks/test_init.py
@@ -655,15 +655,13 @@ def test_run_command_publish_false_no_streaming(command_mocks):
     command_mocks.finish.assert_not_called()
 
 
-def test_run_command_locking_never_acquires(command_mocks):
-    # Pinned quirk: with locking=True a redlock is created and released but
-    # acquire() is never called in the current implementation.
+def test_run_command_locking_lifecycle(command_mocks):
     tasks.run_command("req-1", "echo", {}, locking=True)
     command_mocks.create_redlock.assert_called_once_with(
         key="lock-echo", auto_release_time=3600
     )
     lock = command_mocks.create_redlock.return_value
-    lock.acquire.assert_not_called()
+    lock.acquire.assert_called_once()
     lock.release.assert_called_once()
 
 
--- a/tests/unit/tasks/test_init.py
+++ b/tests/unit/tasks/test_init.py
@@ -1,0 +1,781 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for the module-level helpers in :mod:`osism.tasks`.
+
+Covers the shared Celery-worker foundation: the ``HOST_PATTERN`` regex, the
+Celery ``Config`` class, ``get_container_version``, ``log_play_execution`` and
+the two subprocess runners ``run_ansible_in_environment`` / ``run_command``,
+plus the CLI-side ``handle_task`` wait/revoke helper. None of these are Celery
+tasks, so they are called directly without a broker.
+
+``Config.broker_url`` / ``Config.result_backend`` precedence and
+``task_track_started`` are already pinned by ``test_config.py`` (which reloads
+``osism.settings`` / ``osism.tasks`` to exercise the import-time resolution);
+this module deliberately does not duplicate them and covers only the
+environment-independent ``Config`` attributes. ``Config.enable_ironic`` is
+likewise resolved from ``os.environ`` at import time, so asserting it here
+would depend on the environment of the first import; that reload-based
+pattern lives in ``test_config.py``.
+
+Everything under test is accessed as a module attribute (``tasks.Config``,
+``tasks.get_container_version``, ...) rather than imported by name, because
+``test_config.py`` reloads ``osism.tasks`` and a name bound at import time could
+otherwise go stale depending on test order.
+
+Loguru output is invisible to pytest's ``caplog``; the ``loguru_logs`` fixture
+from ``tests/conftest.py`` is used for every "warning/error logged" assertion.
+"""
+
+import json
+import subprocess
+from pathlib import Path as RealPath
+from types import SimpleNamespace
+from unittest.mock import MagicMock
+
+import pytest
+
+import osism.utils as utils_pkg
+from osism import tasks
+
+
+def make_process(lines, rc=0):
+    """Return a fake ``subprocess.Popen`` driving the runner output loop.
+
+    ``poll`` yields ``None`` once per line and then ``rc`` so the
+    ``while p.poll() is None`` loop reads every line exactly once before it
+    exits; ``wait`` returns the same code.
+    """
+    p = MagicMock()
+    p.poll.side_effect = [None] * len(lines) + [rc]
+    p.stdout.readline.side_effect = [line.encode() for line in lines]
+    p.wait.return_value = rc
+    return p
+
+
+# ---------------------------------------------------------------------------
+# HOST_PATTERN
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    "line, host",
+    [
+        ("ok: [node-1]", "node-1"),
+        ("changed: [host.example.com]", "host.example.com"),
+        ("failed: [x]", "x"),
+        ("skipping: [x]", "x"),
+        ("unreachable: [x]", "x"),
+        # Delegation: [^\]]+ captures the whole "src -> dst" string (pinned).
+        ("ok: [node-1 -> 192.168.0.5]", "node-1 -> 192.168.0.5"),
+    ],
+)
+def test_host_pattern_matches(line, host):
+    match = tasks.HOST_PATTERN.match(line)
+    assert match is not None
+    assert match.group(2) == host
+
+
+@pytest.mark.parametrize(
+    "line",
+    [
+        # Ansible reports task failures as "fatal:", which the pattern does not
+        # list -- documenting the known gap.
+        "fatal: [node-1]: FAILED!",
+        "TASK [Gathering Facts]",
+        "ok:",
+        "",
+        # Leading whitespace: pattern is ^-anchored and the caller strips first.
+        "  ok: [node-1]",
+    ],
+)
+def test_host_pattern_no_match(line):
+    assert tasks.HOST_PATTERN.match(line) is None
+
+
+# ---------------------------------------------------------------------------
+# Config
+# ---------------------------------------------------------------------------
+
+
+def test_config_static_flags():
+    assert tasks.Config.task_default_queue == "default"
+    assert tasks.Config.task_create_missing_queues is True
+    assert tasks.Config.broker_connection_retry_on_startup is True
+    assert tasks.Config.enable_utc is True
+
+
+def test_config_task_routes_complete_mapping():
+    assert tasks.Config.task_routes == {
+        "osism.tasks.ansible.*": {"queue": "osism-ansible"},
+        "osism.tasks.ceph.*": {"queue": "ceph-ansible"},
+        "osism.tasks.conductor.*": {"queue": "conductor"},
+        "osism.tasks.kolla.*": {"queue": "kolla-ansible"},
+        "osism.tasks.kubernetes.*": {"queue": "kubernetes"},
+        "osism.tasks.netbox.*": {"queue": "netbox"},
+        "osism.tasks.openstack.*": {"queue": "openstack"},
+        "osism.tasks.reconciler.*": {"queue": "reconciler"},
+    }
+
+
+# ---------------------------------------------------------------------------
+# get_container_version
+# ---------------------------------------------------------------------------
+
+
+@pytest.fixture
+def patched_path(mocker, tmp_path):
+    """Redirect the hardcoded absolute paths in ``osism.tasks`` into tmp_path.
+
+    Both ``get_container_version`` (``/interface/versions/{worker}.yml``) and
+    ``log_play_execution`` (``/share/ansible-execution-history.json``) build a
+    ``Path`` from a literal path; keeping only the basename under ``tmp_path``
+    lets the real YAML / JSON / ``fcntl`` code run against real temp files.
+    """
+    mocker.patch("osism.tasks.Path", new=lambda p: tmp_path / RealPath(p).name)
+    return tmp_path
+
+
+def test_get_container_version_missing_file_returns_unknown(patched_path, loguru_logs):
+    assert tasks.get_container_version("osism-ansible") == "unknown"
+    assert any(
+        r["level"] == "DEBUG" and "Version file not found" in r["message"]
+        for r in loguru_logs
+    )
+
+
+def test_get_container_version_reads_version(patched_path):
+    (patched_path / "osism-ansible.yml").write_text('osism_ansible_version: "7.0.5a"\n')
+    assert tasks.get_container_version("osism-ansible") == "7.0.5a"
+
+
+@pytest.mark.parametrize(
+    "worker, key",
+    [
+        ("kolla-ansible", "kolla_ansible_version"),
+        ("ceph-ansible", "ceph_ansible_version"),
+        ("osism-kubernetes", "osism_kubernetes_version"),
+    ],
+)
+def test_get_container_version_key_derivation(patched_path, worker, key):
+    (patched_path / f"{worker}.yml").write_text(f'{key}: "1.2.3"\n')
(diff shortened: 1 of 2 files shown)


tests/unit/tasks/test_init.py
```diff
@@ -0,0 +1,775 @@
...
+    command_mocks.finish.assert_called_once_with("req-1", rc=0)
+
+
+def test_run_command_publish_false_no_streaming(command_mocks):
+    tasks.run_command("req-1", "echo", {}, publish=False)
+    command_mocks.push.assert_not_called()
+    command_mocks.finish.assert_not_called()
+
+
+def test_run_command_locking_never_acquires(command_mocks):
```
ideaship: `run_command`'s `locking=True` path creates the redlock and calls `release()` but never `acquire()` (`osism/tasks/__init__.py:392-416`), so it provides no mutual exclusion — compare `run_ansible_in_environment`, which acquires on every branch. This test asserts `acquire.assert_not_called()` as a "Pinned quirk", which locks the missing acquire in as intended behavior; the moment the bug is fixed, this test flips red and reads as a regression.

Suggest fixing forward: add `lock.acquire()` in `run_command` (its own commit, or a separate PR — your call), then update this test to assert `acquire.assert_called_once()` so it characterizes the correct behavior. The `locking` param is forwarded through `netbox.manage` and the `openstack.py` `*_sync` functions, though no caller passes `locking=True` today — so this is latent, not an active race, and there's no rush on the production side.

tests/unit/tasks/test_init.py
```diff
@@ -0,0 +1,775 @@
...
+        "osism.tasks.conductor.*": {"queue": "conductor"},
+        "osism.tasks.kolla.*": {"queue": "kolla-ansible"},
+        "osism.tasks.kubernetes.*": {"queue": "kubernetes"},
+        "osism.tasks.netbox.*": {"queue": "netbox"},
+        "osism.tasks.openstack.*": {"queue": "openstack"},
+        "osism.tasks.reconciler.*": {"queue": "reconciler"},
+    }
+
+
+def test_config_enable_ironic_default_string():
```
ideaship: `enable_ironic` is resolved from `os.environ` at import time (`osism/tasks/__init__.py:25`), so this asserts whatever value was frozen into `Config` at first import — it passes on default CI but fails under `ENABLE_IRONIC=False`. `test_config.py` already covers import-time env resolution for the Config fields using the reload-with-controlled-env pattern, so I'd drop this test rather than duplicate that coverage non-hermetically here.

ideaship (changes requested):

ideaship (approved):

PR #2458, opened 2026-07-13
Add unit tests for osism/tasks/netbox.py

### LINKED ISSUE #2355: Unit tests for osism/tasks/netbox.py
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul infrastructure). Part of **Tier 5** (#2199). `osism/tasks/netbox.py` (410 LOC) defines the Celery tasks of the `netbox` worker: setting device custom fields (`maintenance`, `provision_state`, `power_state`) on the primary NetBox and on all matching `NETBOX_SECONDARIES` instances (Redlock-guarded, semaphore-limited), simple `get_*` lookup tasks, the `manage` task wrapping `/usr/local/bin/netbox-manager` via `run_command`, and `ping`. Two private helpers (`_update_netbox_device_field`, `_matches_netbox_filter`) carry most of the logic.

## Scope

Add `tests/unit/tasks/test_netbox.py` covering `osism/tasks/netbox.py`. No existing coverage: `tests/unit/tasks/conductor/test_netbox.py` covers `osism/tasks/conductor/netbox.py` (`get_device_oob_ip`, query-list helpers) and `tests/unit/commands/test_netbox.py` covers `osism/commands/netbox.py` — both are different modules, no overlap.

All tasks are bound Celery tasks (`bind=True`); invoke them directly (e.g. `netbox.set_maintenance("node-1", True)`) — Celery binds `self` on direct calls, no broker needed.

The three `set_*` tasks are structurally identical (same lock/filter/update/skip flow, only the field name and log wording differ). Parametrize over `(task, field_name)` instead of triplicating every test — same approach as `QUERY_LIST_VARIANTS` in `tests/unit/tasks/conductor/test_netbox.py`.

### `_update_netbox_device_field(nb, device_name, field_name, value)` — `netbox.py:19`

Patch:

- `osism.tasks.netbox.utils.create_netbox_semaphore` → return a `MagicMock()` (supports the `with` protocol)

Pass a `MagicMock` as `nb` (no patching of `utils.nb` needed here).

- Device found → `device.custom_fields.update({field_name: value})` and `device.save()` called, returns `True`
- `nb.dcim.devices.get` returns `None` → returns `False`, no save, no error log (the caller logs)
- `requests.exceptions.ConnectTimeout` raised → returns `False`, "Connection timeout" logged
- `requests.exceptions.ReadTimeout` raised → returns `False`, "Request timeout" logged (use `ReadTimeout`, not `ConnectTimeout` — `ConnectTimeout` inherits from both `ConnectionError` and `Timeout` and is caught by the first handler)
- `requests.exceptions.ConnectionError` raised → returns `False`, "Connection error" logged
- `requests.exceptions.HTTPError` (generic `RequestException`) raised → returns `False`, "Request error" logged
- Exception raised by `device.save()` (not only by `get`) → same handling, returns `False`
- Semaphore is created with `nb.base_url` and entered/exited around the API call (`__enter__`/`__exit__` asserted)

### `_matches_netbox_filter(nb, netbox_filter, is_primary=False)` — `netbox.py:66`

No patching; pass `SimpleNamespace`/`MagicMock` objects with `base_url` and optional `netbox_name`/`netbox_site` attributes.

- `netbox_filter=None` and `netbox_filter=""` → `True` (empty filter matches everything)
- `is_primary=True`, filter `"primary"` → `True`; also case-insensitive (`"PRIMARY"`) and substring (`"primary-dc"` contains `"primary"`)
- `is_primary=False`, filter `"primary"`, URL/name/site without "primary" → `False`
- Filter substring of `base_url`, case-insensitive (`"NETBOX.example"` vs `https://netbox.example.com`) → `True`
- Object without `netbox_name`/`netbox_site` attributes (primary instances) → no `AttributeError`, falls through to `False` when URL does not match
- `netbox_name` set and matching (case-insensitive substring) → `True`
- `netbox_site` set and matching → `True`
- Nothing matches → `False`

### `run(action, arguments)` — `netbox.py:104`

- Patch `osism.tasks.netbox.utils.check_task_lock_and_exit`; direct call returns `None`, lock check invoked once
- Lock check raises `SystemExit` (locked) → propagates

### `set_maintenance` / `set_provision_state` / `set_power_state` — `netbox.py:117` / `netbox.py:190` / `netbox.py:263`

Patch (parametrized over the three tasks and their field names `maintenance` / `provision_state` / `power_state`):

- `osism.tasks.netbox.utils.check_task_lock_and_exit`
- `osism.tasks.netbox.utils.create_redlock` → `MagicMock` whose `.acquire(timeout=120)` returns `True`/`False`
- `osism.tasks.netbox._update_netbox_device_field`
- `osism.utils.nb` — lazy attribute, patch with `mocker.patch("osism.utils.nb", new=MagicMock(), create=True)` (same convention as `tests/unit/tasks/conductor/sonic/conftest.py`)

For secondaries, prefer passing the `secondary_nb_list=...` parameter over patching `osism.utils.secondary_nb_list`.

- `lock.acquire` returns `False` → returns `False`, "Could not acquire lock" logged, `_update_netbox_device_field` never called, `lock.release()` not called
- `create_redlock` called with `key=f"lock_osism_tasks_netbox_{device_name}"`, `auto_release_time=300`
- Happy path, `netbox_filter=None`, `secondary_nb_list=[]` → update called once with `(utils.nb, device_name, field, state)`, returns `True`, `lock.release()` called
- `_update_netbox_device_field` returns `False` → "Could not set ..." error logged, task still returns `True`
- `netbox_filter` not matching the primary (e.g. filter that matches nothing) → primary skipped (update not called for `utils.nb`), still returns `True`
- Two secondaries passed via `secondary_nb_list`, one matching the filter, one not → update called only for the matching one (plus/minus primary depending on filter)
- `secondary_nb_list=None` → falls back to `utils.secondary_nb_list` (patch the lazy attribute with `create=True` to a one-element list and assert it is processed)
- `_update_netbox_device_field` raises → exception propagates but `lock.release()` is still called (`finally`)
- `set_power_state` only: `state=None` → converted to `"n/a"` (update called with `"n/a"`); conversion happens before the lock-check
- One representative test that `check_task_lock_and_exit` raising `SystemExit` aborts before `create_redlock` is called

### `get_location_id` / `get_rack_id` — `netbox.py:340` / `netbox.py:352`

Patch `osism.utils.nb` (lazy, `create=True`). Parametrize over `(task, endpoint)` — `dcim.locations` vs `dcim.racks`:

- Object found → returns `.id`
- `get` returns `None` → returns `None`
- `get` raises `ValueError` (pynetbox raises this for ambiguous matches) → returns `None`

### `get_devices` / `get_device_by_name` / `get_interfaces_by_device` / `get_addresses_by_device_and_interface` — `netbox.py:364` / `netbox.py:369` / `netbox.py:374` / `netbox.py:379`

Thin pass-throughs; patch `osism.utils.nb` and assert delegation:

- `get_devices(role="server", site="x")` → `nb.dcim.devices.filter(role="server", site="x")`, return value passed through
- `get_device_by_name("n1")` → `nb.dcim.devices.get(name="n1")`
- `get_interfaces_by_device("n1")` → `nb.dcim.interfaces.filter(device="n1")`
- `get_addresses_by_device_and_interface("n1", "eth0")` → `nb.dcim.addresses.filter(device="n1", interface="eth0")` (test as written — the module uses `dcim.addresses`)

### `manage(*arguments, publish=True, locking=False, auto_release_time=3600)` — `netbox.py:384`

Patch:

- `osism.tasks.netbox.utils.check_task_lock_and_exit`
- `osism.tasks.netbox.run_command` (imported into the module namespace via `from osism.tasks import ... run_command`)
- `osism.tasks.netbox.settings.NETBOX_URL` / `NETBOX_TOKEN` / `IGNORE_SSL_ERRORS`

Cases:

- Env dict passed to `run_command` contains exactly `NETBOX_MANAGER_URL`, `NETBOX_MANAGER_TOKEN`, `NETBOX_MANAGER_IGNORE_SSL_ERRORS`, `NETBOX_MANAGER_VERBOSE="true"`; all values `str()`-coerced (e.g. `IGNORE_SSL_ERRORS=True` → `"True"`)
- Command is `/usr/local/bin/netbox-manager`; positional `*arguments` forwarded; `publish`/`locking`/`auto_release_time` forwarded (defaults and explicit values)
- Returns `run_command`'s return value
- First positional argument is `self.request.id` — on a direct call this is `None`; that is fine since `run_command` is mocked (alternatively use `manage.apply(args=[...])` for a real task id)
- Lock check raising `SystemExit` → `run_command` never called

### `ping()` — `netbox.py:407`

- Patch `osism.utils.nb`; returns `utils.nb.status()` result unchanged

### `setup_periodic_tasks(sender, **kwargs)` — `netbox.py:15`

- No-op signal handler; call once with `sender=None` for line coverage

## Mocking hints

- `utils.nb` and `utils.secondary_nb_list` are lazy module attributes resolved via `osism.utils.__getattr__` — always patch with `create=True` and target `osism.utils.nb` (never trigger the real `__getattr__`, it would open a NetBox connection). See the `mock_nb` fixtures in `tests/unit/tasks/conductor/sonic/conftest.py` and `tests/unit/tasks/conductor/test_netbox.py:59`.
- `utils.check_task_lock_and_exit` calls `exit(1)` when locked; in tests, set `side_effect=SystemExit(1)` on the patched mock for the "locked" cases and assert with `pytest.raises(SystemExit)`.
- The Redlock mock needs only `acquire(timeout=120) -> bool` and `release()`; assert the exact `timeout=120` kwarg.
- Build secondary instances as `MagicMock(base_url="https://nb2.example.com")` and set `netbox_name` / `netbox_site` attributes explicitly where a test needs them; use `spec`/`SimpleNamespace` for the "attribute absent" cases so `getattr(..., None)` actually returns `None`.
- For the `set_*` tasks, patching `osism.tasks.netbox._update_netbox_device_field` keeps the lock/filter tests independent of the helper, which is tested separately.
- Importing `osism.tasks.netbox` instantiates `Celery("netbox")` with `config_from_object(Config)` — config only, no broker connection; safe in unit tests (same pattern as the other `osism.tasks` modules already imported by existing tests).

## Definition of Done

- [ ] `tests/unit/tasks/test_netbox.py` created
- [ ] `pytest --cov=osism.tasks.netbox` shows ≥ 90 %
- [ ] `pipenv run pytest tests/unit/tasks/test_netbox.py` passes locally

## Dependencies

berendt:
Adds `tests/unit/tasks/test_netbox.py` covering the Celery tasks of the `netbox` worker as specified in #2355:

- `_update_netbox_device_field`: happy path (custom-field update + save), device not found, all four `requests` exception handlers (`ConnectTimeout`, `ReadTimeout`, `ConnectionError`, `HTTPError`), errors raised by `device.save()`, and the semaphore being created from `nb.base_url` and entered/exited around the API call.
- `_matches_netbox_filter`: empty filter, `primary` filter variants (case-insensitive, substring), URL/name/site matching, missing `netbox_name`/`netbox_site` attributes, and the no-match fall-through.
- `set_maintenance` / `set_provision_state` / `set_power_state`: one parametrised body over the three tasks (same approach as `QUERY_LIST_VARIANTS` in the conductor tests) covering Redlock key/timeouts, unacquired lock, failed update still returning `True`, filter-based primary/secondary skipping, the `utils.secondary_nb_list` fallback, lock release on exceptions, the `set_power_state` `None` → `"n/a"` conversion, and task-lock abort before Redlock creation.
- `get_location_id` / `get_rack_id` (parametrised): found, missing, and `ValueError` from ambiguous matches.
- The four pass-through `get_*` tasks: delegation to the expected pynetbox endpoints.
- `manage`: exact `netbox-manager` environment with `str()`-coerced settings, argument/keyword forwarding to `run_command`, return value pass-through, and no command execution when tasks are locked.
- `ping` and `setup_periodic_tasks`.

`utils.nb` and `utils.secondary_nb_list` are lazy attributes resolved via `osism.utils.__getattr__`; they are always patched with `create=True` so no real NetBox connection is ever opened. All tasks are invoked directly — Celery binds `self` on direct calls, so no broker is needed.

Closes #2355

### DIFF
--- a/tests/unit/tasks/test_netbox.py
+++ b/tests/unit/tasks/test_netbox.py
@@ -1,0 +1,683 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for the ``netbox`` worker tasks (``osism/tasks/netbox.py``).
+
+Covers the device-field update helper (semaphore limiting, error handling),
+the NetBox filter matcher, the three structurally identical ``set_*`` tasks
+(Redlock guarding, primary/secondary fan-out, filter skipping), the ``get_*``
+lookup and pass-through tasks, ``manage`` (environment and argument forwarding
+to ``run_command``) and ``ping``.
+
+All tasks are bound Celery tasks and are invoked directly -- Celery binds
+``self`` on direct calls, so no broker is needed. ``utils.nb`` and
+``utils.secondary_nb_list`` are lazy module attributes resolved via
+``osism.utils.__getattr__``; they are always patched with ``create=True`` so
+the real NetBox wiring is never triggered.
+"""
+
+from types import SimpleNamespace
+from unittest.mock import MagicMock, call
+
+import pytest
+import requests.exceptions
+
+from osism.tasks import netbox
+
+# ---------------------------------------------------------------------------
+# Helpers / fixtures
+# ---------------------------------------------------------------------------
+
+
+# The three ``set_*`` tasks share the same lock/filter/update/skip flow; only
+# the custom field name and the log wording differ. Parametrize over them
+# instead of triplicating every test (same approach as ``QUERY_LIST_VARIANTS``
+# in ``tests/unit/tasks/conductor/test_netbox.py``).
+SET_TASK_VARIANTS = [
+    pytest.param(
+        netbox.set_maintenance, "maintenance", "maintenance", True, id="maintenance"
+    ),
+    pytest.param(
+        netbox.set_provision_state,
+        "provision_state",
+        "provision state",
+        "active",
+        id="provision-state",
+    ),
+    pytest.param(
+        netbox.set_power_state,
+        "power_state",
+        "power state",
+        "power on",
+        id="power-state",
+    ),
+]
+
+SET_TASK_PARAMS = "set_task,field_name,log_label,state"
+
+# Base URL of the NetBox instance handed to ``_update_netbox_device_field``.
+NB_API_URL = "https://netbox.example.com"
+
+
+def _has_log(records, level, substring):
+    return any(r["level"] == level and substring in r["message"] for r in records)
+
+
+def _make_secondary(base_url, netbox_name=None, netbox_site=None):
+    """Build a secondary NetBox instance stub.
+
+    ``netbox_name`` / ``netbox_site`` default to ``None`` (attribute present
+    but unset) so the filter matcher falls through to the URL check.
+    """
+    nb = MagicMock()
+    nb.base_url = base_url
+    nb.netbox_name = netbox_name
+    nb.netbox_site = netbox_site
+    return nb
+
+
+@pytest.fixture
+def mock_nb(mocker):
+    """Replace ``osism.utils.nb`` (lazy attribute) with a fresh MagicMock."""
+    nb = MagicMock()
+    nb.base_url = "https://netbox-primary.example.com"
+    nb.netbox_name = None
+    nb.netbox_site = None
+    mocker.patch("osism.utils.nb", new=nb, create=True)
+    return nb
+
+
+@pytest.fixture
+def patch_lock_check(mocker):
+    return mocker.patch("osism.tasks.netbox.utils.check_task_lock_and_exit")
+
+
+@pytest.fixture
+def patch_redlock(mocker):
+    lock = MagicMock()
+    lock.acquire.return_value = True
+    create = mocker.patch("osism.tasks.netbox.utils.create_redlock", return_value=lock)
+    return SimpleNamespace(create=create, lock=lock)
+
+
+@pytest.fixture
+def patch_update_field(mocker):
+    return mocker.patch(
+        "osism.tasks.netbox._update_netbox_device_field", return_value=True
+    )
+
+
+@pytest.fixture
+def patch_semaphore(mocker):
+    semaphore = MagicMock()
+    create = mocker.patch(
+        "osism.tasks.netbox.utils.create_netbox_semaphore", return_value=semaphore
+    )
+    return SimpleNamespace(create=create, semaphore=semaphore)
+
+
+@pytest.fixture
+def nb_api():
+    """NetBox API stub passed into ``_update_netbox_device_field``.
+
+    ``base_url`` is set explicitly rather than left to ``MagicMock``
+    auto-attribute creation: the helper hands it to
+    ``create_netbox_semaphore`` and interpolates it into every error message.
+    """
+    nb = MagicMock()
+    nb.base_url = NB_API_URL
+    return nb
+
+
+# ---------------------------------------------------------------------------
+# _update_netbox_device_field
+# ---------------------------------------------------------------------------
+
+
+def test_update_field_updates_and_saves(patch_semaphore, nb_api):
+    device = nb_api.dcim.devices.get.return_value
+
+    assert (
+        netbox._update_netbox_device_field(nb_api, "node-1", "maintenance", True)
+        is True
+    )
+
+    nb_api.dcim.devices.get.assert_called_once_with(name="node-1")
+    device.custom_fields.update.assert_called_once_with({"maintenance": True})
+    device.save.assert_called_once_with()
+
+
+def test_update_field_device_not_found(patch_semaphore, nb_api, loguru_logs):
+    nb_api.dcim.devices.get.return_value = None
+
+    assert (
+        netbox._update_netbox_device_field(nb_api, "node-1", "maintenance", True)
+        is False
+    )
+
+    # The helper stays silent on a missing device; the caller logs the error.
+    assert not any(r["level"] == "ERROR" for r in loguru_logs)
+
+
+@pytest.mark.parametrize(
+    "exception,message",
+    [
+        pytest.param(
+            requests.exceptions.ConnectTimeout("boom"),
+            "Connection timeout while updating",
+            id="connect-timeout",
+        ),
+        # ReadTimeout, not ConnectTimeout: ConnectTimeout inherits from both
+        # ConnectionError and Timeout and is caught by the first handler.
+        pytest.param(
+            requests.exceptions.ReadTimeout("boom"),
+            "Request timeout while updating",
+            id="read-timeout",
+        ),
+        pytest.param(
+            requests.exceptions.ConnectionError("boom"),
+            "Connection error while updating",
+            id="connection-error",
+        ),
+        pytest.param(
+            requests.exceptions.HTTPError("boom"),
+            "Request error while updating",
+            id="request-error",
+        ),
+    ],
+)
+def test_update_field_request_errors_from_get(
+    patch_semaphore, nb_api, loguru_logs, exception, message
+):
+    nb_api.dcim.devices.get.side_effect = exception
+
+    assert (
+        netbox._update_netbox_device_field(nb_api, "node-1", "maintenance", True)
+        is False
+    )
+
+    assert _has_log(
+        loguru_logs,
+        "ERROR",
+        f"{message} maintenance for device node-1 on {NB_API_URL}",
+    )
+
+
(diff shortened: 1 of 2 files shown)


tests/unit/tasks/test_netbox.py
```diff
@@ -0,0 +1,672 @@
...
+    result = netbox.get_interfaces_by_device("n1")
+
+    mock_nb.dcim.interfaces.filter.assert_called_once_with(device="n1")
+    assert result is mock_nb.dcim.interfaces.filter.return_value
+
+
+def test_get_addresses_by_device_and_interface_delegates_to_filter(mock_nb):
+    result = netbox.get_addresses_by_device_and_interface("n1", "eth0")
+
+    mock_nb.dcim.addresses.filter.assert_called_once_with(device="n1", interface="eth0")
```
ideaship: This assertion pins a broken pynetbox path. The production task at `osism/tasks/netbox.py:380` calls `utils.nb.dcim.addresses.filter(...)`, but pynetbox exposes IP addresses under `ipam.ip_addresses`, not `dcim.addresses` — see `osism/tasks/conductor/netbox.py:122,177` in this same repo, which use `utils.nb.ipam.ip_addresses.filter(...)`. Against a real NetBox this 404s; the full `MagicMock` hides that. Don't land a test that certifies the wrong endpoint: fix the production line to `ipam.ip_addresses` in a separate commit, then assert the corrected path here. (The `device=`/`interface=` filter keys are valid on that endpoint — both filter by name — so only the app/endpoint changes.)

tests/unit/tasks/test_netbox.py
```diff
@@ -0,0 +1,672 @@
...
+
+
+@pytest.mark.parametrize("netbox_filter", [None, ""], ids=["none", "empty"])
+def test_matches_filter_empty_filter_matches_everything(netbox_filter):
+    nb = SimpleNamespace(base_url="https://netbox.example.com")
+
+    assert netbox._matches_netbox_filter(nb, netbox_filter) is True
+
+
+@pytest.mark.parametrize("netbox_filter", ["primary", "PRIMARY", "primary-dc"])
```
ideaship: `"primary-dc"` in this parametrize list cements a production bug. In `_matches_netbox_filter` (`osism/tasks/netbox.py:83`) the primary branch is `"primary" in filter_lower` — reversed from every other check, which does `filter_lower in <field>`. As a result any filter containing the substring "primary" (e.g. a secondary at site `primary-region`) also selects the primary and updates it unintentionally. Don't assert `"primary-dc"` matches here. Correct the production line (exact-keyword match, or `filter_lower in "primary"`) in a separate commit, then update the test to match.

tests/unit/tasks/test_netbox.py
```diff
@@ -0,0 +1,672 @@
...
+    result = set_task("node-1", state, secondary_nb_list=[])
+
+    assert result is False
+    assert _has_log(loguru_logs, "ERROR", "Could not acquire lock for node node-1")
+    patch_update_field.assert_not_called()
+    patch_redlock.lock.release.assert_not_called()
+
+
+@pytest.mark.parametrize(SET_TASK_PARAMS, SET_TASK_VARIANTS)
+def test_set_task_update_failure_logs_error_but_returns_true(
```
ideaship: This asserts `result is True` when the update failed, which cements a contract bug: all three `set_*` tasks fall through to `return True` (`osism/tasks/netbox.py:183`) even though `_update_netbox_device_field` returned `False` (lines 148-153 only log). The docstring — "True if lock was acquired and operation succeeded" (line 132) — establishes the intended contract, so this is a bug, not an ambiguity: a caller (e.g. `conductor/ironic.py:1099`) can't distinguish success from a silently-failed update. Fix the production code in a separate commit to aggregate update failures into a `False` return, and change this test to assert `False` on the failure path as part of that same fix. If the production change isn't happening here, drop this test rather than land one asserting `True`.

ideaship (changes requested):

ideaship (approved):

PR #2460, opened 2026-07-13
Add unit tests for osism/services websocket_manager and event_bridge

### LINKED ISSUE #2357: Unit tests for osism/services/{websocket_manager,event_bridge}.py
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul infrastructure). Part of **Tier 6** (#2199). `osism/services/websocket_manager.py` (271 LOC) implements `EventMessage`, per-connection event/node/service filters and a queue-based broadcaster for FastAPI WebSocket clients. `osism/services/event_bridge.py` (304 LOC) is a Redis-pub/sub bridge that forwards events from the RabbitMQ listener container to the WebSocket manager in the API container, with a local-queue fallback and reconnect logic.

## Scope

Create the `tests/unit/services/` package (`__init__.py`) plus `tests/unit/services/test_websocket_manager.py` and `tests/unit/services/test_event_bridge.py`. There is no existing coverage for either module (`tests/unit/` currently only has `commands/`, `tasks/`, `utils/`).

**New dev dependency required:** most `WebSocketManager` methods are coroutines, so the tests need `pytest-asyncio`. The Pipfile `[dev-packages]` only contains `pytest`, `pytest-cov` and `pytest-mock` — add `pytest-asyncio` (pinned like the other dev packages). Either set `asyncio_mode = auto` in the `[tool:pytest]` section of `setup.cfg` or mark async tests explicitly with `@pytest.mark.asyncio` (the plugin registers the marker, so `--strict-markers` stays happy). No new runtime dependency is needed: `fastapi` and `redis` (via `celery[redis]`) come from `requirements.txt`, which the Zuul job installs via `pipenv run pip install .` (`playbooks/test-unit.yml`, `install_requires = file: requirements.txt`).

Note: `osism/services/event_bridge.py` instantiates a module-level singleton at import time (`event_bridge.py:304`) whose `_init_redis()` attempts a real Redis connection. The exception is caught (the import stays safe), but tests must construct fresh `EventBridge` instances with `redis.Redis` patched — see mocking hints. The `_redis_subscriber_loop` retry machinery is the trickiest part; if it turns out too brittle, this issue may be split further during implementation (as done in Tier 3).

### `EventMessage` — `websocket_manager.py:15`

No patching needed (pure class; sync tests).

- Constructor sets `event_type`, `source`, `data`, `node_name`; `node_name` defaults to `None`
- `id` is a valid UUID4 string; two instances get different ids
- `timestamp` ends with `"Z"` and the prefix parses as ISO 8601 (do not assert an exact value — `datetime.utcnow()`)
- `to_dict()` (`websocket_manager.py:32`) returns exactly the keys `id`, `timestamp`, `event_type`, `source`, `node_name`, `data`
- `to_json()` (`websocket_manager.py:43`) round-trips via `json.loads` back to `to_dict()`

### `WebSocketConnection.matches_filters()` — `websocket_manager.py:57`

No patching needed (`WebSocketConnection(MagicMock())` suffices; sync tests).

- No filters set → `True` for any event (early return)
- `event_filters=["baremetal.node.power_set"]`: matching `event_type` → `True`, non-matching → `False`
- `node_filters=["node1"]`: `event.node_name="node1"` → `True`; `node_name="other"` → `False`; `node_name=None` → `False` (explicit `is not None` check)
- `service_filters=["baremetal"]`: `event_type="baremetal.node.power_set"` → `True` (service = first dot segment); `event_type="compute.instance.update"` → `False`
- `event_type=""` with `service_filters=["unknown"]` → `True` (falsy event_type maps to service `"unknown"`)
- Combined filters are AND-ed: event filter matches but node filter does not → `False`

### `WebSocketManager.connect()` / `disconnect()` — `websocket_manager.py:95` / `websocket_manager.py:106`

Use a `MagicMock` websocket with `accept = AsyncMock()`; patch `WebSocketManager._broadcast_events` (or `asyncio.create_task` at `osism.services.websocket_manager.asyncio.create_task`) so no real broadcaster loop starts, and always instantiate a fresh `WebSocketManager()` inside the async test (do not reuse the module-level singleton `websocket_manager.py:271`).

- `connect()` awaits `websocket.accept()` and registers a `WebSocketConnection` in `self.connections`
- First `connect()` starts the broadcaster task; a second `connect()` while the task is running does not start another one (`done()` check)
- `disconnect()` removes the connection; disconnecting an unknown websocket is a no-op (`pop(..., None)`)

### `WebSocketManager.update_filters()` — `websocket_manager.py:114`

- Sets `event_filters` / `node_filters` / `service_filters` when passed
- Arguments left as `None` keep the existing lists (partial update); passing `[]` clears a filter list
- Unknown websocket → silent no-op, no exception

### `WebSocketManager.add_event()` / `send_heartbeat()` — `websocket_manager.py:136` / `websocket_manager.py:258`

- `add_event()` puts the event on `event_queue` (`qsize() == 1`, `get_nowait()` returns it)
- `send_heartbeat()` with no connections → returns early, queue stays empty
- `send_heartbeat()` with a registered connection → queues an `EventMessage` with `event_type="heartbeat"`, `source="osism"`, `data={"message": "ping"}`

### `WebSocketManager.broadcast_event_from_notification()` — `websocket_manager.py:140`

No patching needed beyond a fresh manager; inspect the queued `EventMessage` via `event_queue.get_nowait()`.

- `baremetal.*` with `"ironic_object.data"` → `node_name` from `name`, `resource_id` from `uuid`
- `compute.*` / `nova.*` with `"nova_object.data"` → `node_name` prefers `host` over `name`; falls back to `name` when `host` missing
- `network.*` / `neutron.*` with `"neutron_object.data"` → `resource_id` from `id` (fallback `uuid`), `node_name` from `name` (fallback `device_id`)
- `volume.*` with `"cinder_object.data"` → `resource_id` `id`/`uuid`, `node_name` `name`/`display_name`
- `image.*` with `"glance_object.data"` and `identity.*` with `"keystone_object.data"` → analogous
- Known service type but expected payload key missing (e.g. `baremetal` without `"ironic_object.data"`) → `node_name`/`resource_id` stay `None`, event still queued
- Unknown service type (`"foo.bar"`) → no extraction, event queued with `service_type="foo"`
- `event_type=""` → `service_type="unknown"`
- Queued event: `source="openstack"`, `data` contains the original payload keys plus `service_type` and `resource_id`; the caller's payload dict is **not** mutated (`payload.copy()`)
- Exception path: pass a payload without `.copy()`/`.get` semantics (e.g. a list) → exception caught, error logged, nothing queued

### `WebSocketManager._broadcast_events()` — `websocket_manager.py:204`

Insert fake connections directly into `manager.connections` (bypassing `connect()`), pre-fill the queue, then run `task = asyncio.create_task(manager._broadcast_events())`, give it a tick (`await asyncio.sleep(0.05)`), assert, and `task.cancel()` + await it in a `finally` block.

- Event matching a connection's filters → `websocket.send_text` awaited once with `event.to_json()`
- Event not matching → `send_text` not called for that connection, but called for a second matching connection (`sent_count` logic)
- `send_text` raises `WebSocketDisconnect` → connection removed from `self.connections`
- `send_text` raises generic `Exception` → error logged, connection removed as well
- Queue event while `self.connections` is empty → event consumed, nothing sent, loop continues
- `task.cancel()` → `CancelledError` branch logs and exits the loop (task finishes, no exception propagates)

### `EventBridge.__init__()` / `_init_redis()` — `event_bridge.py:29` / `event_bridge.py:41`

Patch `osism.services.event_bridge.redis.Redis`; for env vars use `monkeypatch.setenv` / `delenv`.

- Defaults: `redis.Redis` called with `host="redis"`, `port=6379`, `db=0`, `decode_responses=True`, `socket_connect_timeout=10`, `socket_timeout=None`, `health_check_interval=30`
- `REDIS_HOST` / `REDIS_PORT` / `REDIS_DB` env vars override the defaults (ints for port/db)
- Successful `ping()` → `_redis_client` set and `pubsub()` called (subscriber created)
- `ping()` raises → error logged, `_redis_client` and `_redis_subscriber` both `None`
- `osism.services.event_bridge.REDIS_AVAILABLE` patched to `False` → warning logged, `redis.Redis` never called, client stays `None`

### `EventBridge.set_websocket_manager()` — `event_bridge.py:76`

Patch `EventBridge._start_redis_subscriber` and `EventBridge._start_processor_thread` so no real threads start.

- With Redis client present and `_subscriber_thread is None` → `_start_redis_subscriber` called
- Without Redis client (`_redis_client = None`) → `_start_redis_subscriber` not called
- `_subscriber_thread` already set → subscriber not started again
- Processor: `_processor_thread is None` → `_start_processor_thread` called; an existing thread mock with `is_alive() == True` → not called again
- `_websocket_manager` is stored

### `EventBridge.add_event()` — `event_bridge.py:89`

- Redis path: `publish("osism:events", json.dumps({"event_type": ..., "payload": ...}))` called; return value > 0 → info log, no warning
- `publish` returns `0` → warning "No Redis subscribers …" logged
- First `publish` raises → `_init_redis()` re-invoked (patch it), client still available → second `publish` succeeds ("after reconnect" log)
- First `publish` raises and re-init leaves `_redis_client = None` → fallback: event lands on `_event_queue` (local queue)
- No Redis client at all → event goes straight to `_event_queue`
- `queue.Full`: patch `_event_queue.put_nowait` with `side_effect=queue.Full` (the real queue is unbounded) → warning logged, no exception propagates
- Generic exception (e.g. patched `json.dumps` raising) → error logged, swallowed

### `EventBridge._redis_subscriber_loop()` — `event_bridge.py:154`

Call the loop synchronously in the test (no thread) with `_redis_subscriber` replaced by a `MagicMock`; drive termination via `get_message` side effects that call `self._shutdown_event.set()`, and patch `EventBridge._process_single_event`. For retry tests patch `_init_redis` and use a `_shutdown_event.wait` spy or small `retry_delay` tolerance — each test must guarantee loop exit.

- `_redis_subscriber is None` → error logged, immediate return
- `subscribe("osism:events")` called once on entry
- `get_message` returns `None` (timeout) → loop continues until shutdown is set
- Message of `type="message"` with valid JSON and `_websocket_manager` set → `_process_single_event` called with the decoded dict
- Valid message but `_websocket_manager is None` → event pushed to `_event_queue` instead
- Message with invalid JSON → `json.JSONDecodeError` branch, error logged, loop continues
- `_process_single_event` raising → generic "Error processing Redis event" branch, loop continues
- `get_message` raising → inner loop breaks, outer `except` increments retry, `_init_redis` re-called after `wait(retry_delay)`
- Five consecutive failures → "Max Redis reconnection attempts reached, giving up" logged, loop ends
- `finally` closes `_redis_subscriber`; `close()` raising is swallowed

### `EventBridge._process_single_event()` — `event_bridge.py:237`

- `_websocket_manager is None` → warning "No WebSocket manager available", returns without error
- With manager: `broadcast_event_from_notification` must be an `AsyncMock` (the method spins up a fresh event loop via `asyncio.new_event_loop()` and `run_until_complete`); assert it was awaited with `event_data["event_type"]` and `event_data["payload"]`
- Coroutine raising → error logged, exception swallowed

### `EventBridge._process_events()` — `event_bridge.py:263`

Run synchronously (no thread): pre-fill `_event_queue`, patch `_process_single_event` with a side effect that sets `_shutdown_event` after the first event.

- Queued event → `_process_single_event` called, `task_done()` honoured (queue `unfinished_tasks` drops)
- Empty queue (`queue.Empty` after the 1 s timeout) → `continue` until shutdown; verify clean exit when `_shutdown_event` is pre-set with an empty queue
- `_process_single_event` raising → error logged, loop continues (set shutdown in a second side-effect call to terminate)

### `EventBridge.shutdown()` — `event_bridge.py:283`

- Sets `_shutdown_event`
- `_redis_subscriber.close()` called; `close()` raising → error logged, shutdown continues
- Thread mocks with `is_alive() == True` → `join(timeout=5.0)` called for both; `is_alive() == False` or `None` → no join

## Mocking hints

- **Async tests:** everything on `WebSocketManager` except the constructor is `async def`. With `pytest-asyncio` in strict mode, decorate with `@pytest.mark.asyncio`; alternatively set `asyncio_mode = auto` in `setup.cfg` `[tool:pytest]`.
- **Fake websockets:** `ws = MagicMock(); ws.accept = AsyncMock(); ws.send_text = AsyncMock()` — `WebSocket` is only used as a dict key, so no spec is required. `WebSocketDisconnect` is imported in the module namespace; raise the real class from `fastapi` in `send_text.side_effect`.
- **Never use the module-level singletons** (`websocket_manager.py:271`, `event_bridge.py:304`). Create fresh instances per test so queues/locks bind to the running loop and no state leaks. For `EventBridge`, always construct under `mocker.patch("osism.services.event_bridge.redis.Redis")` — otherwise the constructor attempts a real connection. The import-time singleton itself only logs a connection error during collection (exception is caught), which is acceptable.
- **No real threads:** test `_redis_subscriber_loop` and `_process_events` by calling the target function directly with `_shutdown_event`-driven termination, and stub `_start_redis_subscriber` / `_start_processor_thread` in `set_websocket_manager` tests. This keeps the tests deterministic.
- **Broadcaster loop:** the only non-deterministic test is `_broadcast_events`; bound it with `asyncio.wait_for(...)` or a short `sleep` + `cancel()` in `finally` so a regression cannot hang CI.
- Use the `loguru_logs` pattern only where applicable — these two modules use the stdlib `logging` module (`logging.getLogger("osism.websocket")` / `"osism.event_bridge"`), so plain `caplog` with `caplog.set_level(logging.DEBUG, logger="osism.websocket")` works here.
- Notification payload fixture example for the extraction branches:
  ```python
  payload = {"ironic_object.data": {"name": "node1", "uuid": "abc-123"}}
  await manager.broadcast_event_from_notification("baremetal.node.power_set.end", payload)
  event = manager.event_queue.get_nowait()
  assert event.node_name == "node1"
  assert event.data["resource_id"] == "abc-123"
  ```

## Definition of Done

- [ ] `pytest-asyncio` added to Pipfile `[dev-packages]` (pinned); async tests marked or `asyncio_mode = auto` configured
- [ ] `tests/unit/services/__init__.py`, `tests/unit/services/test_websocket_manager.py` and `tests/unit/services/test_event_bridge.py` created
- [ ] `pytest --cov=osism.services.websocket_manager --cov=osism.services.event_bridge` shows ≥ 90 %
- [ ] `pipenv run pytest tests/unit/services/` passes locally
- [ ] No hanging tests: all loop tests terminate via shutdown events / task cancellation

## Dependencies

berendt:
Closes #2357.

Creates the `tests/unit/services/` package with unit tests for `osism/services/websocket_manager.py` and `osism/services/event_bridge.py`, which previously had no coverage.

Writing and reviewing the tests surfaced three production defects in `osism/services/event_bridge.py`; they are fixed in separate commits on top of the test commit.

## Production fixes

- **Close failed Redis subscriber before reconnect creates a new one**: when `subscribe()` raised, `_init_redis()` replaced `self._redis_subscriber` without closing the failed instance, and the per-iteration `finally` then closed the fresh replacement — the failed subscriber leaked and the next attempt subscribed on an already closed object. The failed subscriber is now closed before re-initialization, and the loop closes the current subscriber exactly once on exit.
- **Route `get_message` failures through subscriber back-off**: a failing `get_message()` resubscribed immediately on the closed subscriber and bypassed the retry counter and `_shutdown_event.wait()` back-off entirely, so persistent failures busy-looped forever. The error is now re-raised into the same bounded back-off path as a failing `subscribe()`, which closes the failed subscriber and resubscribes on a fresh one from `_init_redis()`.
- **Marshal bridged events onto the broadcaster event loop**: `_process_single_event()` drove `broadcast_event_from_notification()` on a freshly created event loop in a worker thread, waking the broadcaster's `asyncio.Queue` waiters on the API loop in a non-thread-safe way (unreliable delivery, error swallowed). `WebSocketManager` now records the loop its broadcaster runs on, and the bridge submits the coroutine with `asyncio.run_coroutine_threadsafe()`; the private-loop path remains only as a fallback while no broadcaster loop exists yet (the queue has no waiters then, so it is safe).

## test_websocket_manager.py

- `EventMessage`: constructor attributes, UUID4 id uniqueness, ISO 8601 `Z`-suffixed timestamp, exact `to_dict()` keys, `to_json()` round trip
- `WebSocketConnection.matches_filters()`: no-filter pass-through, event/node/service filters incl. `node_name=None` rejection, `""` → `unknown` service mapping and AND-combination of filters
- `WebSocketManager.connect()`/`disconnect()`: accept + registration, broadcaster started only once while running (`done()` check), restart after completion, no-op disconnect for unknown websockets
- `update_filters()`: full and partial updates, clearing via `[]`, unknown websocket no-op
- `add_event()`/`send_heartbeat()`: queueing, early return without connections, heartbeat event shape
- `broadcast_event_from_notification()`: identifier extraction for baremetal/compute/nova/network/neutron/volume/image/identity payloads incl. fallback keys, missing payload keys, unknown service types, empty event type, payload copy semantics and the swallowed-exception path
- `_broadcast_events()`: filtered delivery, `WebSocketDisconnect` and generic-error connection cleanup, consumption without connections and clean cancellation

## test_event_bridge.py

- `_init_redis()`: default connection parameters, `REDIS_HOST`/`REDIS_PORT`/`REDIS_DB` overrides, ping failure handling and the `REDIS_AVAILABLE = False` local-queue-only path
- `set_websocket_manager()`: thread startup guards for subscriber and processor threads
- `add_event()`: publish payload, zero-subscriber warning, reconnect-then-publish, fallback to the local queue when reconnection fails, `queue.Full` (documented as a defensive branch unreachable with the unbounded queue) and generic error swallowing
- `_redis_subscriber_loop()`: missing subscriber, subscribe/get_message flow, message dispatch with and without a WebSocket manager, invalid JSON, processing errors; `subscribe()` and `get_message()` failures both run through the bounded back-off — the tests install successive distinct subscriber mocks via the `_init_redis()` side effect and assert that each failed subscriber is closed and each retry subscribes on the newly created instance; retry exhaustion and swallowed `close()` errors
- `_process_single_event()`: missing-manager warning, `AsyncMock` broadcast dispatch and swallowed coroutine errors, plus an integration-style test that delivers an event from a worker thread through a real `WebSocketManager` with an active broadcaster to a connected client
- `_process_events()`: queued event processing with `task_done()`, empty-queue continuation, pre-set shutdown exit and error continuation
- `shutdown()`: shutdown event, subscriber close incl. error path and conditional thread joins

## Notes

- All tests use fresh `WebSocketManager`/`EventBridge` instances instead of the module-level singletons; `EventBridge` is always constructed with `redis.Redis` patched.
- The thread loops are called synchronously and terminated via `_shutdown_event`-driven side effects; the asyncio broadcaster tests bound the task with cancellation in `finally` and carry `pytest-timeout` markers, so no test can hang CI.
- `pytest-asyncio` 1.4.0 is added to the Pipfile `[dev-packages]` (compatible with pytest 9, `pytest<10,>=8.4`); async tests are marked explicitly with `@pytest.mark.asyncio`, so `--strict-markers` stays happy.

### DIFF
--- a/tests/unit/services/test_event_bridge.py
+++ b/tests/unit/services/test_event_bridge.py
@@ -329,44 +329,72 @@ class TestRedisSubscriberLoop:
         subscriber.get_message.side_effect = get_message
         bridge._redis_subscriber_loop()
         assert subscriber.subscribe.call_count == 2
-        assert subscriber.close.call_count == 2
+        # The subscriber stays open across the resubscribe and is only
+        # closed once when the loop exits.
+        assert subscriber.close.call_count == 1
         assert "Error getting Redis message: lost" in caplog.text
 
     @pytest.mark.timeout(10)
     def test_subscribe_error_waits_and_reinitializes_redis(
         self, bridge, mocker, caplog
     ):
+        """A failed subscriber is closed before ``_init_redis`` replaces it,
+        and the retry subscribes on the newly created instance."""
         caplog.set_level(logging.INFO, logger="osism.event_bridge")
-        subscriber = MagicMock()
-        subscriber.subscribe.side_effect = ConnectionError("down")
-        bridge._redis_subscriber = subscriber
-        init_redis = mocker.patch.object(bridge, "_init_redis")
+        failed = MagicMock()
+        failed.subscribe.side_effect = ConnectionError("down")
+        bridge._redis_subscriber = failed
+        fresh = MagicMock()
 
-        def wait(timeout=None):
+        def get_message(timeout=None):
             bridge._shutdown_event.set()
-            return True
+            return None
 
+        fresh.get_message.side_effect = get_message
+
+        def install_fresh_subscriber():
+            bridge._redis_subscriber = fresh
+
+        init_redis = mocker.patch.object(
+            bridge, "_init_redis", side_effect=install_fresh_subscriber
+        )
         wait_mock = mocker.patch.object(
-            bridge._shutdown_event, "wait", side_effect=wait
+            bridge._shutdown_event, "wait", return_value=False
         )
         bridge._redis_subscriber_loop()
         wait_mock.assert_called_once_with(5)
         init_redis.assert_called_once_with()
+        failed.subscribe.assert_called_once_with("osism:events")
+        failed.close.assert_called_once_with()
+        fresh.subscribe.assert_called_once_with("osism:events")
+        fresh.close.assert_called_once_with()
         assert "Redis subscriber error (attempt 1/5)" in caplog.text
         assert "Retrying Redis subscription in 5 seconds" in caplog.text
 
     @pytest.mark.timeout(10)
     def test_gives_up_after_max_retries(self, bridge, mocker, caplog):
+        """Every attempt subscribes on a distinct freshly created subscriber
+        and closes it after its failure."""
         caplog.set_level(logging.ERROR, logger="osism.event_bridge")
-        subscriber = MagicMock()
-        subscriber.subscribe.side_effect = ConnectionError("down")
-        bridge._redis_subscriber = subscriber
-        init_redis = mocker.patch.object(bridge, "_init_redis")
+        subscribers = [MagicMock() for _ in range(5)]
+        for subscriber in subscribers:
+            subscriber.subscribe.side_effect = ConnectionError("down")
+        replacements = iter(subscribers[1:])
+
+        def install_fresh_subscriber():
+            bridge._redis_subscriber = next(replacements)
+
+        bridge._redis_subscriber = subscribers[0]
+        init_redis = mocker.patch.object(
+            bridge, "_init_redis", side_effect=install_fresh_subscriber
+        )
         wait_mock = mocker.patch.object(
             bridge._shutdown_event, "wait", return_value=False
         )
         bridge._redis_subscriber_loop()
-        assert subscriber.subscribe.call_count == 5
+        for subscriber in subscribers:
+            subscriber.subscribe.assert_called_once_with("osism:events")
+            subscriber.close.assert_called_once_with()
         # No back-off after the fifth and final failure.
         assert wait_mock.call_count == 4
         assert init_redis.call_count == 4
--- a/tests/unit/services/test_event_bridge.py
+++ b/tests/unit/services/test_event_bridge.py
@@ -22,6 +22,7 @@ The module logs via the stdlib ``logging`` module (``osism.event_bridge``),
 so the plain ``caplog`` fixture is used for log assertions.
 """
 
+import asyncio
 import json
 import logging
 import queue
@@ -30,6 +31,7 @@ from unittest.mock import AsyncMock, MagicMock
 import pytest
 
 from osism.services.event_bridge import EventBridge
+from osism.services.websocket_manager import WebSocketManager
 
 
 @pytest.fixture
@@ -439,6 +441,8 @@ class TestProcessSingleEvent:
     def test_broadcasts_event_via_manager(self, bridge):
         manager = MagicMock()
         manager.broadcast_event_from_notification = AsyncMock()
+        # No broadcaster loop: the coroutine is driven on a private loop
+        manager.loop = None
         bridge._websocket_manager = manager
         bridge._process_single_event({"event_type": "a.b", "payload": {"x": 1}})
         manager.broadcast_event_from_notification.assert_awaited_once_with(
@@ -451,9 +455,41 @@ class TestProcessSingleEvent:
         manager.broadcast_event_from_notification = AsyncMock(
             side_effect=ValueError("boom")
         )
+        manager.loop = None
         bridge._websocket_manager = manager
         bridge._process_single_event({"event_type": "a.b", "payload": {}})
         assert "Error processing event via bridge: boom" in caplog.text
+
+    @pytest.mark.asyncio
+    @pytest.mark.timeout(10)
+    async def test_worker_thread_event_reaches_broadcaster_loop(self, bridge):
+        """End-to-end: an event handed to ``_process_single_event`` in a
+        worker thread is marshalled onto the loop the broadcaster of a real
+        ``WebSocketManager`` runs on and delivered to a connected client."""
+        manager = WebSocketManager()
+        websocket = MagicMock()
+        websocket.accept = AsyncMock()
+        delivered = asyncio.Event()
+        websocket.send_text = AsyncMock(side_effect=lambda message: delivered.set())
+        await manager.connect(websocket)
+        assert manager.loop is asyncio.get_running_loop()
+        bridge._websocket_manager = manager
+        try:
+            await asyncio.to_thread(
+                bridge._process_single_event,
+                {"event_type": "baremetal.node.power_set", "payload": {"x": 1}},
+            )
+            await asyncio.wait_for(delivered.wait(), timeout=5.0)
+        finally:
+            manager._broadcaster_task.cancel()
+            try:
+                await manager._broadcaster_task
+            except asyncio.CancelledError:
+                pass
+        message = json.loads(websocket.send_text.await_args.args[0])
+        assert message["event_type"] == "baremetal.node.power_set"
+        assert message["data"]["x"] == 1
+        assert message["data"]["service_type"] == "baremetal"
 
 
 class TestProcessEvents:
--- a/tests/unit/services/test_event_bridge.py
+++ b/tests/unit/services/test_event_bridge.py
@@ -1,0 +1,511 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for :mod:`osism.services.event_bridge`.
+
+Covers Redis initialization, the publish path with reconnect and local-queue
+fallback in ``add_event``, the subscriber and processor loops, single-event
+processing and shutdown.
+
+Every test constructs a fresh ``EventBridge`` with
+``osism.services.event_bridge.redis.Redis`` patched (the ``bridge`` fixture)
+instead of using the module-level ``event_bridge`` singleton, which attempts
+a real Redis connection at import time.
+
+The thread targets ``_redis_subscriber_loop`` and ``_process_events`` are
+called synchronously; termination is driven through ``_shutdown_event`` from
+``get_message`` / ``get`` side effects, so no real threads or sleeps are
+involved. Note that a failing ``get_message`` only breaks the inner loop and
+triggers a resubscribe; the retry counter and ``_shutdown_event.wait``
+back-off only engage when ``subscribe()`` itself raises.
+
+The module logs via the stdlib ``logging`` module (``osism.event_bridge``),
+so the plain ``caplog`` fixture is used for log assertions.
+"""
+
+import json
+import logging
+import queue
+from unittest.mock import AsyncMock, MagicMock
+
+import pytest
+
+from osism.services.event_bridge import EventBridge
+
+
+@pytest.fixture
+def redis_cls(mocker):
+    return mocker.patch("osism.services.event_bridge.redis.Redis")
+
+
+@pytest.fixture
+def bridge(redis_cls):
+    return EventBridge()
+
+
+class TestInitRedis:
+    def test_defaults(self, redis_cls, monkeypatch):
+        for name in ("REDIS_HOST", "REDIS_PORT", "REDIS_DB"):
+            monkeypatch.delenv(name, raising=False)
+        EventBridge()
+        redis_cls.assert_called_once_with(
+            host="redis",
+            port=6379,
+            db=0,
+            decode_responses=True,
+            socket_connect_timeout=10,
+            socket_timeout=None,
+            health_check_interval=30,
+        )
+
+    def test_environment_variables_override_defaults(self, redis_cls, monkeypatch):
+        monkeypatch.setenv("REDIS_HOST", "redis.example.com")
+        monkeypatch.setenv("REDIS_PORT", "16379")
+        monkeypatch.setenv("REDIS_DB", "2")
+        EventBridge()
+        assert redis_cls.call_args.kwargs["host"] == "redis.example.com"
+        assert redis_cls.call_args.kwargs["port"] == 16379
+        assert redis_cls.call_args.kwargs["db"] == 2
+
+    def test_successful_ping_sets_client_and_subscriber(self, redis_cls):
+        bridge = EventBridge()
+        client = redis_cls.return_value
+        client.ping.assert_called_once_with()
+        assert bridge._redis_client is client
+        client.pubsub.assert_called_once_with()
+        assert bridge._redis_subscriber is client.pubsub.return_value
+
+    def test_connection_failure_leaves_client_unset(self, redis_cls, caplog):
+        caplog.set_level(logging.ERROR, logger="osism.event_bridge")
+        redis_cls.return_value.ping.side_effect = ConnectionError("no redis")
+        bridge = EventBridge()
+        assert bridge._redis_client is None
+        assert bridge._redis_subscriber is None
+        assert "Failed to connect to Redis" in caplog.text
+
+    def test_redis_not_available_uses_local_queue_only(self, redis_cls, mocker, caplog):
+        caplog.set_level(logging.WARNING, logger="osism.event_bridge")
+        mocker.patch("osism.services.event_bridge.REDIS_AVAILABLE", False)
+        bridge = EventBridge()
+        redis_cls.assert_not_called()
+        assert bridge._redis_client is None
+        assert "event bridge will use local queue only" in caplog.text
+
+
+class TestSetWebsocketManager:
+    def test_stores_manager_and_starts_threads(self, bridge, mocker):
+        start_subscriber = mocker.patch.object(bridge, "_start_redis_subscriber")
+        start_processor = mocker.patch.object(bridge, "_start_processor_thread")
+        manager = MagicMock()
+        bridge.set_websocket_manager(manager)
+        assert bridge._websocket_manager is manager
+        start_subscriber.assert_called_once_with()
+        start_processor.assert_called_once_with()
+
+    def test_subscriber_not_started_without_redis_client(self, bridge, mocker):
+        bridge._redis_client = None
+        start_subscriber = mocker.patch.object(bridge, "_start_redis_subscriber")
+        mocker.patch.object(bridge, "_start_processor_thread")
+        bridge.set_websocket_manager(MagicMock())
+        start_subscriber.assert_not_called()
+
+    def test_subscriber_not_started_twice(self, bridge, mocker):
+        bridge._subscriber_thread = MagicMock()
+        start_subscriber = mocker.patch.object(bridge, "_start_redis_subscriber")
+        mocker.patch.object(bridge, "_start_processor_thread")
+        bridge.set_websocket_manager(MagicMock())
+        start_subscriber.assert_not_called()
+
+    def test_processor_not_started_when_alive(self, bridge, mocker):
+        thread = MagicMock()
+        thread.is_alive.return_value = True
+        bridge._processor_thread = thread
+        mocker.patch.object(bridge, "_start_redis_subscriber")
+        start_processor = mocker.patch.object(bridge, "_start_processor_thread")
+        bridge.set_websocket_manager(MagicMock())
+        start_processor.assert_not_called()
+
+    def test_processor_restarted_when_dead(self, bridge, mocker):
+        thread = MagicMock()
+        thread.is_alive.return_value = False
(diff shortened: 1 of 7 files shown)


ideaship (changes requested):
This PR needs a rebase.

tests/unit/services/test_event_bridge.py
```diff
@@ -0,0 +1,506 @@
...
+
+    @pytest.mark.timeout(10)
+    def test_subscribe_error_waits_and_reinitializes_redis(
+        self, bridge, mocker, caplog
+    ):
+        caplog.set_level(logging.INFO, logger="osism.event_bridge")
+        subscriber = MagicMock()
+        subscriber.subscribe.side_effect = ConnectionError("down")
+        bridge._redis_subscriber = subscriber
+        init_redis = mocker.patch.object(bridge, "_init_redis")
```
ideaship: Mocking `_init_redis` as a no-op means `self._redis_subscriber` is never replaced during a reconnect, so these tests can't exercise — and would stay green under — a lifecycle bug in the loop. In `event_bridge.py`, when `subscribe()` (line 169) raises, the `except` calls `_init_redis()` (line 221), which reassigns `self._redis_subscriber = self._redis_client.pubsub()` (line 69) to a *fresh* object *without* closing the failed one. The `finally` (lines 225–230) then closes `self._redis_subscriber` — now the fresh replacement — so the failed subscriber leaks and the next iteration attempts to subscribe using the freshly created but already closed subscriber. Whether Redis internally manages to reconnect from that state can depend on the Redis client implementation, but the lifecycle ordering is unquestionably wrong. Could you fix the ordering in production (close the failed subscriber before re-init, or don't close the fresh one) in its own commit, and make the reconnect side effect here install *successive distinct* subscriber mocks so the test asserts each retry subscribes on the newly created instance? As written the tests only check call counts, which pass either way. (Same applies to `test_gives_up_after_max_retries` below.)

tests/unit/services/test_event_bridge.py
```diff
@@ -0,0 +1,506 @@
...
+
+class TestProcessSingleEvent:
+    def test_without_manager_warns_and_returns(self, bridge, caplog):
+        caplog.set_level(logging.WARNING, logger="osism.event_bridge")
+        bridge._process_single_event({"event_type": "a.b", "payload": {}})
+        assert "No WebSocket manager available, dropping event" in caplog.text
+
+    def test_broadcasts_event_via_manager(self, bridge):
+        manager = MagicMock()
+        manager.broadcast_event_from_notification = AsyncMock()
```
ideaship, line 443: Asserting `broadcast_event_from_notification` is awaited once with the right args is a fine check of `_process_single_event`'s local contract, but the `AsyncMock` removes the one boundary that's actually broken in production. `_process_single_event` (event_bridge.py:237) runs in a worker thread and drives the coroutine on a *freshly created* event loop (lines 247–255), while `broadcast_event_from_notification` awaits `event_queue.put()` (websocket_manager.py:138) on an `asyncio.Queue` (:89) whose broadcaster waiter lives on the API loop (:211). Waking that waiter from the worker thread isn't thread-safe — it raises under asyncio debug mode and, in production, mutates the API loop's ready-queue without waking it, so delivery is unreliable and the error is swallowed at websocket_manager.py:201. Please marshal onto the API loop (e.g. `asyncio.run_coroutine_threadsafe`) in a separate commit, and add one integration-style test with a real manager + active broadcaster + worker thread. The assertion here stays valid under that fix, so it needn't change — the gap is the missing coverage, not this test.

tests/unit/services/test_event_bridge.py
```diff
@@ -0,0 +1,506 @@
...
+        def get_message(timeout=None):
+            calls["count"] += 1
+            if calls["count"] == 1:
+                raise ConnectionError("lost")
+            bridge._shutdown_event.set()
+            return None
+
+        subscriber.get_message.side_effect = get_message
+        bridge._redis_subscriber_loop()
+        assert subscriber.subscribe.call_count == 2
```
ideaship: This asserts `subscribe` and `close` are each called twice, i.e. it encodes "a `get_message` error resubscribes immediately" as the expected recovery — but that recovery is itself broken. On this path the inner loop `break`s (event_bridge.py:203–205) and falls straight into the `finally` (225–230), which closes `self._redis_subscriber` *without* calling `_init_redis()`, so the next outer iteration subscribes on the same, now-closed subscriber (the same lifecycle-ordering issue as the subscribe-failure path, minus the re-init). Layered on top of that is a separate defect: this path never increments `retry_count` and never waits — the `max_retries`/`retry_delay` back-off only guards the outer `subscribe` exception path (207–217), and a successful resubscribe resets `retry_count` to 0 (line 171), so a persistently-failing `get_message` busy-loops forever with no delay. If you fix these in separate commits (recreate the subscriber before resubscribing, and route this path through the same bounded back-off), this test's exact call counts will need to move with the fix — worth updating the assertion at the same time so it documents the hardened behavior rather than the current one.

tests/unit/services/test_event_bridge.py
```diff
@@ -0,0 +1,506 @@
...
+        bridge.add_event("a.b", {"x": 1})
+        assert bridge._event_queue.get_nowait() == {
+            "event_type": "a.b",
+            "payload": {"x": 1},
+        }
+
+    def test_full_queue_drops_event_with_warning(self, bridge, mocker, caplog):
+        caplog.set_level(logging.WARNING, logger="osism.event_bridge")
+        bridge._redis_client = None
+        mocker.patch.object(bridge._event_queue, "put_nowait", side_effect=queue.Full)
```
ideaship, line 206: `_event_queue` is an unbounded `queue.Queue()` (event_bridge.py:30), so `put_nowait` never raises `queue.Full` and the `except queue.Full` handler (133–134) is unreachable in production — this test only reaches it by patching `put_nowait` to raise. Fine to keep as a guard, but could you note in the test name/docstring that it covers a defensive branch unreachable with the current unbounded queue, so it isn't read as real drop behavior?

ideaship (changes requested):

tests/unit/services/test_event_bridge.py
berendt: Fixed in 9213f1f: the except handler now calls a new `_close_subscriber()` helper to close the failed subscriber (and drop the reference) *before* `_init_redis()` replaces it, and the per-iteration `finally` is gone — the loop closes the current subscriber exactly once on exit, so the fresh replacement is never closed prematurely. Both reconnect tests (`test_subscribe_error_waits_and_reinitializes_redis`, `test_gives_up_after_max_retries`) now install successive distinct subscriber mocks via the `_init_redis` side effect and assert that each retry subscribes on the newly created instance and that every failed subscriber is closed exactly once.

tests/unit/services/test_event_bridge.py
berendt: Fixed in 64006d6: `WebSocketManager.connect()` now records the running loop when it starts the broadcaster task (exposed as `WebSocketManager.loop`), and `_process_single_event()` submits the coroutine to that loop with `asyncio.run_coroutine_threadsafe()` and waits on the future. The private-loop path remains only as a fallback while no broadcaster loop exists yet — the queue has no waiters then, so driving it from the worker thread is safe. Added the requested integration-style test (`test_worker_thread_event_reaches_broadcaster_loop`): real `WebSocketManager`, active broadcaster on the test loop, `_process_single_event()` called from a worker thread via `asyncio.to_thread`, asserting delivery to a connected client. The existing unit tests keep their assertions unchanged and only set `manager.loop = None` so they keep exercising the fallback path.

tests/unit/services/test_event_bridge.py
berendt: Fixed in ab74671 (on top of 9213f1f): the `get_message` error is now re-raised into the outer handler instead of `break`ing, so the path runs through the same bounded back-off as a failing `subscribe()` — the failed subscriber is closed, `retry_count` is incremented and capped by `max_retries`, the loop waits `retry_delay`, and the resubscribe happens on the fresh subscriber from `_init_redis()`. One deviation from the suggested split: after 9213f1f fixes the except-handler lifecycle, the single re-raise delivers both the subscriber recreation and the bounded back-off, so both defects land in one commit rather than two. The test is renamed to `test_get_message_error_reconnects_with_backoff` and asserts the back-off wait, the retry accounting and that the resubscribe happens on the newly installed instance.

tests/unit/services/test_event_bridge.py
berendt: Done — `test_full_queue_drops_event_with_warning` now carries a docstring stating it covers a defensive branch that is unreachable with the current unbounded `queue.Queue` and is only reachable by patching `put_nowait`, so it isn't read as real drop behavior. Folded into the test commit (79305cb).

ideaship (changes requested):
Code looks good, but needs a rebase with conflict resolution before it can be merged.

PR #2463, opened 2026-07-13
Add unit tests for osism/api.py helpers and Pydantic models

### LINKED ISSUE #2359: Unit tests for osism/api.py — helper functions, Pydantic models, secret masking
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul infrastructure). Part of **Tier 7** (#2199). `osism/api.py` (1161 LOC) implements the OSISM API server (FastAPI). Because of its size the module is split into two sub-issues: this issue covers the **module-level helper functions, the Pydantic request/response models, and the inventory-secret masking**; the companion issue *"Unit tests for `osism/api.py` — HTTP & WebSocket endpoints (FastAPI TestClient)"* (Tier 7, split 2/2) covers the route handlers.

## Scope

Add `tests/unit/test_api_helpers.py` covering `_mask_inventory_secrets()`, `find_device_by_identifier()`, `process_netbox_webhook()` and the Pydantic models in `osism/api.py`. There is currently no test coverage for `osism/api.py` at all (only `tests/unit/test_smoke.py` imports the `osism` package). Endpoint behavior (HTTP status codes, masking applied in responses) is **out of scope** here — it belongs to the companion endpoint issue.

Note: importing `osism.api` has import-time side effects — `dictConfig(LogConfig().model_dump())` is executed and `osism/services/event_bridge.py:304` instantiates `EventBridge()`, which attempts a Redis connection (`redis.Redis(...).ping()`); the failure is caught and only logged, so the import works without Redis.

### `_mask_inventory_secrets()` — `api.py:35`

Pure function, no patching needed. It delegates key classification to `osism.tasks.conductor.utils._is_secret_key` (matches `password`/`secret` substrings case-insensitively and the `ironic_osism_` prefix) — use the real implementation.

- Key containing `password` (also mixed case, e.g. `ADMIN_Password`) → value replaced by `"***"`
- Key containing `secret` → `"***"`
- Key starting with `ironic_osism_` → `"***"`
- Non-secret key whose string value starts with `$ANSIBLE_VAULT;` → `"***"`; also with leading whitespace (`"  $ANSIBLE_VAULT;..."`, the code calls `.strip()` before the prefix check)
- String value *containing* `$ANSIBLE_VAULT;` but not at the start → unchanged
- Nested dict under a non-secret key → recursed; secret keys masked at any depth
- Secret key whose value is a dict → masked to `"***"` as a whole (the secret-key check wins over recursion)
- Non-dict, non-string values (int, list, `None`, bool) under non-secret keys → passed through unchanged
- List of dicts under a non-secret key → **not** traversed (lists are not recursed) — pin down the current behavior
- Empty dict → `{}`
- Non-string key (e.g. `42`) → `_is_secret_key` returns `False`, value passed through
- Input dict is not mutated (function builds a new dict)

### `find_device_by_identifier()` — `api.py:265`

Patch `utils.nb` — the lazy attribute on `osism.utils` — via `monkeypatch.setattr(osism.utils, "nb", fake_nb, raising=False)` (`raising=False` because the attribute only materializes via the module `__getattr__` on first access).

- `utils.nb` is `None`/falsy → returns `None`, no NetBox queries made
- `nb.dcim.devices.filter(name=identifier)` returns devices → first device returned; the `cf_inventory_hostname` and `serial` filters are **not** called
- Name filter empty, `filter(cf_inventory_hostname=identifier)` hits → that device returned
- Name + custom-field filters empty, `filter(serial=identifier)` hits → that device returned
- All three filters return empty → returns `None`
- Filter returns multiple devices → first element of `list(devices)` returned

### `process_netbox_webhook()` — `api.py:588`

Patch `osism.api.reconciler.run` (assert on `.delay`) and `osism.utils.nb` (only needed for the interfaces branch). Build real `WebhookNetboxData` instances (Celery is not involved beyond the patched `.delay`, no broker needed).

- `data["url"]` contains `"devices"`, tags include `"Managed by OSISM"`, `custom_fields["device_type"] == "server"` → `reconciler.run.delay()` called exactly once
- Managed device with `device_type == "switch"` → no `.delay()` call (TODO branch, log only)
- Managed device with `custom_fields["device_type"]` missing or `None` → falls back to `"node"` → no branch matches, no `.delay()` call
- `url` contains `"interfaces"` → device fetched via `utils.nb.dcim.devices.get(id=data["device"]["id"])`; tags/custom_fields taken from the fetched device; with `"Managed by OSISM"` in tags and `custom_fields["device_type"]` set → log-only branch, no `.delay()`; assert `nb.dcim.devices.get` called with the right `id`
- Unknown `url` (neither `devices` nor `interfaces`) → warning logged, early `return`, no NetBox access, no `.delay()`
- Tags without `"Managed by OSISM"` → "Ignoring change for unmanaged device" path, no `.delay()`
- `data` without `url`/`name` keys → `KeyError` propagates (the endpoint wraps this into a 500 — covered in the companion issue)

### `LogConfig` — `api.py:77`

- Defaults: `LOGGER_NAME == "osism"`, `LOG_LEVEL == "INFO"`, `version == 1`, `disable_existing_loggers is False`
- `model_dump()` yields a dict usable by `logging.config.dictConfig` — contains `formatters["default"]`, `handlers["default"]` (class `logging.StreamHandler`, stream `ext://sys.stderr`) and the loggers `osism`, `api`, `uvicorn`, `uvicorn.error`, `uvicorn.access`
- `dictConfig(LogConfig().model_dump())` does not raise (this already runs at import time, but make it explicit)

### Pydantic models — `api.py:54–374`

`NotificationBaremetal` (`api.py:54`), `WebhookNetboxResponse` (`api.py:63`), `WebhookNetboxData` (`api.py:67`), `DeviceSearchResult` (`api.py:150`), `BaremetalNode` (`api.py:155`), `BaremetalNodesResponse` (`api.py:210`), `BaremetalNodeNetboxInfo` (`api.py:215`), `BaremetalNodesNetboxRequest` (`api.py:228`), `BaremetalNodesNetboxResponse` (`api.py:232`), `BaremetalPort` (`api.py:238`), `BaremetalPortsResponse` (`api.py:247`), `BaremetalNodeParameters` (`api.py:252`), `SinkResponse` (`api.py:314`), `HostsResponse` (`api.py:318`), `HostvarEntry` (`api.py:325`), `HostvarsResponse` (`api.py:330`), `HostvarSingleResponse` (`api.py:336`), `FactEntry` (`api.py:342`), `FactsResponse` (`api.py:347`), `FactSingleResponse` (`api.py:354`), `SearchResultEntry` (`api.py:361`), `SearchResponse` (`api.py:368`). Use parametrized tests; no patching needed.

- `NotificationBaremetal`: valid payload accepted; `message_id` coerced to `UUID`; non-UUID `message_id` → `ValidationError`; each missing required field → `ValidationError`
- `WebhookNetboxData`: ISO-8601 string for `timestamp` coerced to `datetime.datetime`; `request_id` coerced to `UUID`; invalid values → `ValidationError`
- `BaremetalNode`: instantiable with no arguments (all fields optional); defaults `traits == []`, `properties == {}`, `extra == {}`; `default_factory` produces independent objects per instance (mutating one instance's `traits` does not leak into another)
- `BaremetalNodeParameters`, `BaremetalNodeNetboxInfo`, `BaremetalPort`, `DeviceSearchResult`: instantiable with no/partial arguments, optional fields default to `None`
- Required-field models (`SinkResponse`, `HostsResponse`, `HostvarEntry`, `HostvarsResponse`, `HostvarSingleResponse`, `FactEntry`, `FactsResponse`, `FactSingleResponse`, `SearchResultEntry`, `SearchResponse`, `BaremetalNodesResponse`, `BaremetalPortsResponse`, `BaremetalNodesNetboxRequest`, `BaremetalNodesNetboxResponse`, `WebhookNetboxResponse`): missing required field → `ValidationError`; happy-path round trip via `model_dump()`
- `HostvarEntry.value` / `FactEntry.value` / `SearchResultEntry.value` are `Any` → accept str, dict, list, `None`
- `BaremetalNodesNetboxResponse`: plain dicts as values are coerced to `BaremetalNodeNetboxInfo`

## Mocking hints

- `fastapi`/`pydantic`/`starlette` are runtime dependencies (`requirements.txt`; CI installs the package itself via `pipenv run pip install .`, see `playbooks/test-unit.yml`) — nothing extra to install for this issue. The `httpx`-based `TestClient` is only needed in the companion endpoint issue.
- Importing `osism.api` triggers the `EventBridge()` Redis-connect attempt described above; the exception is caught internally, tests need no Redis. No patching required, just be aware of the logged error.
- For `utils.nb` use `monkeypatch.setattr(osism.utils, "nb", fake_nb, raising=False)`; plain `mocker.patch("osism.utils.nb", ...)` would first call the module-level `__getattr__` and try to build a real NetBox connection.
- A minimal valid `WebhookNetboxData` for `process_netbox_webhook` tests:
  ```python
  WebhookNetboxData(
      username="admin",
      data={
          "url": "/api/dcim/devices/1/",
          "name": "sw1",
          "tags": [{"name": "Managed by OSISM"}],
          "custom_fields": {"device_type": "server"},
      },
      snapshots={},
      event="updated",
      timestamp="2026-01-01T00:00:00Z",
      model="device",
      request_id="f8a6e0a4-3c1b-4f53-9c5f-2f4f3b2a1e00",
  )
  ```
- For the interfaces branch set `data={"url": "/api/dcim/interfaces/5/", "name": "eth0", "device": {"id": 1}}` and give the fake device `tags=["Managed by OSISM"]` plus `custom_fields={"device_type": "server"}`.
- `reconciler.run` is a Celery task object; patch the whole attribute (`mocker.patch("osism.api.reconciler.run")`) and assert on `.delay` — no broker involved.

## Definition of Done

- [ ] `tests/unit/test_api_helpers.py` created
- [ ] `pytest --cov=osism.api` shows all lines of `_mask_inventory_secrets`, `find_device_by_identifier`, `process_netbox_webhook` and the model definitions covered (module-wide ≥ 90 % is reached together with the companion endpoint issue)
- [ ] `pipenv run pytest tests/unit/test_api_helpers.py` passes locally

## Dependencies

- Companion issue: *Unit tests for `osism/api.py` — HTTP & WebSocket endpoints (FastAPI TestClient)* (Tier 7, split 2/2).

berendt:
Adds `tests/unit/test_api_helpers.py` covering the module-level helpers and Pydantic models of the OSISM API server (`osism/api.py`), as specified in #2359:

- **`_mask_inventory_secrets()`** — secret-key matching (`password`/`secret` substrings, `ironic_osism_` prefix, using the real `_is_secret_key`), Ansible Vault values (including leading whitespace, and the marker mid-string staying unchanged), nested dict recursion, secret keys with dict values masked as a whole, pass-through of non-dict/non-string values and non-string keys, non-traversal of lists, empty dict, input immutability.
- **`find_device_by_identifier()`** — lookup precedence `name` → `cf_inventory_hostname` → `serial` with call-order assertions, early `None` return without a NetBox connection, first-match semantics for multiple results. `utils.nb` is injected via `patch.dict("osism.utils.__dict__", ...)` to avoid triggering the lazy module `__getattr__`, matching the existing pattern in `tests/unit/commands/`.
- **`process_netbox_webhook()`** — `reconciler.run.delay()` dispatched exactly once for managed servers; no dispatch for switches, interfaces (device fetched via `nb.dcim.devices.get(id=...)`, asserted), `device_type` fallback to `"node"`, unknown URLs (early return without NetBox access), and unmanaged devices; `KeyError` propagation for payloads without `url`/`name`.
- **`LogConfig`** — defaults, `model_dump()` structure (formatter, `StreamHandler` on `ext://sys.stderr`, all five loggers), `dictConfig` compatibility.
- **Pydantic models** — `NotificationBaremetal` and `WebhookNetboxData` validation and coercion (UUID, datetime, per-field required checks), `BaremetalNode` optional fields and `default_factory` independence, all-optional models, and parametrized required-field enforcement plus `model_dump()` round trips for the response models, including dict-to-`BaremetalNodeNetboxInfo` coercion and `Any`-typed value fields.

Endpoint behavior (HTTP status codes, masking applied in responses) is out of scope here and belongs to the companion endpoint issue (Tier 7, split 2/2).

Closes #2359

### DIFF
--- a/tests/unit/test_api_helpers.py
+++ b/tests/unit/test_api_helpers.py
@@ -1,0 +1,541 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for the module-level helpers and Pydantic models in :mod:`osism.api`.
+
+Covers ``_mask_inventory_secrets``, ``find_device_by_identifier``,
+``process_netbox_webhook``, ``LogConfig`` and the request/response models.
+The HTTP and WebSocket endpoints are exercised separately with the FastAPI
+``TestClient`` and are out of scope here.
+
+Importing :mod:`osism.api` has import-time side effects: it applies
+``dictConfig(LogConfig().model_dump())`` and pulls in
+``osism.services.event_bridge``, whose module-level ``EventBridge()``
+instantiation attempts a Redis connection. That failure is caught and only
+logged, so these tests need no Redis.
+
+``osism.utils`` materializes its ``nb`` NetBox connection lazily via a module
+``__getattr__``; tests therefore inject fakes with
+``patch.dict("osism.utils.__dict__", {"nb": ...})`` instead of
+``mocker.patch("osism.utils.nb")``, which would trigger a real connection
+attempt when reading the original attribute.
+"""
+
+import copy
+import datetime
+from logging.config import dictConfig
+from unittest.mock import MagicMock, call, patch
+from uuid import UUID
+
+import pytest
+from pydantic import ValidationError
+
+from osism import api
+
+# ---------------------------------------------------------------------------
+# _mask_inventory_secrets
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    "key",
+    [
+        "password",
+        "admin_password",
+        "ADMIN_Password",
+        "secret",
+        "keystone_secret_key",
+        "ironic_osism_user",
+    ],
+)
+def test_mask_inventory_secrets_masks_secret_keys(key):
+    assert api._mask_inventory_secrets({key: "value"}) == {key: "***"}
+
+
+def test_mask_inventory_secrets_masks_vault_encrypted_values():
+    data = {"var": "$ANSIBLE_VAULT;1.1;AES256\n61323964"}
+    assert api._mask_inventory_secrets(data) == {"var": "***"}
+
+
+def test_mask_inventory_secrets_masks_vault_values_with_leading_whitespace():
+    data = {"var": "  $ANSIBLE_VAULT;1.1;AES256\n61323964"}
+    assert api._mask_inventory_secrets(data) == {"var": "***"}
+
+
+def test_mask_inventory_secrets_keeps_vault_marker_inside_value():
+    data = {"var": "see $ANSIBLE_VAULT;1.1;AES256 for details"}
+    assert api._mask_inventory_secrets(data) == dict(data)
+
+
+def test_mask_inventory_secrets_recurses_into_nested_dicts():
+    data = {"outer": {"inner": {"db_password": "hunter2", "port": 5432}}}
+    assert api._mask_inventory_secrets(data) == {
+        "outer": {"inner": {"db_password": "***", "port": 5432}}
+    }
+
+
+def test_mask_inventory_secrets_masks_dict_value_of_secret_key_as_whole():
+    data = {"secrets": {"inner": "value"}}
+    assert api._mask_inventory_secrets(data) == {"secrets": "***"}
+
+
+@pytest.mark.parametrize("value", [42, [1, 2], None, True])
+def test_mask_inventory_secrets_passes_through_non_dict_non_string_values(value):
+    assert api._mask_inventory_secrets({"key": value}) == {"key": value}
+
+
+def test_mask_inventory_secrets_recurses_into_lists():
+    data = {"hosts": [{"db_password": "hunter2", "port": 5432}]}
+    assert api._mask_inventory_secrets(data) == {
+        "hosts": [{"db_password": "***", "port": 5432}]
+    }
+
+
+def test_mask_inventory_secrets_masks_vault_values_inside_lists():
+    data = {"vars": ["plain", "$ANSIBLE_VAULT;1.1;AES256\n61323964"]}
+    assert api._mask_inventory_secrets(data) == {"vars": ["plain", "***"]}
+
+
+def test_mask_inventory_secrets_recurses_into_nested_lists():
+    data = {"groups": [[{"api_secret": "token"}]]}
+    assert api._mask_inventory_secrets(data) == {"groups": [[{"api_secret": "***"}]]}
+
+
+def test_mask_inventory_secrets_empty_dict():
+    assert api._mask_inventory_secrets({}) == {}
+
+
+def test_mask_inventory_secrets_passes_through_non_string_keys():
+    assert api._mask_inventory_secrets({42: "value"}) == {42: "value"}
+
+
+def test_mask_inventory_secrets_does_not_mutate_input():
+    data = {"db_password": "hunter2", "nested": {"api_secret": "token"}}
+    original = copy.deepcopy(data)
+    api._mask_inventory_secrets(data)
+    assert data == original
+
+
+# ---------------------------------------------------------------------------
+# find_device_by_identifier
+# ---------------------------------------------------------------------------
+
+
+def make_nb(name_result=(), cf_result=(), serial_result=()):
+    """Build a fake NetBox client whose device filter answers by kwarg."""
+    nb = MagicMock()
+
+    def filter_devices(**kwargs):
+        if "name" in kwargs:
+            return list(name_result)
+        if "cf_inventory_hostname" in kwargs:
+            return list(cf_result)
+        if "serial" in kwargs:
+            return list(serial_result)
+        raise AssertionError(f"unexpected filter arguments: {kwargs}")
+
+    nb.dcim.devices.filter.side_effect = filter_devices
+    return nb
+
+
+def test_find_device_returns_none_without_netbox_connection():
+    with patch.dict("osism.utils.__dict__", {"nb": None}):
+        assert api.find_device_by_identifier("node-1") is None
+
+
+def test_find_device_by_name():
+    device = MagicMock()
+    nb = make_nb(name_result=[device])
+    with patch.dict("osism.utils.__dict__", {"nb": nb}):
+        assert api.find_device_by_identifier("node-1") is device
+    assert nb.dcim.devices.filter.call_args_list == [call(name="node-1")]
+
+
+def test_find_device_by_inventory_hostname():
+    device = MagicMock()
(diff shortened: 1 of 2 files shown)


tests/unit/test_api_helpers.py
```diff
@@ -0,0 +1,529 @@
...
+    data = {"secrets": {"inner": "value"}}
+    assert api._mask_inventory_secrets(data) == {"secrets": "***"}
+
+
+@pytest.mark.parametrize("value", [42, [1, 2], None, True])
+def test_mask_inventory_secrets_passes_through_non_dict_non_string_values(value):
+    assert api._mask_inventory_secrets({"key": value}) == {"key": value}
+
+
+def test_mask_inventory_secrets_does_not_traverse_lists():
```
ideaship: This test asserts, as expected behavior, that a secret nested inside a list is returned **unmasked**:

```python
data = {"hosts": [{"db_password": "hunter2"}]}
assert api._mask_inventory_secrets(data) == {"hosts": [{"db_password": "hunter2"}]}
```

That codifies a production leak. `_mask_inventory_secrets` (`osism/api.py:35-51`) recurses into `dict` but has no `list` branch, and it masks the `ansible-inventory --host` output returned by the API at `api.py:833`, `903`, and `1104`. Ansible hostvars routinely contain lists of dicts (e.g. `users: [{name: admin, password: ...}]`), so any secret nested in a list is returned verbatim through `/hostvars` and the search endpoints — the masking layer that exists to prevent exactly this is silently bypassed.

The sibling masker `_mask_secrets_inplace` (`osism/tasks/conductor/utils.py:314-320`) already has an `isinstance(obj, list)` branch, so the missing list handling here is an oversight, not an intentional design choice.

Fix `_mask_inventory_secrets` to traverse lists (mirroring `_mask_secrets_inplace`), and flip this assertion to expect `{"hosts": [{"db_password": "***"}]}`. Keep the production fix in its own commit, separate from the test changes. As written, the test would actively resist that fix.

ideaship (changes requested):

ideaship (approved):

PR #2464, opened 2026-07-13
Add unit tests for task lifecycle commands

### LINKED ISSUE #2362: Unit tests for osism/commands/ — task lifecycle (task, worker, reconciler, lock, service, container, configuration, set, noset)
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul infrastructure). Part of **Tier 8** (#2199). Combined issue for the nine small task-lifecycle CLI command modules under `osism/commands/` (~579 LOC total): `task.py` (20 LOC, revoke Celery tasks), `worker.py` (57 LOC, start Celery workers), `reconciler.py` (71 LOC, deprecated worker runner + inventory sync), `lock.py` (91 LOC, task-lock management), `service.py` (100 LOC, service launcher incl. inventory watchdog), `container.py` (57 LOC, remote `docker` commands via SSH), `configuration.py` (39 LOC, configuration sync), `set.py` and `noset.py` (72 LOC each, maintenance/bootstrap state toggles). All are thin cliff `Command` classes orchestrating Celery `.delay()` calls, `subprocess`, and the Redis task lock — mock-heavy but individually small.

## Scope

Add `tests/unit/commands/test_task.py`, `test_worker.py`, `test_lock.py`, `test_service.py`, `test_container.py`, `test_configuration.py`, `test_set.py`, `test_noset.py` and **extend** the existing `tests/unit/commands/test_reconciler.py`.

Already covered (do not duplicate): `tests/unit/commands/test_reconciler.py` has one test — `Sync.take_action` returns `1` when `utils.fetch_task_output` raises `TimeoutError`. Only the gap cases listed below are in scope for that module.

Follow the established pattern from `test_reconciler.py` / `test_wait.py`: `cmd = module.Class(MagicMock(), MagicMock())`, then `parsed_args = cmd.get_parser("test").parse_args([...])`.

This issue may be split further during implementation if it turns out too large (natural sub-chunks: `set`/`noset`, `lock`, `worker`/`service`/`reconciler`) — Tier 3 issues did the same.

### `Revoke` — `task.py:6`

Patch `celery.Celery` at the source (lazy import inside `take_action`).

- Parser: positional `task` with `nargs=1` → `parse_args(["abc"])` yields `parsed_args.task == ["abc"]`
- `take_action`: `Celery` constructed with name `"task"`, `config_from_object` called with `osism.tasks.Config`, `app.control.revoke` called with `terminate=True`
- Quirk to pin down: `task = parsed_args.task` is the **one-element list** (`nargs=1` is never unpacked), so `control.revoke` receives `["abc"]`, not `"abc"` — assert current behavior and note the latent bug in a comment

### `Run` (worker) — `worker.py:11`

Patch `osism.commands.worker.subprocess.Popen`, `osism.commands.worker.utils.check_task_lock_and_exit`; env via `monkeypatch`.

**`get_parser()` — `worker.py:12`**

- Default `--number-of-workers` is `min(cpu_count, 4)` when `OSISM_CELERY_CONCURRENCY` is unset (use `monkeypatch.delenv(..., raising=False)`; patch `multiprocessing.cpu_count` for determinism)
- `OSISM_CELERY_CONCURRENCY=7` set **before** `get_parser()` → default is `7` (the env var is read at parser-build time)

**`take_action()` — `worker.py:30`**

- `utils.check_task_lock_and_exit` is called before anything else
- Queue mapping: `openstack` / `netbox` / `conductor` → `celery -A osism.tasks.<queue> worker -n <queue> ... -Q <queue>`
- `osism-kubernetes` → both tasks module and queue become `kubernetes`
- Suffix stripping: `kolla-ansible` → `-A osism.tasks.kolla ... -Q kolla-ansible`; `ceph-ansible` → `osism.tasks.ceph`
- Special case `osism-ansible` → tasks `"osism"` triggers the dedicated branch: `-A osism.tasks.ansible worker -n osism-ansible ... -Q osism-ansible`
- `-c <n>` in the command string reflects `--number-of-workers 3`
- `p.wait()` is called on the `Popen` return value
- Edge documenting current behavior: an unknown short type (e.g. `"foo"`) yields `"foo"[:-8] == ""` → command `celery -A osism.tasks. worker ...` — assert as-is with a comment (latent bug, no validation)

### `Run` (reconciler, deprecated) — `reconciler.py:13` — gap

Patch `osism.commands.reconciler.subprocess.Popen`; env via `monkeypatch`.

- `take_action` (`reconciler.py:14`) logs the deprecation warning ("deprecated ... Use osism service reconciler")
- Concurrency: `OSISM_CELERY_CONCURRENCY` unset → `min(cpu_count, 4)`; set to `2` → `-c 2` in the command string
- Command string is `celery -A osism.tasks.reconciler worker -n reconciler --loglevel=INFO -Q reconciler -c <n>` and `p.wait()` is called

### `Sync` (reconciler) — `reconciler.py:30` — gaps only

Patch `osism.tasks.reconciler.run.delay` (lazy import at the call site), `osism.commands.reconciler.utils.fetch_task_output`, `osism.commands.reconciler.utils.check_task_lock_and_exit`. Timeout → `1` is already covered.

- Wait path (default): `delay` called with `publish=True`, `fetch_task_output` called with the task id and `timeout=300`; its return value is returned by `take_action`
- `--no-wait`: `delay` called with `publish=False`, `fetch_task_output` **not** called, returns `None`, "No more output" logged
- `--task-timeout 60` → `fetch_task_output` called with `timeout=60`
- `OSISM_TASK_TIMEOUT=600` set before `get_parser()` → parsed default is `int` 600 (argparse applies `type=int` to string defaults)
- `check_task_lock_and_exit` invoked before scheduling

### `Lock` — `lock.py:9`

Patch `osism.commands.lock.utils.is_task_locked`, `osism.commands.lock.utils.set_task_lock`.

- Already locked (`{"locked": True, "user": "alice", "timestamp": "..."}`) → warning logged, returns `None`, `set_task_lock` **not** called; with `reason` present an additional "Existing reason" warning is logged
- Not locked (`None` or `{"locked": False}`), `set_task_lock` returns `True` → returns `None`; `--user bob --reason maintenance` passed through to `set_task_lock("bob", "maintenance")`
- `--user` omitted → defaults to `settings.OPERATOR_USER` (`"dragon"` unless `OSISM_OPERATOR_USER` set — patch `osism.commands.lock.settings.OPERATOR_USER` for determinism); note the default is also baked into the `--user` help string at parser-build time
- `set_task_lock` returns `False` → "Failed to set task lock" logged, returns `1`

### `Unlock` — `lock.py:51`

Patch `osism.commands.lock.utils.is_task_locked`, `osism.commands.lock.utils.remove_task_lock`.

- Not locked (`None` or `{"locked": False}`) → "Tasks are not currently locked", returns `None`, `remove_task_lock` **not** called
- Locked, `remove_task_lock` returns `True` → returns `None`, log mentions previous user/timestamp; previous `reason` logged only when present
- `remove_task_lock` returns `False` → returns `1`

### `LockStatus` — `lock.py:78`

Patch `osism.commands.lock.utils.is_task_locked`.

- Locked with reason → "Tasks are LOCKED by ..." plus "Reason: ..." logged
- Locked without reason and with missing `user`/`timestamp` keys → falls back to `"unknown"`
- Not locked → "Tasks are UNLOCKED"

### `Run` (service) — `service.py:13`

Patch `osism.commands.service.subprocess.Popen`, `osism.commands.service.utils.check_task_lock_and_exit`; env via `monkeypatch`.

- `take_action` (`service.py:19`) calls `check_task_lock_and_exit` first for every service type
- `api` → `uvicorn osism.api:app --host 0.0.0.0 --port 8000`, `p.wait()` called
- `listener` → `python3 -c 'from osism.services import listener; listener.main()'`
- `beat` → exactly **seven** `Popen` calls, one per task module (`ansible`, `ceph`, `conductor`, `kolla`, `netbox`, `openstack`, `reconciler`), each with `--broker=redis://redis beat -s /tmp/celerybeat-schedule-<module>.db`; `wait()` called on every process (inspect `call_args_list`)
- `flower` → `celery --broker=redis://redis flower`
- `reconciler` → same concurrency/env logic and command string as `reconciler.py:14`
- Unknown service type (e.g. `"bogus"`) → no `Popen` call, returns `None` (silent no-op; assert current behavior)

**`watchdog` branch**

Patch `watchdog.observers.polling.PollingObserver` at the source (lazy import inside the branch) and make `osism.commands.service.time.sleep` raise a sentinel exception to escape `while True`.

- Observer scheduled on `/opt/configuration/inventory` with `recursive=True`, `start()` called
- The `finally` block calls `observer.stop()` and `observer.join()` even when the loop exits via the sentinel exception (use `pytest.raises`)
- The event handler's `on_any_event` is rebound to `self.watchdog_inventory_on_any_event`

### `Run.watchdog_inventory_on_any_event()` — `service.py:97`

Patch `osism.tasks.reconciler.run.delay` (lazy import).

- Any event object → `reconciler.run.delay()` called exactly once with no arguments

### `Run` (container) — `container.py:14`

Patch `osism.commands.container.subprocess.call`, `osism.commands.container.ensure_known_hosts_file`, `osism.commands.container.prompt` (imported at module top from `prompt_toolkit`).

- Parser (`container.py:15`): `host` `nargs=1` plus `command` as `argparse.REMAINDER` → `parse_args(["node1", "ps", "-a"])` yields `command == ["ps", "-a"]`
- Non-interactive: command given → exactly one `subprocess.call` whose command string contains `/usr/bin/ssh -i /ansible/secrets/id_rsa.operator`, `-o UserKnownHostsFile=/share/known_hosts` (`KNOWN_HOSTS_PATH`), `<OPERATOR_USER>@node1` and `docker ps -a`; `prompt` not called
- `ensure_known_hosts_file` returns `False` → warning logged, SSH call still executed
- Interactive: empty `command` → `prompt` loop; `side_effect=["ps", "exit"]` → one `subprocess.call` with `docker ps`, then loop terminates
- Loop exit keywords: `"Exit"`, `"exit"`, `"EXIT"` each break the loop immediately (parametrize); `prompt` returning the keyword first → zero `subprocess.call` invocations
- Operator user comes from `settings.OPERATOR_USER` — patch `osism.commands.container.settings.OPERATOR_USER` for a deterministic assertion

### `Sync` (configuration) — `configuration.py:11`

Patch `osism.tasks.ansible.run.delay`, `osism.tasks.handle_task` (both lazily imported inside `take_action`), `osism.commands.configuration.utils.check_task_lock_and_exit`.

- Parser (`configuration.py:12`): REMAINDER `arguments` → `parse_args(["-e", "foo=1"])` forwards `["-e", "foo=1"]`
- `take_action` (`configuration.py:19`): `delay` called with `("manager", "configuration", arguments)` and `auto_release_time=60`
- `handle_task` called with the task object, `True`, and timeout `60`; its return value is returned as the command's rc (e.g. `0` and `2`)
- `check_task_lock_and_exit` invoked before scheduling
- Quirk to pin down: the third `handle_task` argument is the Python **builtin `format` function**, not a format string like `"log"` (`rc = handle_task(t, True, format, 60)`) — assert current behavior and note the latent bug in a comment

### `Maintenance` / `Bootstrap` — `set.py:9` / `set.py:42`

Patch `osism.tasks.ansible.run.delay` (lazy import), `osism.commands.set.utils.check_task_lock_and_exit`.

- `Maintenance.take_action` (`set.py:20`): host from `nargs=1`; `delay` called with `("generic", "state-maintenance", ["-e status=True", "-l node1"])`
- `Bootstrap.take_action` (`set.py:53`): same but playbook `"state-bootstrap"`
- `check_task_lock_and_exit` called before scheduling in both
- Parametrize over the two classes — they differ only in the playbook name

### `NoMaintenance` / `NoBootstrap` — `noset.py:9` / `noset.py:42`

Patch `osism.tasks.ansible.run.delay`, `osism.commands.noset.utils.check_task_lock_and_exit`.

- `NoMaintenance.take_action` (`noset.py:20`): `delay` called with `("generic", "state-maintenance", ["-e status=False", "-l node1"])`
- `NoBootstrap.take_action` (`noset.py:53`): playbook `"state-bootstrap"`, `-e status=False`
- `check_task_lock_and_exit` called before scheduling in both
- Mirror the parametrized structure of `test_set.py` (`status=True` vs. `status=False` is the only difference between the module pairs)

## Mocking hints

- Instantiate commands as `Cmd(MagicMock(), MagicMock())` and build args via `cmd.get_parser("test").parse_args([...])`, matching the existing `tests/unit/commands/test_reconciler.py`.
- `utils.check_task_lock_and_exit` calls `exit(1)` when locked — always patch it at the command module (`osism.commands.<mod>.utils.check_task_lock_and_exit`) and only assert it was invoked; the lock logic itself is covered by the `osism.utils` tier. Do not test the locked/`SystemExit` path here.
- Lazy imports inside `take_action` mean patching at the **source** module: `osism.tasks.ansible.run.delay`, `osism.tasks.reconciler.run.delay`, `osism.tasks.handle_task`, `celery.Celery`. The Celery task objects are imported attributes, so `mocker.patch("osism.tasks.ansible.run.delay")` works without a broker — tests never execute the task bodies.
- `delay()` return values: a `MagicMock()` suffices; `Sync` only reads `t.task_id` / `t.id`.
- `subprocess.Popen` / `subprocess.call` are patched per command module; assert the exact shell command via `mock.call_args[0][0]` and `shell=True` where relevant.
- Env-dependent parser defaults (`OSISM_CELERY_CONCURRENCY`, `OSISM_TASK_TIMEOUT`) are evaluated at `get_parser()` time — set/del env vars with `monkeypatch` **before** building the parser. argparse applies `type=int` to string defaults, so env values arrive as `int`.
- `set` shadows a builtin module name — import as `from osism.commands import set as set_cmd` (and `noset` normally).
- For the watchdog loop, a custom `class _Stop(Exception)` as `time.sleep` side effect keeps the test independent of `KeyboardInterrupt` handling.

## Definition of Done

- [ ] `tests/unit/commands/test_task.py`, `test_worker.py`, `test_lock.py`, `test_service.py`, `test_container.py`, `test_configuration.py`, `test_set.py`, `test_noset.py` created
- [ ] `tests/unit/commands/test_reconciler.py` extended with the listed gap cases (no duplication of the existing timeout test)
- [ ] `pytest --cov` shows ≥ 90 % for each module in scope
- [ ] `pipenv run pytest tests/unit/commands/` passes locally

## Dependencies

berendt:
Implements #2362 (part of the Tier 8 meta issue #2199).

Adds unit tests for the nine small task-lifecycle CLI command modules under `osism/commands/`, following the established pattern from `tests/unit/commands/test_reconciler.py` / `test_wait.py` (`Cmd(MagicMock(), MagicMock())` + `get_parser("test").parse_args([...])`).

## New test files

- `test_task.py` — parser contract and Celery revoke with `terminate=True`; pins the latent bug that the one-element `nargs=1` list reaches `control.revoke`
- `test_worker.py` — concurrency defaults (CPU count capped at 4, `OSISM_CELERY_CONCURRENCY` read at parser-build time), worker-type → tasks-module/queue mapping including the `osism-ansible` special case and suffix stripping, `-c` propagation, and the unvalidated unknown-type command pinned as current behavior
- `test_lock.py` — `Lock`/`Unlock`/`LockStatus` flows: already-locked warnings (with/without reason), `--user`/`--reason` pass-through, `OPERATOR_USER` default (also baked into the `--user` help at parser-build time), unknown-key fallbacks, and non-zero returns when the lock backend fails
- `test_service.py` — exact subprocess command lines for `api`, `listener`, `beat` (seven schedulers, one per task module), `flower`, and `reconciler` (env/CPU concurrency logic); watchdog observer setup on `/opt/configuration/inventory`, cleanup via `finally`, `on_any_event` rebinding, and the inventory event triggering `reconciler.run.delay()`; unknown service type as silent no-op
- `test_container.py` — parser `host`/REMAINDER split, one-shot SSH `docker` execution (key, known-hosts option, operator user), known-hosts init failure warning, interactive prompt loop, and `Exit`/`exit`/`EXIT` loop termination
- `test_configuration.py` — REMAINDER forwarding, `delay(..., auto_release_time=60)`, rc propagation via `handle_task`; pins the latent builtin-`format` argument
- `test_set.py` / `test_noset.py` — state playbook scheduling parametrized over the `Maintenance`/`Bootstrap` command pairs (`status=True` vs. `status=False`)

## Extended

- `test_reconciler.py` — deprecation warning and concurrency of the deprecated worker runner; `Sync` wait/no-wait paths, `--task-timeout` forwarding, and the `OSISM_TASK_TIMEOUT` parser default (the existing timeout test is kept unchanged)

All Celery task calls, subprocesses, and the Redis task lock are mocked at the command module (lazy imports patched at the source module); the tests also assert that `check_task_lock_and_exit` runs before any task is scheduled or process is started.

Closes #2362

### DIFF
--- a/tests/unit/commands/test_configuration.py
+++ b/tests/unit/commands/test_configuration.py
@@ -1,0 +1,61 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for the ``osism configuration sync`` command.
+
+``Sync`` schedules the ``configuration`` playbook on the manager environment
+and returns the task's exit code via ``handle_task``.
+"""
+
+from unittest.mock import patch
+
+import pytest
+
+from osism.commands import configuration
+
+from ._helpers import assert_not_called_before_lock_check, parse_args
+
+
+def test_parser_forwards_mixed_positional_and_option_arguments():
+    _, parsed_args = parse_args(configuration.Sync, ["playbook", "-e", "foo=1"])
+    assert parsed_args.arguments == ["playbook", "-e", "foo=1"]
+
+
+def test_parser_forwards_leading_option_arguments():
+    # The parser delegates to ``parse_known_args`` (see ``_PassthroughParser``),
+    # so option-like tokens (``-e``, ``--limit``, ...) are forwarded verbatim
+    # even without a leading positional.
+    _, parsed_args = parse_args(configuration.Sync, ["-e", "foo=1"])
+    assert parsed_args.arguments == ["-e", "foo=1"]
+
+
+@pytest.mark.parametrize("rc", [0, 2])
+def test_take_action_schedules_configuration_sync_and_returns_rc(rc):
+    cmd, parsed_args = parse_args(configuration.Sync, ["playbook", "-e", "foo=1"])
+
+    with patch(
+        "osism.commands.configuration.utils.check_task_lock_and_exit"
+    ) as mock_check, patch("osism.tasks.ansible.run.delay") as mock_delay, patch(
+        "osism.tasks.handle_task", return_value=rc
+    ):
+        mock_check.side_effect = assert_not_called_before_lock_check(mock_delay)
+        result = cmd.take_action(parsed_args)
+
+    mock_check.assert_called_once()
+    mock_delay.assert_called_once_with(
+        "manager",
+        "configuration",
+        ["playbook", "-e", "foo=1"],
+        auto_release_time=60,
+    )
+    assert result == rc
+
+
+def test_take_action_waits_with_log_format():
+    cmd, parsed_args = parse_args(configuration.Sync, [])
+
+    with patch("osism.commands.configuration.utils.check_task_lock_and_exit"), patch(
+        "osism.tasks.ansible.run.delay"
+    ) as mock_delay, patch("osism.tasks.handle_task", return_value=0) as mock_handle:
+        cmd.take_action(parsed_args)
+
+    mock_handle.assert_called_once_with(mock_delay.return_value, True, "log", 60)
--- a/tests/unit/commands/test_container.py
+++ b/tests/unit/commands/test_container.py
@@ -1,0 +1,85 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for the ``osism container`` command.
+
+``Run`` executes ``docker`` commands on a remote host via SSH, either from
+the command line (one-shot) or from an interactive prompt loop.
+"""
+
+from unittest.mock import patch
+
+import pytest
+
+from osism.commands import container
+
+from ._helpers import parse_args
+
+
+def _run(args, *, prompt_side_effect=None, known_hosts=True):
+    cmd, parsed_args = parse_args(container.Run, args)
+
+    with patch("osism.commands.container.subprocess.call") as mock_call, patch(
+        "osism.commands.container.ensure_known_hosts_file", return_value=known_hosts
+    ), patch(
+        "osism.commands.container.prompt", side_effect=prompt_side_effect
+    ) as mock_prompt, patch(
+        "osism.commands.container.settings.OPERATOR_USER", "testuser"
+    ):
+        cmd.take_action(parsed_args)
+
+    return mock_call, mock_prompt
+
+
+def _ssh_command(mock_call):
+    """Flatten the ``subprocess.call`` invocation to a single command string.
+
+    The tests assert only the observable contract (ssh target and docker
+    command), so they hold whether the command is passed as a shell string or
+    as an argv vector.
+    """
+    command = mock_call.call_args[0][0]
+    return command if isinstance(command, str) else " ".join(command)
+
+
+def test_parser_splits_host_and_remainder_command():
+    _, parsed_args = parse_args(container.Run, ["node1", "ps", "-a"])
+    assert parsed_args.host == ["node1"]
+    assert parsed_args.command == ["ps", "-a"]
+
+
+def test_one_shot_command_runs_docker_via_ssh():
+    mock_call, mock_prompt = _run(["node1", "ps", "-a"])
+
+    mock_prompt.assert_not_called()
+    mock_call.assert_called_once()
+    ssh_command = _ssh_command(mock_call)
+    assert "testuser@node1" in ssh_command
+    assert ssh_command.endswith("docker ps -a")
+
+
+def test_known_hosts_failure_warns_but_still_connects(loguru_logs):
+    mock_call, _ = _run(["node1", "ps"], known_hosts=False)
+
+    assert any(
+        record["level"] == "WARNING" and "/share/known_hosts" in record["message"]
+        for record in loguru_logs
+    )
+    mock_call.assert_called_once()
+
+
+def test_interactive_prompt_runs_commands_until_exit():
+    mock_call, mock_prompt = _run(["node1"], prompt_side_effect=["ps", "exit"])
+
+    assert mock_prompt.call_count == 2
+    mock_call.assert_called_once()
+    ssh_command = _ssh_command(mock_call)
+    assert "testuser@node1" in ssh_command
+    assert ssh_command.endswith("docker ps")
+
+
+@pytest.mark.parametrize("keyword", ["Exit", "exit", "EXIT"])
+def test_interactive_exit_keywords_break_immediately(keyword):
+    mock_call, mock_prompt = _run(["node1"], prompt_side_effect=[keyword])
+
+    mock_prompt.assert_called_once()
+    mock_call.assert_not_called()
--- a/tests/unit/commands/test_reconciler.py
+++ b/tests/unit/commands/test_reconciler.py
@@ -2,19 +2,119 @@
 
 """Tests for the ``osism reconciler`` commands.
 
-These focus on the exit-code contract: a command must return a non-zero exit
-status when it gives up waiting for a task (a timeout is an operational
-failure), rather than falling through to an implicit success.
+``Run`` is the deprecated foreground worker runner; ``Sync`` schedules an
+inventory sync via Celery and (by default) waits for its output. The exit-code
+contract matters most: a command must return a non-zero exit status when it
+gives up waiting for a task (a timeout is an operational failure), rather than
+falling through to an implicit success.
 """
 
 from unittest.mock import MagicMock, patch
 
 from osism.commands import reconciler
 
+from ._helpers import assert_not_called_before_lock_check, parse_args
+
+
+def _run_worker(monkeypatch, *, env=None, cpu_count=16):
+    monkeypatch.delenv("OSISM_CELERY_CONCURRENCY", raising=False)
+    if env is not None:
+        monkeypatch.setenv("OSISM_CELERY_CONCURRENCY", env)
+
+    cmd, parsed_args = parse_args(reconciler.Run, [])
+
+    with patch(
+        "osism.commands.reconciler.utils.check_task_lock_and_exit"
+    ) as mock_check, patch(
+        "osism.commands.reconciler.subprocess.Popen"
+    ) as mock_popen, patch(
+        "osism.commands.reconciler.multiprocessing.cpu_count",
+        return_value=cpu_count,
+    ):
+        mock_check.side_effect = assert_not_called_before_lock_check(mock_popen)
+        cmd.take_action(parsed_args)
+
+    return mock_check, mock_popen
+
+
+def _run_sync(args, *, fetch_return=0):
+    cmd, parsed_args = parse_args(reconciler.Sync, args)
+
+    with patch(
+        "osism.commands.reconciler.utils.check_task_lock_and_exit"
+    ) as mock_check, patch("osism.tasks.reconciler.run.delay") as mock_delay, patch(
+        "osism.commands.reconciler.utils.fetch_task_output",
+        return_value=fetch_return,
+    ) as mock_fetch:
+        mock_check.side_effect = assert_not_called_before_lock_check(mock_delay)
+        result = cmd.take_action(parsed_args)
+
+    return result, mock_check, mock_delay, mock_fetch
+
+
+def test_run_logs_deprecation_warning(loguru_logs, monkeypatch):
+    _run_worker(monkeypatch)
+    assert any(
+        "deprecated" in record["message"]
+        and "osism service reconciler" in record["message"]
+        for record in loguru_logs
+    )
+
+
+def test_run_checks_task_lock_before_starting_worker(monkeypatch):
+    mock_check, mock_popen = _run_worker(monkeypatch)
+    mock_check.assert_called_once()
+    mock_popen.assert_called_once()
+
+
+def test_run_concurrency_defaults_to_cpu_count_capped_at_four(monkeypatch):
+    _, mock_popen = _run_worker(monkeypatch, cpu_count=16)
+    mock_popen.assert_called_once_with(
+        "celery -A osism.tasks.reconciler worker -n reconciler --loglevel=INFO -Q reconciler -c 4",
+        shell=True,
+    )
+    mock_popen.return_value.wait.assert_called_once_with()
+
+
+def test_run_concurrency_from_env(monkeypatch):
+    _, mock_popen = _run_worker(monkeypatch, env="2")
+    assert "-c 2" in mock_popen.call_args[0][0]
+
+
+def test_sync_waits_for_task_output_and_returns_its_result():
+    result, mock_check, mock_delay, mock_fetch = _run_sync([], fetch_return=0)
+
+    mock_check.assert_called_once()
+    mock_delay.assert_called_once_with(publish=True)
+    mock_fetch.assert_called_once_with(mock_delay.return_value.id, timeout=300)
+    assert result == 0
+
+
+def test_sync_no_wait_schedules_without_fetching_output(loguru_logs):
+    result, _, mock_delay, mock_fetch = _run_sync(["--no-wait"])
+
+    mock_delay.assert_called_once_with(publish=False)
+    mock_fetch.assert_not_called()
+    assert result is None
+    assert any("No more output" in record["message"] for record in loguru_logs)
+
+
+def test_sync_task_timeout_argument_is_forwarded():
+    _, _, _, mock_fetch = _run_sync(["--task-timeout", "60"])
+    mock_fetch.assert_called_once()
+    assert mock_fetch.call_args[1]["timeout"] == 60
+
+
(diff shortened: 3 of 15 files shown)


tests/unit/commands/test_service.py
```diff
@@ -0,0 +1,167 @@
...
+
+
+def test_reconciler_concurrency_from_env(monkeypatch):
+    monkeypatch.setenv("OSISM_CELERY_CONCURRENCY", "2")
+    _, _, mock_popen = _run_service(["reconciler"])
+    assert "-c 2" in mock_popen.call_args[0][0]
+
+
+@pytest.mark.latent_bug
+def test_unknown_service_type_is_a_silent_noop():
```
ideaship: **This pins an unknown service type passing without error.** The test is green only because `service.py` silently does nothing for an unrecognized type, and `latent_bug` does not exclude it from the default suite — so whoever adds the error path turns this test red and must rewrite it. Assert the intended behavior (unknown type should error) and fix `service.py` in a dedicated commit.

tests/unit/commands/test_worker.py
```diff
@@ -0,0 +1,107 @@
...
+    mock_popen.return_value.wait.assert_called_once_with()
+
+
+def test_take_action_respects_number_of_workers():
+    _, mock_popen = _run_worker(["--number-of-workers", "3", "openstack"])
+    assert "-c 3" in mock_popen.call_args[0][0]
+
+
+@pytest.mark.latent_bug
+def test_take_action_unknown_type_yields_empty_tasks_module():
```
ideaship: **This pins a broken command as intended output.** An unvalidated short type makes `queue[:-8]` empty, producing `celery -A osism.tasks. worker …`. Reject unknown worker types (or map them explicitly) in a dedicated commit, and assert the error rather than the malformed command.

tests/unit/commands/test_configuration.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def test_parser_forwards_positional_remainder_arguments():
+    # ``argparse.REMAINDER`` captures every token once a positional appears.
+    _, parsed_args = parse_args(configuration.Sync, ["playbook", "-e", "foo=1"])
+    assert parsed_args.arguments == ["playbook", "-e", "foo=1"]
+
+
+@pytest.mark.latent_bug
+def test_parser_rejects_leading_option_arguments():
```
ideaship: **This pins a parser that rejects the Ansible options its `arguments` remainder exists to forward.** `nargs=REMAINDER` only starts capturing at the first *positional*, so `osism configuration sync -e foo=1` is rejected — and since every forwardable option starts with `-` and there is no natural leading positional here, no option can be passed in practice. Switch to the `parse_known_args` approach already used in `openstack`, in a dedicated commit, and assert the option is forwarded.

tests/unit/commands/test_configuration.py
```diff
@@ -0,0 +1,70 @@
...
+    mock_delay.assert_called_once_with(
+        "manager",
+        "configuration",
+        ["playbook", "-e", "foo=1"],
+        auto_release_time=60,
+    )
+    assert result == rc
+
+
+@pytest.mark.latent_bug
```
ideaship: **This pins a real bug — the wrong value is passed.** Only this caller passes the builtin `format` instead of a format string (`manage.py`, `netbox.py` → `"script"`; `apply.py`, `validate.py` → `parsed_args.format`); it is inert today only because `wait=True` short-circuits before it is read, but `--no-wait` (as `reconciler.Sync` already has) would make the `else` branch print nothing. Fix production to pass `"log"` (or wire `--format`) in a dedicated commit and assert that; do not pin the builtin as intended.

tests/unit/commands/test_task.py
```diff
@@ -0,0 +1,38 @@
...
+from ._helpers import parse_args
+
+
+def test_parser_task_is_a_one_element_list():
+    _, parsed_args = parse_args(task.Revoke, ["abc"])
+    assert parsed_args.task == ["abc"]
+
+
+@pytest.mark.latent_bug
+def test_revoke_passes_task_list_instead_of_id():
```
ideaship: **This is the only `nargs=1` positional passed on without `[0]` unpacking** — `service.py:25`, `worker.py:34`, and `set`/`noset`/`container` all unpack. `control.revoke` accepts both a string and a list, so the current call is not broken; this is a sibling-consistency gap, not a bug. If uniformity is the goal, unpack here too — `app.control.revoke(task[0], terminate=True)` — in a dedicated commit, and drop the `latent_bug` marker; keep the behavioral assertion.

tests/unit/commands/test_reconciler.py
```diff
@@ -2,19 +2,108 @@
...
 """
 
 from unittest.mock import MagicMock, patch
 
 from osism.commands import reconciler
 
+from ._helpers import assert_not_called_before_lock_check, parse_args
+
+
+def _run_worker(monkeypatch, *, env=None, cpu_count=16):
```
ideaship, line 19: **The commit message's lock-ordering claim is false because the ordering is missing on this path.** `reconciler.Run.take_action` (`osism/commands/reconciler.py:14-27`) starts a worker subprocess with no `check_task_lock_and_exit()` — only `Sync` calls it (`:49`) — so the commit body's "before any task is scheduled or process is started" does not hold, and this `_run_worker` helper neither patches nor asserts the lock. It is a deprecated path (its replacement `service.Run` guards the lock), so severity is low. Pick one: **(a)** add the missing check in a dedicated commit — then the commit message is accurate as written and this test should assert the lock; or **(b)** leave the deprecated path and correct the commit message to name `reconciler.Run` as the exception rather than claiming universal coverage.

tests/unit/commands/test_container.py
```diff
@@ -0,0 +1,77 @@
...
+    assert parsed_args.command == ["ps", "-a"]
+
+
+def test_one_shot_command_runs_docker_via_ssh():
+    mock_call, mock_prompt = _run(["node1", "ps", "-a"])
+
+    mock_prompt.assert_not_called()
+    mock_call.assert_called_once()
+    ssh_command = mock_call.call_args[0][0]
+    assert mock_call.call_args[1] == {"shell": True}
```
ideaship: **These assertions would block a safe rewrite.** `call_args[1] == {"shell": True}` plus the exact command string would fail a future hardening to an argv vector (`subprocess.call([...])`, no `shell=True`) for no behavioral reason. Assert only the observable contract — ssh targets `testuser@node1` and runs `docker ps -a`. Separately, the `shell=True` interpolation in `osism/commands/container.py:47,55` (already `# FIXME`'d) is a pre-existing, codebase-wide hardening candidate for its own PR; it is operator-supplied input, so not a security blocker here.

ideaship (changes requested):

ideaship (approved):

PR #2472, opened 2026-07-16
manage images: default --cloud to admin

ideaship:
## Problem

A bare `osism manage images` fails on every standard OSISM deployment with:

```
ConfigException: Cloud openstack was not found.
```

The `Images` command defaulted `--cloud` to `openstack`, a profile that `cfg-cookiecutter`'s `clouds.yml` never ships — it provides `admin`, `admin-system`, and `octavia`. The `openstack` value was inherited from the standalone `openstack-image-manager` tool rather than OSISM's own operator-profile convention.

This diverged from the sibling subcommands, all of which already default to `admin`:

| Subcommand | old default | in shipped `clouds.yml`? |
|---|---|---|
| `manage images` | `openstack` | no → fails bare |
| `manage flavors` | `admin` | yes → works bare |
| `manage project *` | `admin` | yes → works bare |

## Change

- Default `manage images` `--cloud` to `admin`, so the documented bare invocation works out of the box.
- Add a regression guard asserting the parsed default.

### DIFF
--- a/osism/commands/manage.py
+++ b/osism/commands/manage.py
@@ -531,7 +531,7 @@ class Images(Command):
             action="store_true",
         )
         parser.add_argument(
-            "--cloud", type=str, help="Cloud name in clouds.yaml", default="openstack"
+            "--cloud", type=str, help="Cloud name in clouds.yaml", default="admin"
         )
         parser.add_argument(
             "--filter",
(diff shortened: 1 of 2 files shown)


jklare (approved):
LGTM

PR #2473, opened 2026-07-16
Add unit tests for osism/services/listener.py

### LINKED ISSUE #2358: Unit tests for osism/services/listener.py
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul infrastructure). Part of **Tier 6** (#2199). `osism/services/listener.py` (518 LOC) is the RabbitMQ notification listener: a `BaremetalEvents` dispatcher mapping Ironic event types to NetBox Celery tasks, and a `NotificationsDump(ConsumerMixin)` consumer with passive exchange discovery (retry loop, background discovery thread, consumer restart) plus message handling (event-bridge forwarding, OSISM API delivery with retries, handler dispatch).

## Scope

Add `tests/unit/services/test_listener.py`. There is no `tests/unit/services/` package and no listener coverage yet. The package (`tests/unit/services/__init__.py`) is also created by the companion Tier 6 issue covering the other `osism/services/` modules (`event_bridge.py` / `websocket_manager.py`) — coordinate so whichever PR lands first creates the `__init__.py` and the second rebases.

Focus: the `BaremetalEvents` state machine and exchange handling. `main()` is a `while True` loop — cover only the two reachable branches with an escape hatch (see hints); the module is large, so the issue may be split further during implementation.

### Module constants — `listener.py:25`

- `EXCHANGES_CONFIG` contains exactly the 6 services (`ironic`, `nova`, `neutron`, `cinder`, `keystone`, `glance`); for each, `exchange == service`, `routing_key == f"{service}_versioned_notifications.info"`, `queue == f"osism-listener-{service}"`
- Legacy constants `EXCHANGE_NAME`/`ROUTING_KEY`/`QUEUE_NAME` (`listener.py:59`) match the `ironic` entry

### `BaremetalEvents.get_handler()` — `listener.py:97`

No patching needed (returns bound methods).

- Each of the 8 registered event types resolves to the matching bound method, e.g. `"baremetal.node.power_set.end"` → `node_power_set_end`, `"baremetal.node.provision_set.start"` → `node_provision_set_start`, `"baremetal.node.delete.end"` → `node_delete_end` (parametrize over the full `_handler` tree)
- Unknown leaf (`"baremetal.node.power_set.start"`) and unknown branch (`"baremetal.port.create.end"`) → `KeyError` caught, returns the default handler; calling it with `{"ironic_object.data": {"name": "node-1"}}` logs `"<event_type> ## node-1"` and triggers no NetBox task
- More than 4 segments (`"baremetal.node.provision_set.end.extra"`) → still resolves via keys 0–3
- Fewer than 4 segments (`"baremetal.node.power_set"`) → raises `IndexError` (only `KeyError` is caught) — assert with `pytest.raises` to document current behavior
- `get_object_data()` (`listener.py:94`) with payload missing `"ironic_object.data"` → `KeyError`

### `BaremetalEvents` handler methods — `listener.py:113`–`listener.py:169`

Patch (the tasks are Celery-wrapped; only `.delay` is called here, no broker needed):

- `osism.services.listener.netbox.set_power_state.delay`
- `osism.services.listener.netbox.set_provision_state.delay`
- `osism.services.listener.netbox.set_maintenance.delay`

Cases (parametrize where sensible; payload is always `{"ironic_object.data": {...}}`):

- `node_power_set_end` (`listener.py:113`) → `set_power_state.delay("node-1", "power on")`
- `node_power_state_corrected_success` (`listener.py:121`) → `set_power_state.delay("node-1", "power off")`
- `node_maintenance_set_end` (`listener.py:129`) → `set_maintenance.delay("node-1", state=True)` — note the keyword argument
- `node_provision_set_start` (`listener.py:146`), `node_provision_set_end` (`listener.py:154`), `node_provision_set_success` (`listener.py:137`) → each calls `set_provision_state.delay("node-1", "<provision_state>")`
- `node_delete_end` (`listener.py:162`) → `set_provision_state.delay("node-1", None)` **and** `set_power_state.delay("node-1", None)`
- `node_create_end` (`listener.py:169`) → `set_provision_state.delay(...)` and `set_power_state.delay(...)` with the payload values

### `NotificationsDump.__init__()` — `listener.py:178`

Patch `osism.services.listener.settings.OSISM_API_URL` and `osism.services.listener.requests.Session`; pass a `MagicMock()` connection.

- `OSISM_API_URL = None` → `osism_api_session is None`, `osism_baremetal_api_url is None`
- `OSISM_API_URL = "http://api:8000/"` (trailing slash) → session created, URL `"http://api:8000/notifications/baremetal"` (`rstrip("/")` applied)
- Event bridge import succeeds → `self.event_bridge` is `osism.services.event_bridge.event_bridge` (the module singleton)
- ImportError branch → `mocker.patch.dict(sys.modules, {"osism.services.event_bridge": None})` makes the `from ... import` raise → `event_bridge is None`, warning "Event bridge not available" logged
- Initial state: `_available_exchanges == {}`, `_discovery_thread is None`, `_stop_discovery`/`_new_exchanges_found` events not set

### `_get_exchange_properties()` — `listener.py:208`

- Channel mock: `exchange_declare` succeeds → returns `{"type": "topic", "durable": True}`; assert call used `passive=True` and `type="topic"`
- `exchange_declare` raises (any `Exception`) → returns `None`, debug logged

### `_check_for_new_exchanges()` — `listener.py:232`

Mock `self.connection.channel()` as a context manager (`__enter__` returns a channel mock); stub `_get_exchange_properties` per exchange.

- No exchange exists (props always `None`) → returns `False`, `_available_exchanges` stays empty
- One exchange exists (e.g. `ironic`) → entry added with merged config **plus** `exchange_props` key, returns `True`
- Service already in `_available_exchanges` → skipped (`_get_exchange_properties` not called for it)
- `connection.channel()` raises → warning "Error checking for new exchanges" logged, returns `False`

### `_exchange_discovery_loop()` — `listener.py:259`

Call the method directly (no real thread); replace `self._stop_discovery` with a `MagicMock` event to script iterations — never rely on real timeouts.

- `wait()` returns `True` (stop requested) → breaks immediately, no exchange check
- All exchanges already in `_available_exchanges` (fill with `EXCHANGES_CONFIG`) → breaks with "Stopping exchange discovery" log, `_check_for_new_exchanges` not called
- `_check_for_new_exchanges` returns `True` → `_new_exchanges_found` set and `self.should_stop is True`; loop exits on next scripted `is_set()`/`wait()` step
- `_check_for_new_exchanges` returns `False` → loop continues without signaling (script `wait` side effect `[False, True]`)

### `_start_exchange_discovery()` — `listener.py:290` / `_stop_exchange_discovery()` — `listener.py:305`

Patch `osism.services.listener.threading.Thread`.

- All exchanges already available → no thread created
- Otherwise → `_stop_discovery.clear()` called, `Thread(target=self._exchange_discovery_loop, name="exchange-discovery", daemon=True)` constructed and `.start()` called
- Stop: sets `_stop_discovery`; thread alive → `join(timeout=5)`; `_discovery_thread is None` → no error; thread not alive → `join` not called

### `_wait_for_exchanges()` — `listener.py:311`

Patch `osism.services.listener.time.sleep` and stub `_check_for_new_exchanges`.

- `_available_exchanges` already populated → returns immediately, no check, no sleep
- First check finds nothing, second populates `_available_exchanges` (use a `side_effect` function mutating the dict) → `sleep` called once with `EXCHANGE_RETRY_INTERVAL` (60)
- Safety: give the `sleep` mock a bounded `side_effect` (raise after N calls) so a regression cannot hang pytest

### `get_consumers()` — `listener.py:331`

Stub `_wait_for_exchanges` and `_start_exchange_discovery` on the instance; pre-populate `_available_exchanges`; patch `osism.services.listener.Exchange` and `osism.services.listener.Queue`; pass a `MagicMock` consumer factory.

- Two available exchanges → two consumers returned; `Exchange(...)` called with `type`/`durable` from `exchange_props` and `passive=True`; `Queue(...)` called with `routing_key`, `auto_delete=False`, `no_ack=True`; `consumer(queue, callbacks=[self.on_message])`
- `exchange_props` missing `type`/`durable` → defaults `"topic"`/`True` used
- `Exchange` constructor raises for one service → error logged, remaining consumers still returned
- `_available_exchanges` left empty after the stubbed wait → returns `[]`, error "No consumers could be configured" logged

### `on_message()` — `listener.py:377`

Build `body = {"oslo.message": json.dumps(data)}` with `data` containing `event_type`, `payload`, `priority`, `timestamp`, `publisher_id`, `message_id`. Patch `osism.services.listener.time.sleep` for retry tests; set `self.event_bridge` / `self.osism_api_session` directly on the instance.

**Payload-info extraction / logging**

- `baremetal` event with `ironic_object.data` → only `name`/`provision_state`/`power_state` keys logged
- `compute`/`nova` event with `nova_object.data` → only `uuid`/`host`/`state`/`task_state` keys
- `network`/`neutron` event → `{"service": "neutron"}`
- Other service (`identity`) → `{"service": "identity"}`; missing `event_type` → service `"unknown"` (note: dispatch via `data["event_type"]` then raises `KeyError` if no API session — keep `payload` absent only where the code path allows)

**Event bridge forwarding**

- `event_bridge` set → `add_event(data["event_type"], data["payload"])` called once
- `add_event` raises → both error logs emitted (second one digs `payload["ironic_object.data"]["name"]` with `.get` fallbacks → `"unknown"` for non-ironic payloads), processing continues
- `event_bridge = None` → no forwarding attempted

**OSISM API delivery (`osism_api_session` set)**

- `post` returns status 204 → success log, exactly one call, no handler dispatch
- `requests.ConnectionError` on all 3 tries → 3 posts, give-up log includes `json.dumps(data)`, `sleep` called with 3 then 9 (`pow(3, tries - 1)`)
- `requests.Timeout` → same retry/give-up path
- `HTTPError` with `response.status_code = 404` → gives up after first try ("client side error")
- `status_code = 500` → also gives up early (condition is `<= 500`) — assert to document current behavior
- `status_code = 503` → retried up to 3 times
- Status 200 (not 204, `raise_for_status()` no-op) → falls through to the failure branch and retries — document current behavior
- Assert the posted JSON body contains exactly `priority`, `event_type`, `timestamp`, `publisher_id`, `message_id`, `payload`

**Handler dispatch (no API session)**

- `osism_api_session = None` → `baremetal_events.get_handler(data["event_type"])` resolved and called with `data["payload"]` (stub `get_handler` or patch the three `netbox.*.delay` targets and assert the task call)

### `main()` — `listener.py:486`

Patch `osism.services.listener.Connection`, `osism.services.listener.NotificationsDump`, `osism.services.listener.time.sleep`. Escape the `while True` loop by raising a sentinel exception (e.g. from `sleep` or the second `Connection` call) and asserting it with `pytest.raises`.

- `Connection` raises `ConnectionRefusedError` → error logged, `sleep(60)` called, loop retries (sentinel on second iteration)
- Restart path: consumer mock with `_new_exchanges_found.is_set()` → `True` once → `_stop_exchange_discovery()` called, `_new_exchanges_found.clear()` called, `_available_exchanges` carried over into the next `NotificationsDump` instance (sentinel ends the second iteration)

## Mocking hints

- `BROKER_URI` is read from the environment at import time (`listener.py:62`) — patch `osism.services.listener.BROKER_URI` in `main()` tests, not the env var.
- Importing the module pulls in `osism.tasks.netbox` and `osism.settings`; the shared `tests/conftest.py` ansible stubs already make this import safe in the unit-test venv.
- `NotificationsDump` always needs a connection argument — a plain `MagicMock()` suffices; for `_check_for_new_exchanges` configure `connection.channel.return_value.__enter__.return_value` as the channel.
- Never start the real discovery thread or hit real timeouts: test `_exchange_discovery_loop` synchronously with a scripted `_stop_discovery` mock, and patch `threading.Thread` in `_start_exchange_discovery`.
- `should_stop` is a `ConsumerMixin` property-backed attribute; asserting `consumer.should_stop is True` after the discovery loop works on the instance.
- Use the shared `loguru_logs` fixture from `tests/conftest.py` to assert log messages (pytest `caplog` does not capture loguru).
- Celery tasks: only `.delay` is invoked here — patch the three `osism.services.listener.netbox.*.delay` attributes; no broker or task execution involved.
- For the `sys.modules` ImportError trick, mapping a module name to `None` makes `from osism.services.event_bridge import event_bridge` raise `ImportError`; `mocker.patch.dict(sys.modules, ...)` restores it automatically.

## Definition of Done

- [ ] `tests/unit/services/__init__.py` present (created here or by the companion Tier 6 services issue — coordinate to avoid conflicts)
- [ ] `tests/unit/services/test_listener.py` created
- [ ] `pytest --cov=osism.services.listener` shows ≥ 90 % (the `if __name__ == "__main__"` guard and `consumer.run()` plumbing may stay uncovered)
- [ ] `pipenv run pytest tests/unit/services/test_listener.py` passes locally

## Dependencies

- Coordinates with the companion Tier 6 issue for `osism/services/` (shared `tests/unit/services/__init__.py`).

berendt:
Closes #2358.

Creates `tests/unit/services/test_listener.py` with unit tests for the RabbitMQ notification listener (`osism/services/listener.py`), which previously had no coverage.

## Coverage

- **Module constants**: `EXCHANGES_CONFIG` contains exactly the six OpenStack services with the expected exchange/routing-key/queue naming; the legacy `EXCHANGE_NAME`/`ROUTING_KEY`/`QUEUE_NAME` constants match the ironic entry
- **`BaremetalEvents.get_handler()`**: all eight registered event types resolve to their bound methods (parametrized over the full `_handler` tree), unknown leaf/branch events fall back to the default handler (logs, no NetBox task), extra event-type segments are ignored, short event types leak an `IndexError` (documented with `pytest.raises`), `get_object_data()` raises `KeyError` on missing `ironic_object.data`
- **Handler methods**: each handler calls the expected `netbox.set_power_state`/`set_provision_state`/`set_maintenance` task via `.delay` (incl. the `state=` keyword for maintenance and the double reset in `node_delete_end`)
- **`NotificationsDump.__init__()`**: no API session without `OSISM_API_URL`, trailing-slash stripping when set, event bridge singleton wiring and the `ImportError` fallback via `sys.modules`, initial discovery state
- **Exchange discovery**: `_get_exchange_properties()` (passive declare, exception → `None`), `_check_for_new_exchanges()` (new/known/none/channel-error paths), `_exchange_discovery_loop()` driven synchronously with a scripted stop event (stop request, all-available break, restart signaling via `_new_exchanges_found`/`should_stop`, quiet continuation), thread start/stop with `threading.Thread` patched, `_wait_for_exchanges()` retry with `EXCHANGE_RETRY_INTERVAL`
- **`get_consumers()`**: consumer per available exchange with `passive=True` exchanges and `no_ack=True` queues, `exchange_props` defaults, per-service error isolation, empty-exchange error path
- **`on_message()`**: payload-info extraction for baremetal/nova/neutron/other services, missing `event_type` behavior (documented `KeyError`), event bridge forwarding incl. both error logs and the `unknown` fallback for non-ironic payloads, OSISM API delivery (204 success with exact JSON body, retry/backoff `sleep(3)`/`sleep(9)` on `ConnectionError`/`Timeout`, early give-up on 4xx client errors, a strict `xfail` asserting that a 500 is retryable — the current `<= 500` guard drops it as a client error, so the test turns red the moment the boundary is fixed to `< 500` — retries on 503 and on the 200-falls-through path), handler dispatch without an API session
- **`main()`**: `ConnectionRefusedError` retry with `sleep(60)`, and the consumer restart path when the discovery thread finds new exchanges (exchange carry-over into the next `NotificationsDump` instance)

## Notes

- Threads and timeouts are never real: the discovery loop is called synchronously with a scripted `_stop_discovery` mock, `threading.Thread` and `time.sleep` are patched (the sleep mock carries a bounded guard so a regression cannot hang pytest), and `main()`'s `while True` loop is escaped with a sentinel exception.
- The NetBox Celery tasks are only invoked via `.delay`, so the three `osism.services.listener.netbox.*.delay` attributes are patched; no broker is needed.
- Tests that pin behavior of currently dormant code paths (API delivery without `OSISM_API_URL` set, non-ironic events, malformed event types) carry comments noting that the path is dormant and the pinned behavior is not the intended end-state.
- `tests/unit/services/__init__.py` is byte-identical to the one created by #2460, so whichever PR lands second merges cleanly.

### DIFF
--- a/tests/unit/services/test_listener.py
+++ b/tests/unit/services/test_listener.py
@@ -1,0 +1,1057 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for the RabbitMQ notification listener (``osism/services/listener.py``).
+
+Covers the ``EXCHANGES_CONFIG`` module constants, the ``BaremetalEvents``
+dispatcher (handler resolution and the NetBox task calls of every handler),
+the ``NotificationsDump`` consumer (initialization, passive exchange
+discovery, consumer setup and message handling including event-bridge
+forwarding and OSISM API delivery retries) and the ``main()`` retry loop.
+
+The NetBox Celery tasks are only ever invoked via ``.delay`` here, so the
+three ``osism.services.listener.netbox.*.delay`` attributes are patched and
+no broker is needed. Threads and timeouts are never real: the discovery loop
+is driven synchronously with a scripted ``_stop_discovery`` mock,
+``threading.Thread`` and ``time.sleep`` are patched, and ``main()``'s
+``while True`` loop is escaped with a sentinel exception.
+"""
+
+import json
+import sys
+from unittest.mock import MagicMock, call
+
+import pytest
+import requests
+
+from osism.services import listener
+
+# ---------------------------------------------------------------------------
+# Helpers / fixtures
+# ---------------------------------------------------------------------------
+
+SERVICES = ["ironic", "nova", "neutron", "cinder", "keystone", "glance"]
+
+# Full ``BaremetalEvents._handler`` tree: event type -> handler method name.
+HANDLER_EVENT_TYPES = [
+    ("baremetal.node.power_set.end", "node_power_set_end"),
+    (
+        "baremetal.node.power_state_corrected.success",
+        "node_power_state_corrected_success",
+    ),
+    ("baremetal.node.maintenance_set.end", "node_maintenance_set_end"),
+    ("baremetal.node.provision_set.start", "node_provision_set_start"),
+    ("baremetal.node.provision_set.end", "node_provision_set_end"),
+    ("baremetal.node.provision_set.success", "node_provision_set_success"),
+    ("baremetal.node.delete.end", "node_delete_end"),
+    ("baremetal.node.create.end", "node_create_end"),
+]
+
+API_URL = "http://api:8000/notifications/baremetal"
+
+
+class LoopExit(Exception):
+    """Sentinel used to escape ``while True`` loops in ``main()`` tests."""
+
+
+def _has_log(records, level, substring):
+    return any(r["level"] == level and substring in r["message"] for r in records)
+
+
+def _make_data(event_type="baremetal.node.power_set.end", payload=None):
+    if payload is None:
+        payload = {"ironic_object.data": {"name": "node-1", "power_state": "power on"}}
+    return {
+        "event_type": event_type,
+        "payload": payload,
+        "priority": "INFO",
+        "timestamp": "2026-01-01 00:00:00.000000",
+        "publisher_id": "baremetal.host.example.com",
+        "message_id": "00000000-0000-0000-0000-000000000000",
+    }
+
+
+def _make_body(data):
+    return {"oslo.message": json.dumps(data)}
+
+
+@pytest.fixture
+def netbox_delays(mocker):
+    """Patch the ``.delay`` attribute of the three NetBox tasks used by handlers."""
+    from types import SimpleNamespace
+
+    return SimpleNamespace(
+        set_power_state=mocker.patch(
+            "osism.services.listener.netbox.set_power_state.delay"
+        ),
+        set_provision_state=mocker.patch(
+            "osism.services.listener.netbox.set_provision_state.delay"
+        ),
+        set_maintenance=mocker.patch(
+            "osism.services.listener.netbox.set_maintenance.delay"
+        ),
+    )
+
+
+@pytest.fixture
+def baremetal_events():
+    return listener.BaremetalEvents()
+
+
+@pytest.fixture
+def consumer(mocker):
+    """``NotificationsDump`` with the OSISM API disabled and a mock connection."""
+    mocker.patch("osism.services.listener.settings.OSISM_API_URL", None)
+    return listener.NotificationsDump(MagicMock())
+
+
+@pytest.fixture
+def api_consumer(consumer):
+    """Consumer wired for OSISM API delivery with stubbed session and events."""
+    consumer.event_bridge = None
+    consumer.baremetal_events = MagicMock()
+    consumer.osism_api_session = MagicMock()
+    consumer.osism_baremetal_api_url = API_URL
+    return consumer
+
+
+@pytest.fixture
+def sleep_mock(mocker):
+    """Patch ``time.sleep`` with a bounded guard so regressions cannot hang."""
+    sleep = mocker.patch("osism.services.listener.time.sleep")
+    sleep.side_effect = [None] * 5 + [RuntimeError("time.sleep called too often")]
+    return sleep
+
+
+# ---------------------------------------------------------------------------
+# Module constants
+# ---------------------------------------------------------------------------
+
+
+def test_exchanges_config_contains_exactly_the_six_services():
+    assert sorted(listener.EXCHANGES_CONFIG) == sorted(SERVICES)
+
+
+@pytest.mark.parametrize("service", SERVICES)
+def test_exchanges_config_entry(service):
+    assert listener.EXCHANGES_CONFIG[service] == {
+        "exchange": service,
+        "routing_key": f"{service}_versioned_notifications.info",
+        "queue": f"osism-listener-{service}",
+    }
+
+
+def test_legacy_constants_match_ironic_entry():
+    ironic = listener.EXCHANGES_CONFIG["ironic"]
+    assert listener.EXCHANGE_NAME == ironic["exchange"]
+    assert listener.ROUTING_KEY == ironic["routing_key"]
+    assert listener.QUEUE_NAME == ironic["queue"]
+
+
+# ---------------------------------------------------------------------------
+# BaremetalEvents.get_handler() / get_object_data()
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("event_type,method_name", HANDLER_EVENT_TYPES)
+def test_get_handler_resolves_registered_event_types(
+    baremetal_events, event_type, method_name
+):
+    handler = baremetal_events.get_handler(event_type)
+    assert handler == getattr(baremetal_events, method_name)
+
+
+@pytest.mark.parametrize(
+    "event_type",
+    [
+        pytest.param("baremetal.node.power_set.start", id="unknown-leaf"),
+        pytest.param("baremetal.port.create.end", id="unknown-branch"),
+    ],
+)
+def test_get_handler_unknown_event_returns_default_handler(
+    baremetal_events, netbox_delays, loguru_logs, event_type
+):
+    handler = baremetal_events.get_handler(event_type)
+    assert handler.__name__ == "default_handler"
+
+    handler({"ironic_object.data": {"name": "node-1"}})
+
+    assert _has_log(loguru_logs, "INFO", f"{event_type} ## node-1")
+    netbox_delays.set_power_state.assert_not_called()
+    netbox_delays.set_provision_state.assert_not_called()
+    netbox_delays.set_maintenance.assert_not_called()
+
+
+def test_get_handler_ignores_extra_segments(baremetal_events):
+    handler = baremetal_events.get_handler("baremetal.node.provision_set.end.extra")
+    assert handler == baremetal_events.node_provision_set_end
+
+
+def test_get_handler_too_few_segments_raises_index_error(baremetal_events):
+    # Only KeyError is caught; short event types leak an IndexError. This
+    # path is currently dormant (ironic event types always carry four
+    # segments) and the leak is not the intended end-state.
+    with pytest.raises(IndexError):
+        baremetal_events.get_handler("baremetal.node.power_set")
+
+
+def test_get_object_data_missing_key_raises_key_error(baremetal_events):
+    with pytest.raises(KeyError):
+        baremetal_events.get_object_data({"payload": {}})
+
+
+# ---------------------------------------------------------------------------
+# BaremetalEvents handler methods
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize(
+    "method_name,power_state",
+    [
+        ("node_power_set_end", "power on"),
+        ("node_power_state_corrected_success", "power off"),
+    ],
+)
+def test_power_handlers_set_power_state(
+    baremetal_events, netbox_delays, method_name, power_state
+):
+    payload = {"ironic_object.data": {"name": "node-1", "power_state": power_state}}
+    getattr(baremetal_events, method_name)(payload)
+    netbox_delays.set_power_state.assert_called_once_with("node-1", power_state)
+    netbox_delays.set_provision_state.assert_not_called()
+    netbox_delays.set_maintenance.assert_not_called()
+
+
+def test_node_maintenance_set_end_sets_maintenance_keyword(
+    baremetal_events, netbox_delays
+):
+    payload = {"ironic_object.data": {"name": "node-1", "maintenance": True}}
+    baremetal_events.node_maintenance_set_end(payload)
+    netbox_delays.set_maintenance.assert_called_once_with("node-1", state=True)
+
+
+@pytest.mark.parametrize(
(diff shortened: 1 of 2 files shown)


tests/unit/services/test_listener.py
```diff
@@ -0,0 +1,1038 @@
...
+    assert sleep_mock.call_args_list == [call(3), call(9)]
+    give_up = next(
+        r
+        for r in loguru_logs
+        if r["level"] == "ERROR" and "Giving up delivering notification" in r["message"]
+    )
+    assert json.dumps(data) in give_up["message"]
+
+
+@pytest.mark.parametrize("status_code", [404, 500])
```
ideaship: Bundling `500` in with `404` here pins an off-by-one as the expected result. The early-exit guard is `status_code <= 500` (`osism/services/listener.py`), so a server-side 500 is treated as a client error and the notification is dropped without retry — inconsistent with `test_on_message_api_delivery_retries_on_server_error` just below, where a 503 retries three times. A 500 is retryable.

Rather than asserting this green, suggest splitting the server case out and marking it `xfail(strict=True)` so it turns red the moment the guard is corrected (`<= 500` -> `< 500`):

```python
@pytest.mark.parametrize("status_code", [404])
def test_on_message_api_delivery_gives_up_early_on_http_error(
    api_consumer, sleep_mock, loguru_logs, status_code
):
    # A 4xx client error is not retried.
    response = MagicMock(status_code=status_code)
    response.raise_for_status.side_effect = requests.HTTPError(response=response)
    api_consumer.osism_api_session.post.return_value = response

    api_consumer.on_message(_make_body(_make_data()), MagicMock())

    assert api_consumer.osism_api_session.post.call_count == 1
    sleep_mock.assert_not_called()
    assert _has_log(loguru_logs, "ERROR", "client side error, giving up early")


@pytest.mark.xfail(
    strict=True,
    reason="listener guards with status_code <= 500; a server 500 is retryable "
    "and the boundary should be < 500",
)
def test_on_message_api_delivery_retries_on_500(api_consumer, sleep_mock):
    # A 500 is a server error and should be retried like the 503 case.
    response = MagicMock(status_code=500)
    response.raise_for_status.side_effect = requests.HTTPError(response=response)
    api_consumer.osism_api_session.post.return_value = response

    api_consumer.on_message(_make_body(_make_data()), MagicMock())

    assert api_consumer.osism_api_session.post.call_count == 3
```

The new test asserts the intended retry behavior, so it fails today and passes once the guard is fixed — at which point `strict=True` flags it for removal.

ideaship (commented):
Solid, mechanically clean test suite — good coverage of the discovery loop, the handler dispatch, and the delivery retries.

One structural note. Several tests characterize behavior on code paths that a standard deployment doesn't exercise, and a couple of those paths are themselves buggy — so the tests currently freeze the bug as the expected result. Two dormant paths are involved:

- **API delivery** (`on_message`'s HTTP branch) only runs when `OSISM_API_URL` is set. In the default configuration it isn't, and delivery happens in-process instead.
- **Non-ironic events** only reach the listener if a service is configured to emit versioned notifications on a `<service>_versioned_notifications` topic. Only ironic does that by default, so nova/neutron/etc. events don't currently arrive at all.

Because of that, none of this blocks merging — it can't affect a running default deployment. The ask is narrower: don't pin *buggy* behavior as the spec.

The one case worth changing here is `test_on_message_api_delivery_gives_up_early_on_http_error`, which parametrizes `500` alongside `404`. The guard is `status_code <= 500`, so a server 500 is classified as a client error and dropped without retry — inconsistent with the 503-retry test just below. A 500 is retryable. Suggest splitting the server case out and marking it `xfail(strict=True)` (see inline comment), so it turns red the moment the one-character fix lands rather than asserting the current behavior green.

The remaining characterization tests — short/invalid event types, missing `event_type`, non-baremetal events posting to the baremetal endpoint, a 200 not counting as success, and single-channel exchange discovery — are reasonable to keep. Consider rewording their "documents the current behavior" notes to say the path is currently dormant, so a future reader doesn't read them as intended end-state.

PR #2480, opened 2026-07-17
Add unit tests for workflow commands

### LINKED ISSUE #2363: Unit tests for osism/commands/ — workflow (apply, check, validate, wait, compose, sync, get, log, console)
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul infrastructure). Part of **Tier 8** (#2199). This issue covers the workflow-oriented CLI command modules under `osism/commands/`: `apply.py` (501 LOC), `check.py` (778 LOC), `validate.py` (217 LOC), `wait.py` (159 LOC), `compose.py` (45 LOC), `sync.py` (406 LOC), `get.py` (302 LOC), `log.py` (237 LOC) and `console.py` (296 LOC) — together ~2,940 LOC. They are cliff `Command` classes that schedule Celery tasks (`apply`, `validate`, `sync`), inspect Celery state (`wait`, `get tasks`), shell out via SSH/clush/docker (`compose`, `log`, `console`) or compare filesystem metadata (`check`).

## Scope

Add `tests/unit/commands/test_apply.py`, `test_check.py`, `test_compose.py`, `test_sync.py`, `test_log.py` and `test_console.py`; extend the existing `test_validate.py`, `test_wait.py` and `test_get.py`.

Already covered (do not duplicate):

- `test_validate.py`: `Run._handle_task` returns 1 on `TimeoutError` while waiting.
- `test_wait.py`: exit-code contract of the `--live` path (timeout with one/two tasks, task-rc passthrough, success → 0).
- `test_get.py`: `Hosts.take_action` (inventory load failure → 1, empty inventory → success) and `Hostvars.take_action` (inventory query failure → 1, missing variable → success).

This is a large group: prioritize the pure-logic / high-value functions listed below; interactive loops (`log.Opensearch`, `console` `container_prompt`) are lowest priority. Like the Tier 3 issues, this issue may be split further during implementation (suggested cut: `apply`+`check` / `sync`+`get` / `compose`+`log`+`console` + the `validate`/`wait` extensions).

### `Run._prepare_task()` — `apply.py:269`

Patch `osism.tasks.ansible.run`, `osism.tasks.ceph.run`, `osism.tasks.kolla.run`, `osism.tasks.kubernetes.run` (the `.si` attribute is what gets called; imports happen inside the method body, so patch the canonical task module paths). `MAP_ROLE2ENVIRONMENT` / `MAP_ROLE2RUNTIME` are lazy module attributes of `osism.data.playbooks` — see mocking hints.

- `role="ceph"` → environment forced to `"ceph"`, `ceph.run.si("ceph", "ceph", arguments, auto_release_time=task_timeout)`
- `role="ceph-osds"`, environment `"ceph"` → `ceph-` prefix stripped: `ceph.run.si("ceph", "osds", ...)`
- `sub="zone-a"` with ceph/kubernetes/kolla environments → environment becomes `"ceph.zone-a"` etc.
- environment `"kubernetes"` → `kubernetes.run.si(...)`
- `role="loadbalancer-ng"` → returns chain of `kolla.run.si(env, "loadbalancer-ng", ...)` piped into a `group` over `enums.LOADBALANCER_PLAYBOOKS`
- environment `"kolla"`, `role="kolla-keystone"` → `kolla-` prefix stripped; `kolla.run.si` called with `["-e kolla_action=deploy"] + arguments`
- `role="mariadb-ng"` / `"rabbitmq-ng"` → argument is `-e kolla_action_ng=<action>` instead
- kolla role listed in `MAP_ROLE2RUNTIME["osism-ansible"]` (and not `"common"`) → routed to `ansible.run.si` with the original arguments (no `kolla_action`)
- environment `None` and role not in `MAP_ROLE2ENVIRONMENT` → bare-except fallback to `"custom"`, info log "Trying to run play ...", `ansible.run.si("custom", ...)`
- `overwrite="other"` in the default branch → environment replaced by `"other"`

### `Run._handle_collection()` / `Run.handle_collection()` — `apply.py:135` / `apply.py:218`

Patch the task modules as above plus `osism.tasks.ansible.noop`; patch `celery.chain` / `celery.group` or assert on the returned signature structure. Drive via small hand-built `Role` trees (`osism.data.enums.Role`) instead of the real `MAP_ROLE2ROLE`.

- item that is not a `Role` → `TypeError` raised and error logged
- flat list of `Role`s without dependencies → one prepared task per role, wrapped in a `group`
- `Role` with nested dependencies → `chain(parent_task, child_group)` built recursively
- `dry_run=True` → `ansible.noop.si()` used instead of `_prepare_task`
- `show_tree=True` → no tasks created, returns `None`, tree logged (`"A [0] ..."` / indentation grows with counter)
- `handle_collection`: `apply_async()` called on the result when not `show_tree`; not called for `show_tree`; distinct log messages for dry-run / show-tree / normal mode (patch `osism.data.enums.MAP_ROLE2ROLE` with `dict`-patch for the collection lookup)

### `Run.handle_role()` / `Run.handle_loadbalancer_task()` — `apply.py:371` / `apply.py:106`

Patch `osism.tasks.handle_task` (imported inside the method) and stub `_prepare_task` to return a `MagicMock` whose `apply_async()` yields either a plain result or a `celery.result.GroupResult` instance.

- plain task → `handle_task(task, wait, format, timeout)` rc passed through
- `GroupResult` → `handle_loadbalancer_task` path taken
- `handle_loadbalancer_task` with `wait=True`: rc comes from `handle_task(t.parent, ...)`, `t.get()` called
- `wait=False`: `t.parent.get()` additionally called (garbage-collector workaround), children logged for `format="log"`

### `Run.take_action()` — `apply.py:417`

Patch `osism.commands.apply.utils.check_task_lock_and_exit`, `osism.commands.apply.utils.check_ansible_facts`, and the `handle_role` / `handle_collection` methods on the command instance.

- `check_task_lock_and_exit` always invoked first
- no role → table of `MAP_ROLE2ENVIRONMENT` printed (capsys), returns 0
- ansible-facts freshness check: performed when a role is given and >300 s since `utils._last_ansible_facts_check`; skipped for roles `"gather-facts"`/`"facts"`, for `--show-tree`, and when the last check is recent (reset the `_last_ansible_facts_check` attribute on `osism.utils` between tests)
- `role="a//b"` → `handle_role` called once per segment
- `--retry 2` with `handle_role` always returning 1 → called 3 times, rc 1; success on 2nd attempt → 2 calls, rc 0
- collection branch: `handle_collection` returns `None`, so `rc != 0` is truthy and the loop breaks with `take_action` returning `None` — pin this current behavior in a test (candidate for a follow-up fix)

### `get_file_info()` / `collect_file_info()` — `check.py:31` / `check.py:59`

Pure filesystem helpers — use `tmp_path`, no mocking needed for the happy paths.

- small regular file → dict with `inode`, `mtime`, `size`, `mode`, `uid`, `gid`, `is_link=False` and an md5 `hash`
- file ≥ 1 MiB → `hash is None`
- unreadable file (patch `builtins.open` to raise `IOError`) → `hash is None`, rest populated
- nonexistent path → `{"error": ...}`
- `collect_file_info`: directory tree with `.git`/`venv`/`__pycache__` subdirs → skipped; symlinks skipped; both files and directories included with relative paths
- `max_files=2` on a larger tree → scan stops, warning logged

### `parse_stat_output()` — `check.py:88`

Pure parser, no mocks.

- well-formed `FILE:`/`INODE:`/`SIZE:`/`MTIME:`/`HASH:` blocks → typed dict (`int`, `int`, `float`)
- `HASH:NONE` → `None`
- `ERROR:` lines captured under `"error"`
- empty lines and key/value lines before any `FILE:` line ignored
- multiple files in one output parsed independently

### `Mount._compare_file_info()` — `check.py:338`

Pure comparison, no mocks.

- differing inodes → entry in `inode_mismatches` with `local_inode`/`fresh_inode`
- differing hashes with `check_content=True` → `content_mismatches`; same input with `check_content=False` → empty
- file only in fresh info → `missing_in_local`; only in local info → `missing_in_fresh`
- entries containing `"error"` skipped entirely
- falsy inode (`None`/`0`) on either side → no mismatch recorded
- results sorted by file path

### `Mount._get_container_id()` / `Mount._get_mount_source()` — `check.py:180` / `check.py:222`

Patch `builtins.open` with `mock_open` payloads and `os.uname`.

- cgroup line containing `docker` with a 64-char id → first 12 chars returned; 12-char id returned as-is
- cgroup unreadable + 12-char hostname → hostname returned
- cgroup + hostname fail, `/proc/self/mountinfo` containing `/docker/containers/<id>/` → id truncated to 12
- nothing matches → `None`
- `_get_mount_source`: mountinfo line whose 5th field equals the mount path → source after the `-` separator returned (must start with `/`); no separator / non-absolute source / no matching line / `IOError` → `None`

### `Mount.take_action()` — `check.py:391`

Focus on the early-exit guards and final rc; patch `osism.commands.check.DOCKER_AVAILABLE`, `osism.commands.check.docker`, `os.path.exists`, and the instance helpers (`_get_container_id`, `_get_volume_mount_info`, `_get_mount_source`, `_run_fresh_container`) plus `collect_file_info`.

- path does not exist → 1 (`format="script"` prints `FAILED: ...`)
- `DOCKER_AVAILABLE=False` → 1
- Docker socket missing → 1
- `docker.from_env` raises → 1
- mount source not determinable (no `--host-path`, no Docker mount info, no mountinfo) → 1
- bind mount → source taken from mount info; volume mount with `--volume-name` → override used
- `_run_fresh_container` raises → 1
- consistent comparison → 0 (`script` prints `PASSED`); inode mismatches → 1 with `INODE_MISMATCHES:<n>` in script format

### `Inode.take_action()` — `check.py:662`

- explicit file list under `tmp_path` → rows with type/inode/size; symlinks and missing files skipped
- no files given → random sampling from `environments/*` and `inventory/*` (patch `random.sample` for determinism); up to 2 entries per subdirectory plus direct files
- all three formats (`table` via capsys, `log`, `script`); returns 0

### `validate.Run.take_action()` — `validate.py:74` (gap)

Patch `osism.tasks.ansible.run`, `osism.tasks.ceph.run`, `osism.tasks.kolla.run` (`.delay`), `osism.commands.validate.utils.check_task_lock_and_exit`, and stub `_handle_task`.

- kolla validator (e.g. `keystone-config`) → `kolla.run.delay("kolla", "keystone", arguments)` with `-e kolla_action=config_validate` appended to arguments
- `--environment custom` honored for ceph/kolla runtimes (no default applied)
- ceph validator (`ceph-config`) → `ceph.run.delay("ceph", "validate", ...)` (playbook rewritten via `VALIDATE_PLAYBOOKS`)
- osism-ansible validator without explicit `playbook` key (e.g. `ntp`) → `ansible.run.delay("generic", "validate-ntp", ...)`
- `check_task_lock_and_exit` invoked

### `validate.Run._handle_task()` — `validate.py:55` (gap; timeout case exists)

- `wait=True`, `fetch_task_output` returns rc → rc passed through
- `wait=False`, `format="log"` → info log, returns 0
- `wait=False`, `format="script"` → task id printed, returns 0

### `validate.Scs.take_action()` — `validate.py:159`

Patch `osism.tasks.openstack.setup_cloud_environment`, `osism.tasks.openstack.cleanup_cloud_environment` (imported inside the method) and `osism.commands.validate.subprocess.run`.

- `setup_cloud_environment` returns `success=False` → returns 1, no subprocess call, no cleanup
- happy path → command contains `-s <cloud>`, `-a os_cloud=<cloud>`, `-V <version>`, ends with `scs-compatible-iaas.yaml`; `OS_CLIENT_CONFIG_FILE=/tmp/clouds.yaml` in env; returncode passed through
- `--verbose`/`--debug`/`--tests`/`--output`/`--sections` each append the corresponding flag
- `subprocess.run` raises `FileNotFoundError` → 1; generic exception → 1
- `cleanup_cloud_environment(temp_files, original_cwd)` called in all post-setup paths (finally)

### `wait.Run.get_all_task_ids()` / `take_action()` — `wait.py:50` / `wait.py:62` (gap; `--live` exit codes exist)

Patch `celery.Celery`, `celery.result.AsyncResult`, `osism.commands.wait.time.sleep`, and `osism.utils._init_redis`. Make `AsyncResult` return objects whose `state` changes between iterations (via `side_effect`) so re-queue loops terminate.

- `get_all_task_ids`: merges ids from `i.scheduled()` and `i.active()`, returns them sorted
- no task ids on CLI → ids pulled from `get_all_task_ids`, refresh mode enabled
- `PENDING` + `query_task` finds nothing → "unavailable" logged / `<id> = UNAVAILABLE` printed, task not re-queued
- `PENDING` + task known to a worker → re-queued, then `SUCCESS` on the next pass terminates
- `SUCCESS` with `--output` → `result.get()` printed
- `STARTED` without `--live` → re-queued, finishes when state flips to `SUCCESS`
- `--refresh 1` → after the queue drains, `get_all_task_ids` consulted one more time
- `format="script"` prints `<id> = <STATE>` lines instead of log output

### `compose.Run.take_action()` — `compose.py:25`

Patch `osism.commands.compose.subprocess.call` and `osism.commands.compose.ensure_known_hosts_file`.

- builds `ssh ... <OPERATOR_USER>@<host> 'docker compose --project-directory=/opt/<environment> <arguments>'` with `UserKnownHostsFile=<KNOWN_HOSTS_PATH>`; arguments are joined without separators (current behavior — pin it)
- `ensure_known_hosts_file` returns `False` → warning logged, SSH still attempted

### `sync.Facts` / `sync.CephKeys` / `sync.Sonic` — `sync.py:21` / `sync.py:46` / `sync.py:90`

Patch `osism.commands.sync.utils.check_task_lock_and_exit`, `osism.tasks.ansible.run` / `osism.tasks.conductor.sync_sonic` (`.delay`) and `osism.tasks.handle_task`.

- `Facts`: `ansible.run.delay("generic", "gather-facts", [], auto_release_time=3600)`, rc from `handle_task`
- `CephKeys`: `manager`/`copy-ceph-keys` playbook; `--no-wait` → `handle_task(t, False)`
- `Sonic`: `conductor.sync_sonic.delay(device, show_diff)`; device-specific vs. generic log message; `--no-diff` → `show_diff=False`

### `sync.Versions._get_kolla_version_from_release()` — `sync.py:248`

Patch `requests.get` (imported inside the method).

- response with `docker_images: {kolla: "0.20250928.0"}` → version returned, URL is `<repo>/<release>/base.yml`
- HTTP error (`raise_for_status` raises `RequestException`) → `RuntimeError`
- invalid YAML body → `RuntimeError`
- YAML without `docker_images.kolla` → `RuntimeError` "Kolla version not found"

### `sync.Versions._sync_kolla_versions()` / `take_action()` — `sync.py:311` / `sync.py:288`

Stub `_extract_sbom_with_skopeo` and `_get_kolla_version_from_release`; use `tmp_path` as `--configuration-path`.

- `--release 9.4.0` → image `=<sbom-image-base>:<version-from-release>`; `_get_kolla_version_from_release` raising → rc 1
- version tag with a date part (`0.20251128.0`, also `v`-prefixed) → release SBOM image base; plain OpenStack version (`2025.1`) → `registry.osism.cloud/kolla/sbom:2025.1`
- explicit `--sbom-image` → used verbatim, no derivation
- non-dry-run with missing configuration path → rc 1
- `_extract_sbom_with_skopeo` raising `RuntimeError` / `YAMLError` → rc 1
- `openstack_version` from the SBOM overrides the CLI value in the rendered template
- `--dry-run` → rendered `versions.yml` printed, nothing written, rc 0
- happy path → file written to `<config>/environments/kolla/versions.yml` (directory auto-created), rc 0

### `sync.Versions._extract_sbom_with_skopeo()` — `sync.py:169`

Patch `osism.commands.sync.subprocess.run` to a no-op and pre-build a fake OCI layout (the `tmpdir` comes from `tempfile.TemporaryDirectory` — patch `osism.commands.sync.tempfile.TemporaryDirectory` to return a `tmp_path`-backed context): `index.json` → manifest blob → tar layer containing `images.yml`.

- happy path → parsed `images.yml` dict returned
- `skopeo` exits non-zero (`CalledProcessError`) → `RuntimeError` "skopeo copy failed"
- `skopeo` binary missing (`FileNotFoundError`) → `RuntimeError` "skopeo not found"
- layer that is not a tarfile → skipped, next layer used
- no layer contains `images.yml` → `RuntimeError` "images.yml not found"

### `get.VersionsManager.take_action()` — `get.py:21`

Patch `docker.from_env` (imported inside the method).

- three containers with `org.opencontainers.image.version` labels → table rows; `ceph-ansible` adds `de.osism.release.ceph`, `kolla-ansible` adds `de.osism.release.openstack`, `osism-ansible` has empty release
- `client.containers.get` raising `docker.errors.NotFound` for one name → that row skipped, others still printed

### `get.Tasks.take_action()` — `get.py:58`

Patch `celery.Celery` so `app.control.inspect()` returns a mock with `active()` / `scheduled()` dicts.

- active and scheduled tasks rendered with worker, id, name, status `ACTIVE`/`SCHEDULED`, start time (`datetime.fromtimestamp`) and args
- empty inspect results → empty table, no exception

### `get.Facts.take_action()` / `get.States.take_action()` — `get.py:189` / `get.py:277`

Patch the lazy redis client (`mocker.patch("osism.utils._init_redis", return_value=MagicMock())` or patch `osism.commands.get.utils.redis`).

- `Facts`: no cache entry → error "No facts found in cache"; specific fact present → single row; fact missing → error logged; full listing truncates the four `ansible_ssh_host_key_*_public` facts to 40 chars + `...`
- `States`: facts with `ansible_local.osism` → one row per role with state/timestamp, `bootstrap` skipped; missing `ansible_local`/`osism` key or no cache entry → nothing printed, no exception

### `get.Hostvars` / `get.Hosts` happy paths — `get.py:127` / `get.py:238` (gap)

- `Hostvars` with variable present → grid table row printed; without variable argument → one row per variable
- `Hosts` with hosts in inventory → psql table of hostnames (capsys)

### `log.Ansible` / `log.Container` — `log.py:31` / `log.py:54`

Patch `osism.commands.log.subprocess.call` and `osism.commands.log.ensure_known_hosts_file`.

- `Ansible`: parameters joined and appended to `/usr/local/bin/ara`
- `Container`: command contains `docker logs <parameters> <container>` and `<OPERATOR_USER>@<host>`; `ensure_known_hosts_file` failure → warning, call still made

### `log.File.take_action()` — `log.py:105`

Patch `osism.commands.log.get_hosts_from_group`, `osism.commands.log.resolve_host_with_fallback`, `osism.commands.log.subprocess.call`, `osism.commands.log.ensure_known_hosts_file`.

- path traversal (`../../etc/passwd`) → error "must stay within /var/log", rc 1, no subprocess call; `kolla/nova/nova-compute.log` → `/var/log/kolla/nova/nova-compute.log` accepted
- tail command: `-n <lines>` always, `-f` only with `--follow`, path shell-quoted via `shlex.quote`
- group with multiple hosts → clush invoked with `-w host1,host2`; clush rc != 0 → rc passed through with error log
- group with exactly one host → host substituted, ssh path used
- non-group host → `resolve_host_with_fallback` result used in `user@host`; ssh rc != 0 → passthrough; success → 0

### `log.Opensearch.take_action()` — `log.py:203` (low priority)

Patch `osism.commands.log.PromptSession` (session.prompt `side_effect=["<query>", "exit"]`) and `osism.commands.log.requests.post`.

- `exit` immediately breaks the loop
- response with `hits` → `Payload` printed per hit; `--verbose` prints `timestamp | Hostname | [programname |] Payload`, falling back to `@timestamp` when `timestamp` is absent
- response without `hits` → raw JSON printed

### `console` module helpers — `console.py:18` / `console.py:37` / `console.py:65` / `console.py:97` / `console.py:128`

- `resolve_hostname_to_ip` (`console.py:18`): patch `socket.gethostbyname` — success → IP; `socket.gaierror` → `None`
- `get_primary_ipv4_from_netbox` (`console.py:37`): patch `osism.commands.console.utils.nb` — `nb` falsy → `None`; device with `primary_ip4.address = "10.0.0.1/24"` → `"10.0.0.1"`; device `None` / no `primary_ip4` → `None`; query raising → `None` with warning
- `resolve_host_with_fallback` (`console.py:65`): DNS hit → IP; DNS miss + Netbox hit → Netbox IP; both miss → original hostname returned with warning
- `get_hosts_from_group` (`console.py:97`): patch `osism.commands.console.subprocess.check_output`, `get_inventory_path`, `get_hosts_from_inventory` — valid inventory → sorted host list; any exception (e.g. `CalledProcessError`) → `[]`
- `select_host_from_list` (`console.py:128`): patch `osism.commands.console.prompt` — valid number → that host; `q`/`quit`/`exit` → `None`; non-numeric then valid input → retries; out-of-range then valid → retries

### `console.Run.take_action()` — `console.py:172`

Patch `osism.commands.console.subprocess.call`, `ensure_known_hosts_file`, `get_hosts_from_group`, `resolve_host_with_fallback`, `select_host_from_list`, `prompt`.

- host syntax routing: `"ctl001/"` → container_prompt loop; `"ctl001/rabbitmq"` → container; `".ctl001"` → `/run-ansible-console.sh ctl001`; `":ctl"` → clush with `-g ctl` and `-l <OPERATOR_USER>`
- ssh type: group resolving to one host → that host used; multiple hosts → `select_host_from_list`; selection cancelled (`None`) → returns without SSH call
- ssh call uses `resolve_host_with_fallback` result and `UserKnownHostsFile=<KNOWN_HOSTS_PATH>`
- container type: `docker exec -it <container> bash` with both parts `shlex.quote`d, `RequestTTY=force` in options, host part resolved
- container_prompt: command then `exit` → one SSH call with `docker <quoted command>`

## Mocking hints

- Instantiate cliff commands as in the existing tests: `cmd = module.Class(MagicMock(), MagicMock())`, then `parsed_args = cmd.get_parser("test").parse_args([...])` — this exercises the real argparse defaults (`tests/unit/commands/test_wait.py` shows the pattern).
- Celery task modules (`osism.tasks.ansible`, `ceph`, `kolla`, `kubernetes`, `conductor`) and `osism.tasks.handle_task` are imported *inside* method bodies — patch them at their canonical module path (e.g. `mocker.patch("osism.tasks.ansible.run")`), not on the command module. Assert on `.si(...)` / `.delay(...)` call args; no broker is needed since `apply_async`/`delay` are mocked.
- `osism.data.playbooks.MAP_ROLE2ENVIRONMENT` / `MAP_ROLE2RUNTIME` are lazy module attributes loaded from `/interface/playbooks` via module-level `__getattr__`; set them with `monkeypatch.setattr(playbooks, "MAP_ROLE2ENVIRONMENT", {...}, raising=False)` and call `osism.data.playbooks._reset_caches()` in teardown (an autouse fixture keeps this tidy).
- The lazy redis attribute on `osism.utils`: patch `osism.utils._init_redis` to return a `MagicMock` (as `test_wait.py` already does) before anything touches `utils.redis`.
- `apply.take_action` stores `utils._last_ansible_facts_check` as an attribute on the `osism.utils` module — delete/reset it between tests (`monkeypatch.delattr(osism.utils, "_last_ansible_facts_check", raising=False)`).
- `wait.take_action` loops until states converge: give the patched `AsyncResult` a `side_effect` list whose states end in `SUCCESS`, and always patch `osism.commands.wait.time.sleep`.
- `check.py` guards the docker import with try/except — patch `osism.commands.check.DOCKER_AVAILABLE` and `osism.commands.check.docker` directly. The pure helpers (`get_file_info`, `collect_file_info`, `parse_stat_output`, `_compare_file_info`) need only `tmp_path`.
- `sync.Versions`: `requests` and `jinja2` are imported inside methods — patch `requests.get`; for the template-render assertions just inspect the written file/printed output instead of mocking jinja2.
- `log.py` imports `get_hosts_from_group` and `resolve_host_with_fallback` from `osism.commands.console` at module level — patch them as `osism.commands.log.<name>`.
- Interactive prompts: patch `osism.commands.console.prompt` / `osism.commands.log.PromptSession` with `side_effect` sequences ending in `"exit"`/`"q"` so loops terminate.
- Use capsys for the tabulate/print-based assertions (`apply` role table, `get` tables, `check` script/table formats).

## Definition of Done

- [ ] `tests/unit/commands/test_apply.py`, `test_check.py`, `test_compose.py`, `test_sync.py`, `test_log.py`, `test_console.py` created
- [ ] `tests/unit/commands/test_validate.py`, `test_wait.py`, `test_get.py` extended (existing tests kept unchanged)
- [ ] `pytest --cov` shows ≥ 80 % for each module in scope (≥ 95 % for the pure helpers in `check.py` and the module-level functions in `console.py`)
- [ ] `pipenv run pytest tests/unit/commands/` passes locally

## Dependencies

berendt:
Implements #2363: unit tests for the workflow-oriented CLI command modules under `osism/commands/` (`apply`, `check`, `compose`, `sync`, `get`, `log`, `console`, `validate`, `wait`).

## What's covered

- **New test modules**: `test_apply.py`, `test_check.py`, `test_compose.py`, `test_sync.py`, `test_log.py`, `test_console.py`
- **Extended** (existing tests unchanged): `test_validate.py`, `test_wait.py`, `test_get.py`

All test targets from the issue are implemented, including the pure helpers in `check.py` and the module-level host-resolution helpers in `console.py` that the issue flags for ≥95 % coverage. All Celery task calls, subprocesses, Docker/Netbox/Redis clients and interactive prompts are mocked; log assertions use a loguru sink (`loguru_logs` fixture) since `caplog` does not capture loguru records.

## Misbehavior pinned as strict xfail

Per review, tests that would have pinned misbehavior green now assert the **corrected** behavior and are marked `xfail(strict=True)` — they turn red (XPASS) the moment the follow-up fix lands, forcing removal of the marker:

- `compose` mangles multi-token subcommands into unrunnable commands: `osism compose <host> <env> up -d` is sent as `up-d` (`"".join` instead of `" ".join`, compose.py:28).
- A successful collection segment silently skips the remaining `//` segments and exits 0, because `handle_collection` returns `None` instead of an exit code (apply.py:218, loop at apply.py:455).
- A failed loadbalancer child playbook does not propagate its rc — `handle_loadbalancer_task` returns only the parent rc and reads children solely for logging; the child rcs live in the task-output stream (`handle_task`/`fetch_task_output`), not in the Celery result (apply.py:106-133). Note: since the `-ng` role drop this handler is no longer reachable in production (see follow-ups).

## Remaining follow-up fix candidates (pinned as-is)

- Since the `-ng` roles were dropped (ebde575), no `_prepare_task` path returns a Celery group anymore, so the `GroupResult` dispatch in `handle_role` (apply.py:388) and `handle_loadbalancer_task` are unreachable in production — candidates for removal together with their tests (including the strict xfail on child-rc propagation).
- `handle_collection` under `--dry-run` logs "No tasks are scheduled" (and the `--dry-run` help says "do not initiate tasks"), yet still dispatches the noop group via `apply_async()` (apply.py:235, :261).
- `wait --refresh` is a no-op when task ids are passed on the CLI: `do_refresh` is only enabled when the id list comes from the inspector.
- `log.File` error logs use printf-style `%s` placeholders, which loguru does not interpolate — the placeholders appear literally in the log output.
- `validate.Run._handle_task` mixes `t.id` (fetch) and `t.task_id` (log/print), and its timeout message says "(sync inventory)".

`flake8` and `black --check` pass locally; the test suite itself runs in CI per project workflow.

Closes #2363

https://claude.ai/code/session_01WgHtkSVmEuwwPxiEsSnXL7

### DIFF
--- a/tests/unit/commands/test_apply.py
+++ b/tests/unit/commands/test_apply.py
@@ -1,0 +1,679 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for the ``osism apply`` command.
+
+The apply command routes a role to one of the celery task queues
+(osism-ansible, ceph-ansible, kolla-ansible, kubernetes) based on the
+role-to-environment mapping, expands collections into task groups and
+chains, and retries failed roles. These tests characterize:
+
+- ``_prepare_task``: environment resolution (explicit, mapped, ``custom``
+  fallback), prefix stripping (``ceph-``/``kolla-``), sub-environments,
+  the ``kolla_action`` extra argument and the osism-ansible runtime
+  override;
+- ``_handle_collection``/``handle_collection``: recursive group/chain
+  construction from ``Role`` trees, dry-run (noop tasks), show-tree
+  (log-only) mode and the scheduling/log contract;
+- ``handle_role``/``handle_loadbalancer_task``: exit-code pass-through,
+  the ``GroupResult`` dispatch and the child-task semantics (a group
+  failure propagates as an exception; a strict xfail pins that a failed
+  child must propagate its rc);
+- ``take_action``: task-lock enforcement, the role/environment table,
+  the time-based Ansible-facts freshness check, ``//`` role splitting,
+  retries and the collection branch (a strict xfail pins that a
+  successful collection must not skip the following ``//`` segments).
+"""
+
+import time
+import types
+from unittest.mock import MagicMock, call, patch
+
+import pytest
+
+from osism import utils as osism_utils
+from osism.commands import apply
+from osism.data import enums, playbooks
+from osism.data.enums import Role
+
+from ._helpers import assert_not_called_before_lock_check, make_command
+
+
+@pytest.fixture(autouse=True)
+def _reset_lazy_state():
+    """Keep lazily initialized module state from leaking between tests.
+
+    ``osism.data.playbooks`` caches the playbook maps injected via
+    ``_set_playbook_maps`` and ``take_action`` stores the facts-check
+    backoff timestamp on ``osism.utils``.
+    """
+    osism_utils.__dict__.pop("_last_ansible_facts_check", None)
+    yield
+    playbooks._reset_caches()
+    osism_utils.__dict__.pop("_last_ansible_facts_check", None)
+
+
+@pytest.fixture
+def task_mocks(mocker):
+    """Patch the celery task entry points used by ``_prepare_task``."""
+    return types.SimpleNamespace(
+        ansible_run=mocker.patch("osism.tasks.ansible.run"),
+        ansible_noop=mocker.patch("osism.tasks.ansible.noop"),
+        ceph_run=mocker.patch("osism.tasks.ceph.run"),
+        kolla_run=mocker.patch("osism.tasks.kolla.run"),
+        kubernetes_run=mocker.patch("osism.tasks.kubernetes.run"),
+    )
+
+
+@pytest.fixture
+def take_action_mocks(mocker):
+    """Patch the utils helpers consulted by ``take_action``."""
+    return types.SimpleNamespace(
+        lock=mocker.patch("osism.commands.apply.utils.check_task_lock_and_exit"),
+        facts=mocker.patch("osism.commands.apply.utils.check_ansible_facts"),
+    )
+
+
+def _set_playbook_maps(role2environment=None, role2runtime=None):
+    """Inject the lazy playbook maps; the autouse fixture resets them."""
+    playbooks.MAP_ROLE2ENVIRONMENT = role2environment or {}
+    playbooks.MAP_ROLE2RUNTIME = role2runtime or {}
+
+
+def _prepare_task(cmd, **overrides):
+    params = dict(
+        arguments=[],
+        environment=None,
+        overwrite=None,
+        sub=None,
+        role="testrole",
+        action="deploy",
+        wait=True,
+        format="log",
+        timeout=300,
+        task_timeout=3600,
+    )
+    params.update(overrides)
+    return cmd._prepare_task(**params)
+
+
+def _public_collection_kwargs(**overrides):
+    params = dict(
+        arguments=[],
+        environment=None,
+        overwrite=None,
+        sub=None,
+        collection="testcollection",
+        action="deploy",
+        wait=True,
+        format="log",
+        timeout=300,
+        task_timeout=3600,
+        retry=0,
+        dry_run=False,
+        show_tree=False,
+    )
+    params.update(overrides)
+    return params
+
+
+def _collection_kwargs(**overrides):
+    counter = overrides.pop("counter", 0)
+    params = _public_collection_kwargs(**overrides)
+    params["counter"] = counter
+    return params
+
+
+def _handle_role(cmd, **overrides):
+    params = dict(
+        arguments=[],
+        environment=None,
+        overwrite=None,
+        sub=None,
+        role="testrole",
+        action="deploy",
+        wait=True,
+        format="log",
+        timeout=300,
+        task_timeout=3600,
+    )
+    params.update(overrides)
+    return cmd.handle_role(**params)
+
+
+# _prepare_task
+
+
+def test_prepare_task_ceph_role_forces_ceph_environment(task_mocks):
+    _set_playbook_maps()
+    cmd = make_command(apply.Run)
+
+    t = _prepare_task(cmd, role="ceph", environment=None, arguments=["-v"])
+
+    task_mocks.ceph_run.si.assert_called_once_with(
+        "ceph", "ceph", ["-v"], auto_release_time=3600
+    )
+    assert t is task_mocks.ceph_run.si.return_value
+
+
+def test_prepare_task_strips_ceph_prefix(task_mocks):
+    _set_playbook_maps()
+    cmd = make_command(apply.Run)
+
+    t = _prepare_task(cmd, role="ceph-osds", environment="ceph")
+
+    task_mocks.ceph_run.si.assert_called_once_with(
+        "ceph", "osds", [], auto_release_time=3600
+    )
+    assert t is task_mocks.ceph_run.si.return_value
+
+
+def test_prepare_task_sub_environment_suffix(task_mocks):
+    _set_playbook_maps()
+    cmd = make_command(apply.Run)
+
+    _prepare_task(cmd, role="ceph-osds", environment="ceph", sub="zone-a")
+    _prepare_task(cmd, role="kubeadm", environment="kubernetes", sub="zone-a")
+    _prepare_task(cmd, role="keystone", environment="kolla", sub="zone-a")
+
+    task_mocks.ceph_run.si.assert_called_once_with(
+        "ceph.zone-a", "osds", [], auto_release_time=3600
+    )
+    task_mocks.kubernetes_run.si.assert_called_once_with(
+        "kubernetes.zone-a", "kubeadm", [], auto_release_time=3600
+    )
+    task_mocks.kolla_run.si.assert_called_once_with(
+        "kolla.zone-a", "keystone", ["-e kolla_action=deploy"], auto_release_time=3600
+    )
+
+
+def test_prepare_task_kubernetes_environment_from_mapping(task_mocks):
+    _set_playbook_maps(role2environment={"kubeadm": "kubernetes"})
+    cmd = make_command(apply.Run)
+
+    t = _prepare_task(cmd, role="kubeadm", arguments=["-l node1"])
+
+    task_mocks.kubernetes_run.si.assert_called_once_with(
+        "kubernetes", "kubeadm", ["-l node1"], auto_release_time=3600
+    )
+    assert t is task_mocks.kubernetes_run.si.return_value
+
+
+def test_prepare_task_kolla_strips_prefix_and_adds_action(task_mocks):
+    _set_playbook_maps()
+    cmd = make_command(apply.Run)
+
+    t = _prepare_task(
+        cmd, role="kolla-keystone", environment="kolla", arguments=["-l control"]
+    )
+
+    task_mocks.kolla_run.si.assert_called_once_with(
+        "kolla",
+        "keystone",
+        ["-e kolla_action=deploy", "-l control"],
+        auto_release_time=3600,
+    )
+    assert t is task_mocks.kolla_run.si.return_value
+
+
+def test_prepare_task_kolla_role_in_osism_ansible_runtime(task_mocks):
+    _set_playbook_maps(
+        role2environment={"keystone": "kolla"},
+        role2runtime={"osism-ansible": ["keystone"]},
+    )
+    cmd = make_command(apply.Run)
+
+    t = _prepare_task(cmd, role="keystone", arguments=["-l control"])
+
+    task_mocks.ansible_run.si.assert_called_once_with(
+        "kolla", "keystone", ["-l control"], auto_release_time=3600
+    )
+    task_mocks.kolla_run.si.assert_not_called()
+    assert t is task_mocks.ansible_run.si.return_value
+
+
+def test_prepare_task_common_role_stays_in_kolla(task_mocks):
+    _set_playbook_maps(
+        role2environment={"common": "kolla"},
+        role2runtime={"osism-ansible": ["common"]},
+    )
+    cmd = make_command(apply.Run)
+
+    _prepare_task(cmd, role="common")
+
+    task_mocks.kolla_run.si.assert_called_once_with(
+        "kolla", "common", ["-e kolla_action=deploy"], auto_release_time=3600
+    )
+    task_mocks.ansible_run.si.assert_not_called()
+
+
+def test_prepare_task_unknown_role_falls_back_to_custom(task_mocks, loguru_logs):
+    _set_playbook_maps()
+    cmd = make_command(apply.Run)
+
+    t = _prepare_task(cmd, role="myplay", arguments=["-v"])
+
+    task_mocks.ansible_run.si.assert_called_once_with(
+        "custom", "myplay", ["-v"], auto_release_time=3600
+    )
+    messages = [r["message"] for r in loguru_logs]
+    assert "Trying to run play myplay in environment custom" in messages
+    assert t is task_mocks.ansible_run.si.return_value
+
+
+def test_prepare_task_overwrite_replaces_default_environment(task_mocks):
+    _set_playbook_maps(role2environment={"myplay": "generic"})
+    cmd = make_command(apply.Run)
+
+    _prepare_task(cmd, role="myplay", overwrite="other")
+
+    task_mocks.ansible_run.si.assert_called_once_with(
+        "other", "myplay", [], auto_release_time=3600
+    )
+
+
+# _handle_collection
+
+
+def test_handle_collection_rejects_non_role_items(loguru_logs):
+    cmd = make_command(apply.Run)
+
+    with pytest.raises(TypeError, match="Expected Role object, got str"):
+        cmd._handle_collection(["not-a-role"], **_collection_kwargs())
+
+    assert any(
+        r["level"] == "ERROR" and "Expected Role object" in r["message"]
+        for r in loguru_logs
+    )
+
+
+def test_handle_collection_flat_roles_wrapped_in_group(mocker):
+    group_mock = mocker.patch("celery.group")
+    cmd = make_command(apply.Run)
+    prepared = [MagicMock(name="pt-a"), MagicMock(name="pt-b")]
+    cmd._prepare_task = MagicMock(side_effect=prepared)
+
+    result = cmd._handle_collection([Role("a"), Role("b")], **_collection_kwargs())
+
+    assert [c.args[4] for c in cmd._prepare_task.call_args_list] == ["a", "b"]
+    group_mock.assert_called_once_with(prepared)
+    assert result is group_mock.return_value
+
+
+def test_handle_collection_nested_dependencies_chained(mocker):
+    group_mock = mocker.patch("celery.group")
+    chain_mock = mocker.patch("celery.chain")
+    cmd = make_command(apply.Run)
+    parent_pt = MagicMock(name="pt-parent")
+    child_pt = MagicMock(name="pt-child")
+    cmd._prepare_task = MagicMock(side_effect=[parent_pt, child_pt])
+
+    result = cmd._handle_collection(
+        [Role("parent", dependencies=[Role("child")])], **_collection_kwargs()
+    )
+
+    chain_mock.assert_called_once_with(parent_pt, group_mock.return_value)
+    assert group_mock.call_args_list == [
+        call([child_pt]),
+        call([chain_mock.return_value]),
+    ]
+    assert result is group_mock.return_value
+
+
+def test_handle_collection_dry_run_uses_noop_tasks(mocker, task_mocks):
+    group_mock = mocker.patch("celery.group")
+    cmd = make_command(apply.Run)
+    cmd._prepare_task = MagicMock()
+
+    result = cmd._handle_collection(
+        [Role("a"), Role("b")], **_collection_kwargs(dry_run=True)
+    )
+
+    cmd._prepare_task.assert_not_called()
+    assert task_mocks.ansible_noop.si.call_count == 2
+    task_mocks.ansible_noop.si.assert_called_with()
+    group_mock.assert_called_once_with([task_mocks.ansible_noop.si.return_value] * 2)
+    assert result is group_mock.return_value
+
+
+def test_handle_collection_show_tree_only_logs(task_mocks, loguru_logs):
+    cmd = make_command(apply.Run)
+    cmd._prepare_task = MagicMock()
+
+    result = cmd._handle_collection(
+        [Role("parent", dependencies=[Role("child")])],
+        **_collection_kwargs(show_tree=True),
+    )
+
+    assert result is None
+    cmd._prepare_task.assert_not_called()
+    task_mocks.ansible_noop.si.assert_not_called()
+    messages = [r["message"] for r in loguru_logs]
+    assert "A [0] - parent" in messages
+    assert "A [1] -- child" in messages
+
+
+# handle_collection
+
+
+def test_handle_collection_applies_prepared_group(loguru_logs):
+    cmd = make_command(apply.Run)
+    prepared = MagicMock()
+    cmd._handle_collection = MagicMock(return_value=prepared)
+    collection_roles = [Role("a")]
+
+    with patch.dict(enums.MAP_ROLE2ROLE, {"testcollection": collection_roles}):
+        cmd.handle_collection(**_public_collection_kwargs())
+
+    args = cmd._handle_collection.call_args
+    assert args.args[0] is collection_roles
+    assert args.args[1] == 0
+    prepared.apply_async.assert_called_once_with()
+    messages = [r["message"] for r in loguru_logs]
+    assert "Collection testcollection is prepared for execution" in messages
+    assert (
+        "All tasks of the collection testcollection are prepared for execution"
+        in messages
+    )
+    assert "Tasks are running in the background" in messages
+
+
+def test_handle_collection_show_tree_does_not_apply(loguru_logs):
+    cmd = make_command(apply.Run)
+    prepared = MagicMock()
+    cmd._handle_collection = MagicMock(return_value=prepared)
+
+    with patch.dict(enums.MAP_ROLE2ROLE, {"testcollection": [Role("a")]}):
+        cmd.handle_collection(**_public_collection_kwargs(show_tree=True))
+
+    prepared.apply_async.assert_not_called()
+    messages = [r["message"] for r in loguru_logs]
+    assert "Showing execution tree for collection testcollection" in messages
+    assert "Tasks are running in the background" not in messages
+
+
+def test_handle_collection_dry_run_logs_but_still_applies(loguru_logs):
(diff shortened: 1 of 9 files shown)


tests/unit/commands/test_compose.py
```diff
@@ -0,0 +1,58 @@
...
+        "'docker compose --project-directory=/opt/production ps'"
+    )
+    mock_call.assert_called_once_with(expected, shell=True)
+
+
+def test_run_joins_remainder_arguments_without_separator():
+    mock_call = _run_compose(["testhost", "production", "up", "-d"])
+
+    command = mock_call.call_args[0][0]
+    assert "'docker compose --project-directory=/opt/production up-d'" in command
```
ideaship: This pins misbehavior that breaks any multi-token compose command: `osism compose <host> <env> up -d` is sent to the host as `docker compose … up-d` — an unrunnable command — and the test asserts that mangled `up-d` string green. Cause: `compose.py:28` uses `"".join(parsed_args.arguments)` instead of `" ".join(...)`. I checked the broken join isn't load-bearing: `compose.Run` has no internal callers, `arguments` feeds only the SSH string, and single-token subcommands (`ps`, `pull`, `down`) are identical under either join — so `" ".join` is a safe, strictly-better change. Since this is a test-only PR: assert `up -d` (red until the compose fix) or `xfail(strict=True)`; fix `"".join` → `" ".join` in a follow-up.

tests/unit/commands/test_apply.py
```diff
@@ -0,0 +1,679 @@
...
+    _set_playbook_maps()
+    cmd = make_command(apply.Run)
+    cmd.handle_collection = MagicMock(return_value=None)
+    cmd.handle_role = MagicMock()
+    parsed_args = cmd.get_parser("test").parse_args(["testcollection//other"])
+
+    with patch.dict(enums.MAP_ROLE2ROLE, {"testcollection": [Role("a")]}):
+        rc = cmd.take_action(parsed_args)
+
+    assert rc is None
```
ideaship: This pins misbehavior that silently drops deploy steps. `osism apply` lets you chain applies with `//` — it runs each `//`-separated segment in turn — so `osism apply nutshell//openstackclient` should apply the `nutshell` collection, then apply `openstackclient`. Instead the second segment is skipped entirely and the command still exits 0 — a partial deploy that scripts/CI read as success. The test asserts both `rc is None` and `handle_role.assert_not_called()` green. Cause: `handle_collection` (`apply.py:218`) has no `return`, so `take_action` (`:462`) gets `rc = None`, treats it as failure at `:477`, breaks the `//` loop before reaching `openstackclient`, and returns `None` (exit 0). I checked the `None` isn't relied on: `handle_collection` is called only at `:462`, and for a single, unchained collection the `None`→exit-0 is already today's result, so an int return only changes the chained-segment path — the bug itself. Since this is a test-only PR: assert the corrected behavior (collection returns `0`, the next segment runs; red until fixed) or `xfail(strict=True)`; give `handle_collection` an int return in a follow-up.

tests/unit/commands/test_apply.py
```diff
@@ -0,0 +1,679 @@
...
+    mocker.patch("osism.tasks.handle_task", return_value=0)
+    cmd = make_command(apply.Run)
+    t = MagicMock()
+    child = MagicMock(task_id="child-1")
+    child.get.return_value = 99
+    t.children = [child]
+
+    rc = cmd.handle_loadbalancer_task(t, True, "log", 300)
+
+    assert rc == 0
```
ideaship: This pins misbehavior that reports success when a load-balancer child playbook failed: a failed child leaves the command exiting 0, so scripted callers never see the failure (an operator watching the log still does — which is why it's stayed hidden). The test asserts `rc == 0` and `child.get.assert_not_called()` green. Cause: `handle_loadbalancer_task` (`apply.py:106-133`) returns only the parent rc and reads children solely to log; and `run_ansible_in_environment` (`tasks/__init__.py:411-419`) returns the output string while recording the rc separately via `finish_task_output(rc=...)` — it never raises on rc≠0, so `GroupResult.get()` succeeds regardless. The mock is also unrepresentative: production never calls `child.get()`, so `child.get.return_value = 99` models nothing — and `child.get()` can't supply the rc in any case, because the rc lives in the task-output stream, not in the Celery result. I checked nothing depends on discarding child rcs: the return flows only to `take_action` at `:411`, so propagating the worst child rc only changes the failure case. Since this is a test-only PR: assert that a failing child propagates to a non-zero return (red until fixed) or `xfail(strict=True)`. Follow-up fix: read each child's recorded rc the same way the parent's is read — `handle_task(child, ...)` (i.e. `fetch_task_output(child.id, ...)`), not `child.get()` — and propagate the worst.

tests/unit/commands/test_apply.py
```diff
@@ -0,0 +1,679 @@
...
+    with patch.dict(enums.MAP_ROLE2ROLE, {"testcollection": [Role("a")]}):
+        cmd.handle_collection(**_public_collection_kwargs(show_tree=True))
+
+    prepared.apply_async.assert_not_called()
+    messages = [r["message"] for r in loguru_logs]
+    assert "Showing execution tree for collection testcollection" in messages
+    assert "Tasks are running in the background" not in messages
+
+
+def test_handle_collection_dry_run_logs_but_still_applies(loguru_logs):
```
ideaship, line 394: Cleanup-only, not blocking: this test honestly documents a real contradiction — in `dry_run`, `handle_collection` logs "No tasks are scheduled" (`apply.py:235`) yet still calls `t.apply_async()` (`apply.py:261`) on the no-op group, dispatching a group that needs live workers and occupies the queue. It's not only the log: the `--dry-run` help (`apply.py:90`) also says "do not initiate tasks", so a wording-only fix would leave the help inaccurate too. Impact is mild (noop tasks do no deploy work). Follow-up: either skip `apply_async()` under `dry_run`, or make both the log and the `--dry-run` help say plainly that no-op tasks are scheduled.

ideaship (changes requested):

ideaship (approved):

Issue #2478, opened 2026-07-17
[bug] Generation of SONiC Config Breakout Port speed from Port Type fails


ZoTA02:
### OSISM release version

latest

### What's the problem?

The problem occurs with `osism --version`: osism 0.20260615.1
The problem was not present in `osism --version`: osism 0.20260414.1

We found the issue with the switch type `Edgecore 9726-32DB-O-AC-F`. 400G interfaces are broken out into 4x100G interfaces, which works.
However, in the 100G interface definition, the interface speed is wrong and the switch cannot apply the configuration file.

Expected example output from older versions:
```
    "Ethernet118": {
      "admin_status": "down",
      "alias": "Eth1/15/7",
      "index": "15",
      "lanes": "151,152",
      "speed": "100000",
      "mtu": "9100",
      "adv_speeds": "all",
      "autoneg": "off",
      "link_training": "off",
      "unreliable_los": "auto",
      "valid_speeds": "100000,50000,25000,10000,1000"
    },
```

Current, faulty output:
```
    "Ethernet118": {
      "admin_status": "down",
      "alias": "Eth1/15/7",
      "index": "15",
      "lanes": "151,152",
      "speed": "100",
      "mtu": "9100",
      "adv_speeds": "all",
      "autoneg": "off",
      "link_training": "off",
      "unreliable_los": "auto",
      "valid_speeds": "100,10000,1000"
    },
```

The issue seems to be that interface speeds from netbox can be in two different units (mbit/kbit) depending on the source (interface type/explicit).
The issue seems to have been introduced in [https://github.com/osism/python-osism/pull/2328](https://github.com/osism/python-osism/pull/2328).

Disclaimer: We partly used AI (Claude Opus 4.8) to analyze the issue. I add the full output, which includes suggested fixes and references to the code, below.

<details>

<summary>Complete AI output</summary>

> ============ Start of AI output ============

Your `speed` of `100` (instead of `100000`) comes from a **unit-conversion bug** in the breakout-port branch of `_add_port_configurations`. The value gets divided by 1000 when it shouldn't be.

Two different sources feed `netbox_interfaces[port_name]["speed"]`, and they use **different units**:

1. In config_generator.py, when the NetBox interface has an explicit `speed`, that value is in **kbps** (NetBox convention).
2. When `interface.speed` is unset, the code falls back to `get_speed_from_port_type(...)`, which returns **Mbps** (interface.py — e.g. `100gbase-x-qsfp28` → `100000` Mbps).

[...]

PR #2246, opened 2026-05-03
Add unit tests for SONiC config_generator port/interface helpers

berendt:
Covers the port / interface / portchannel / breakout helpers in osism/tasks/conductor/sonic/config_generator.py:

- _add_port_configurations: natural sort, master-port skip, admin_status, NetBox speed override (explicit vs derived, kbps→Mbps), breakout speed (NetBox + brkout_mode fallback), index propagation from master, default port_data shape, valid_speeds branches, and the post-loop _add_missing_breakout_ports / _add_tagged_vlans_to_ports invocations.
- _get_breakout_port_valid_speeds: known speeds and falsy inputs.
- _calculate_breakout_port_lane: 4-lane and 8-lane breakouts, range syntax, single-lane master, unexpected counts, out-of-bounds and regex no-match fallbacks.
- _add_missing_breakout_ports: skip-existing, NetBox speed kbps→Mbps conversion, brkout_mode fallback (incl. default), admin state, master index copy, valid_speeds override, alias kwargs.
- _add_tagged_vlans_to_ports: numeric VLAN ID sort, untagged ignored, unknown NetBox name skipped, port absent from PORT silently skipped.
- _add_interface_configurations: IPv4 base + suffixed entry, IPv6 link-local fallback, port-channel-member skip, disconnected skip, missing netbox_interfaces entry.
- _get_transfer_role_ipv4_addresses: happy path, dedup per interface, IPv6 ignored, mgmt/virtual filtered, invalid prefix and IP swallowed, IP without assigned_object_id, unknown id, top-level exception.
- _has_direct_ipv4_address, _has_transfer_role_ipv4, _is_untagged_vlan_member: happy path and empty-input early returns.
- _add_portchannel_configuration: populated and empty cases.

Also fixes a kbps→Mbps conversion bug in _add_missing_breakout_ports (flagged in review): the helper previously wrote the raw NetBox kbps value, while _add_port_configurations divides by 1000. With realistic NetBox units (e.g. 25 G = 25 000 000 kbps) this would emit "25000000" instead of "25000". The conversion is now consistent across both code paths, and the test exercises realistic units.

Closes #2222

AI-assisted: Claude Code

PR #2483, opened 2026-07-17
Add E2E golden tests for SONiC config generation

ideaship:
## Summary

Adds an end-to-end golden test for the SONiC config generator: provision NetBox on a local kind cluster, seed it with the netbox-manager example data plus a small regression-scenario overlay, run `sync_sonic()`, and compare every exported `config_db.json` against golden files committed under `tests/e2e/golden/`. Any behavior change in the generator becomes a reviewable golden diff in the same PR.

## Motivation

Unit tests verify the generator against fixture values that encode the author's assumptions, so a class of regressions — such as mixed kbps/Mbps interface speeds — can pass the unit suite while producing broken switch configurations. Nothing exercised the full pipeline (real NetBox data in, complete config_db out) until now:

- #2478 — the customer-visible breakage (breakout ports with `"speed": "100"`)
- #2246 — where that regression slipped in, past both author and review

## Changes

- **`SONIC_PORT_CONFIG_PATH`** setting so the generator can read the in-repo `files/sonic/port_config/` outside the container image
- **`tests/e2e/generate.py`** — drives `sync_sonic()`; since it returns only a device→config dict and swallows per-device failures, the driver asserts success itself (expected device set derived from golden filenames, empty-config detection, a loguru sink that fails the run on any ERROR record)
- **`tests/e2e/compare.py`** — exact file-set equality (missing / extra / mismatched reported separately), structural diff paths like `PORT|Ethernet4.speed`, canonical unified diff, `--regenerate` for intentional output changes
- **`tests/e2e/run.sh` + Makefile targets** `sonic-e2e`, `sonic-e2e-regen`, `sonic-e2e-up`, `sonic-e2e-down`; netbox-manager is installed from a checkout (sibling dir or `NETBOX_MANAGER_DIR`) into a dedicated venv
- **Golden files** for the five example-testbed switches, cross-checked against their independent sources (seed YAML, port_config `.ini` speeds, the ASN derivation rule) and reproduced byte-for-byte by a second seed-and-generate cycle
- **Breakout regression scenario** (`tests/e2e/scenario/`) — the base example models no breakout ports and sets no explicit interface speeds, so those code paths were unexercised. A second seeding pass adds two devices (a minimal synthetic `Accton-AS9726-32D` device type) that break a 400G master into a 4x100G group whose non-master sub-ports take the `_add_missing_breakout_ports` path; one derives the sub-port speed from the interface type, the other sets it explicitly in kbps, and both must yield `100000`. A tagged-VLAN port adds VLAN / VLAN_MEMBER coverage. The seed data is synthetic and modeled on real deployment conventions (no customer identifiers).
- **Zuul job `python-osism-sonic-e2e`** in check (restricted by a files matcher to changes that can alter generator output) and periodic-daily (unconditional, bounds seed-data drift); netbox-manager comes from `required-projects`, so `Depends-On` is honored for its code and seed data
- **Drive-by fix**: `_add_snmp_configuration` no longer builds the Redis-backed Ansible vault for devices without encrypted secrets — it logged three ERRORs per device in any environment without vault infrastructure (found by the new error guard)

## Verification

- Unit tests pass (2043); the comparator and driver logic are unit-tested (`tests/unit/e2e/`)
- Full local E2E cycle against kind + NetBox 4.5.10: golden bootstrap, then a compare-mode rerun against a fresh re-seed reproduced all seven goldens byte-for-byte
- **Regression guard verified**: reverting the speed-normalization fix against these goldens turns exactly the affected device red — the derived breakout device shows `100000 → 100` (the customer-visible symptom), the explicit device `100000 → 100000000` (raw kbps) — and the two revert variants isolate the two historical bugs; the current code keeps both green
- `python-osism-sonic-e2e` is green in CI, seeding the overlay and matching all seven goldens (the job validates itself on this PR — the files matcher covers `.zuul.yaml`, `tests/e2e/`, and the playbooks)

## Follow-up work

- Broaden coverage to the generator subsystems the current seed data still leaves unexercised: VXLAN / EVPN + VRF, port-channels, additional hwskus, DNS / NTP. The base example populates 16 config-db tables and emits roughly as many empty; each seeded device's golden pins its entire config, so coverage grows by seeding more representative devices. A higher-fidelity option is to generate against sanitized copies of real configs.
- Factor the kind / kubectl / helm pre-run into a shared zuul-jobs role instead of the copy adapted from netbox-manager's `pre-e2e.yml`.


berendt:
Let's merge https://github.com/osism/python-osism/pull/2488 first and rebase afterward.

ideaship:
Superseded. This is where the investigation started — the root cause was that the job had no headroom, consuming 80-92% of its 2400s budget as a matter of course, so a node-local apt stall pushed it over. The fix turned into a rebuild of the suite on docker compose plus synthetic in-repo fixtures, which is being reopened as a sequence of smaller PRs with a tracking issue to collect them. Two production bugs it surfaced are already out separately: #2557 and #2560.

PR #2485, opened 2026-07-18
Add unit tests for osism/api.py HTTP and WebSocket endpoints

### LINKED ISSUE #2360: Unit tests for osism/api.py — HTTP & WebSocket endpoints (FastAPI TestClient)
## Background

Follow-up to #2192 (foundation) and PR #2193 (pytest + Zuul infrastructure). Part of **Tier 7** (#2199). `osism/api.py` (1161 LOC) implements the OSISM API server (FastAPI). Because of its size the module is split into two sub-issues: this issue covers the **route handlers (HTTP + WebSocket) exercised through `fastapi.testclient.TestClient`**; the companion issue *"Unit tests for `osism/api.py` — helper functions, Pydantic models, secret masking"* (Tier 7, split 1/2) covers `_mask_inventory_secrets()`, `find_device_by_identifier()`, `process_netbox_webhook()` and the models directly.

**Dev dependency required:** `fastapi` itself is a runtime dependency (`requirements.txt:13`; CI installs the package via `pipenv run pip install .`, see `playbooks/test-unit.yml`), but `TestClient` is httpx-based since Starlette 0.21 and **`httpx` is not declared anywhere** (not in `Pipfile`, `Pipfile.lock` or `requirements.txt`; `import httpx` fails in the current pipenv environment). Add `httpx` to `[dev-packages]` in the `Pipfile` (pinned like the other dev packages, e.g. `httpx = "==0.28.1"`) and relock as part of this issue.

## Scope

Add `tests/unit/test_api_endpoints.py` covering all HTTP endpoints and the WebSocket endpoint of `osism/api.py` via `TestClient(app)`. There is currently no test coverage for `osism/api.py`. Direct unit tests for the helper functions and models are out of scope (companion issue) — but response-level effects such as secret masking in hostvars responses belong here. This issue is large (20 routes); it may be split further during implementation, as was done for Tier 3 issues.

### Health endpoints — `root()` `api.py:293`, `v1()` `api.py:299`, `events_info()` `api.py:305`

No patching needed.

- `GET /` → 200, `{"result": "ok"}`
- `GET /v1` → 200, `{"result": "ok"}`
- `GET /v1/events` → 200, body contains `websocket_endpoint == "/v1/events/openstack"`

### Telemetry sinks — `write_sink_meters()` `api.py:378`, `write_sink_events()` `api.py:396`

No patching needed (handlers only parse JSON and log).

- `POST /v1/meters/sink` with a JSON list → 200, `{"result": "ok"}`
- `POST /v1/meters/sink` with a JSON object → 200
- `POST /v1/meters/sink` with a malformed JSON body (`content=b"{not json"`) → 500, detail `"Failed to process meters data"`
- Same three cases for `POST /v1/events/sink` (detail `"Failed to process events data"`)

### `get_baremetal_nodes_list()` — `api.py:416`

Patch `osism.api.openstack.get_baremetal_nodes`.

- Returns two node dicts → 200, `count == 2`, fields mapped into `BaremetalNode` entries
- Returns `[]` → 200, `count == 0`
- Raises → 500, detail starts with `"Failed to retrieve baremetal nodes:"` and includes the exception text

### NetBox info — `get_baremetal_node_netbox_info()` `api.py:443`, `get_baremetal_nodes_netbox_info()` `api.py:464`

Patch `osism.api.openstack.get_baremetal_node_netbox_info` / `osism.api.openstack.get_baremetal_nodes_netbox_info`.

- `GET /v1/baremetal/nodes/{node_name}/netbox` happy path → 200, `device_role`/`primary_ip4`/`primary_ip6`/`netbox_url` mapped; task function called with `node_name`
- Single-node lookup raises → 500, detail includes the node name
- `POST /v1/baremetal/nodes/netbox` with `{"node_names": ["n1", "n2"]}` → 200, `nodes` keyed by name; task function called with the list
- `POST` body missing `node_names` → 422 (FastAPI validation, handler never called)
- Bulk lookup raises → 500, detail `"Failed to retrieve NetBox info: ..."`

### `get_baremetal_node_ports()` `api.py:488`, `get_baremetal_node_parameters()` `api.py:507`

Patch `osism.api.openstack.get_baremetal_node_ports` / `osism.api.openstack.get_baremetal_node_parameters`.

- Ports happy path → 200, `count` matches, port fields mapped; raises → 500 with node UUID in detail
- Parameters happy path (`kernel_append_params`, `netplan_parameters`, `frr_parameters`) → 200; raises → 500 with node UUID in detail

### `notifications_baremetal()` — `api.py:526`

Patch `osism.api.baremetal_events.get_handler` (method on the module-level `BaremetalEvents` instance) via `mocker.patch.object`.

- Valid notification body → 204 with empty body; `get_handler` called with `event_type`, returned handler called with `payload`
- Handler raises → 500, detail `"Failed to process baremetal notification"`
- `message_id` not a UUID → 422
- Optional: without patching, an unknown `event_type` (e.g. `"foo.bar.baz.qux"`) with `payload={"ironic_object.data": {"name": "n1"}}` exercises the real default handler → 204

### `sonic_ztp_complete()` — `api.py:549`

Patch `osism.utils.nb` (see hints) and `osism.api.find_device_by_identifier` (already unit-tested in the companion issue).

- `utils.nb` is `None` → 503, detail `"NetBox is not enabled"`
- Device found → 200, `{"result": "ok", "device": <name>}`; `device.custom_fields["provision_state"]` set to `"active"` and `device.save()` called
- No device found → 404, detail contains the identifier (the inner `HTTPException` is re-raised unchanged through the generic handler)
- `device.save()` raises → 500, detail `"Failed to complete ZTP process"`

### `webhook()` — `api.py:704`

Patch `osism.utils.nb` and `osism.api.process_netbox_webhook` (its branch logic is unit-tested in the companion issue).

- `utils.nb` is `None` → 503, detail `"NetBox webhook processing is not enabled"`
- Valid body → 200, `{"result": "ok"}`; `process_netbox_webhook` called with the parsed `WebhookNetboxData`
- `process_netbox_webhook` raises → 500, detail `"Failed to process NetBox webhook"`
- Body missing a required field (e.g. `request_id`) → 422
- Note: `content_length: int = Header(...)` is required, but httpx/TestClient sets `Content-Length` automatically for JSON bodies; `x_hook_signature` is optional and currently unvalidated

### `websocket_openstack_events()` — `api.py:632`

Patch `osism.api.websocket_manager` with a stub whose async `connect` accepts the socket (see hints) — this avoids starting the real broadcaster background task.

- `client.websocket_connect("/v1/events/openstack")` → `connect` awaited with the WebSocket
- Send `{"action": "set_filters", "event_filters": [...], "node_filters": [...], "service_filters": [...]}` → `update_filters` awaited with exactly these values; acknowledgment frame received: `{"type": "filter_update", "status": "success", ...}` echoing the filters
- `set_filters` with only some keys → missing ones passed as `None` and echoed as `null` in the ack
- Send invalid JSON text → no ack, no `update_filters` call, connection stays usable (a subsequent valid `set_filters` still gets an ack)
- Send valid JSON with `action != "set_filters"` → no ack, no `update_filters` call
- `update_filters` raising → inner `except Exception` logs and the loop continues (connection still open)
- Closing the client context → `disconnect` awaited (the `finally` block)

### `get_inventory_hosts()` — `api.py:736`

Patch `osism.api.get_inventory_path`, `osism.api.os.path.exists`, `osism.api.subprocess.run`, `osism.api.get_hosts_from_inventory` (the inventory helpers themselves are covered by #2194).

- Inventory file missing (`os.path.exists` → `False`) → 503, detail contains the path
- `subprocess.run` result with `returncode != 0` → 500, `"Failed to load Ansible inventory"`
- Happy path → 200, `hosts` + `count`; assert the command is `["ansible-inventory", "-i", <path>, "--list"]`
- `?limit=compute*` → command extended with `["--limit", "compute*"]`
- `subprocess.TimeoutExpired` → 504
- Non-JSON stdout → 500, `"Failed to parse Ansible inventory"`
- `get_hosts_from_inventory` raises → 500, generic `"Failed to retrieve hosts: ..."`

### Hostvars — `get_host_hostvars()` `api.py:803`, `get_host_hostvar()` `api.py:866`

Patch `osism.api.get_inventory_path` and `osism.api.subprocess.run`.

- Happy path → 200; variables sorted by name, `count` correct; assert `get_inventory_path` called with `prefer_minified=False` and the command uses `--host <host>`
- Secret masking applied in the response: stdout containing `{"ansible_password": "x", "vaulted": "$ANSIBLE_VAULT;1.1;AES256\n..."}` → both values are `"***"`
- `returncode != 0` with `"Could not match"` in stderr → 404, `"Host '<host>' not found in inventory"`
- `returncode != 0` with `"Unable to parse"` in stderr → 404
- `returncode != 0` with other stderr → 500
- `subprocess.TimeoutExpired` → 504
- Non-JSON stdout → 500
- Single-variable endpoint: variable present → 200 with `host`/`name`/`value`; secret variable (e.g. `database_password`) → value `"***"` (masking happens after the membership check)
- Variable not in hostvars → 404, `"Variable '<variable>' not found for host '<host>'"`
- Host not found / timeout / bad JSON → 404 / 504 / 500 as above

### Facts — `get_host_facts()` `api.py:932`, `get_host_fact()` `api.py:971`

Patch `utils.redis` — the lazy attribute on `osism.utils` — via `monkeypatch.setattr(osism.utils, "redis", fake_redis, raising=False)`.

- `redis.get` returns `None` → 404, `"No facts found in cache for host '<host>'"`; assert the (separator-less) cache key `f"ansible_facts{host}"` is used
- Returns JSON bytes → 200; facts sorted by name, `count` correct, `from_cache is True`
- Returns invalid JSON → 500, `"Failed to parse facts for <host>"`
- `redis.get` raises → 500, generic `"Failed to retrieve facts: ..."`
- Single-fact endpoint: fact present → 200 with `value` and `from_cache is True`; fact missing → 404 `"Fact '<fact>' not found..."`; no cache entry → 404; invalid JSON → 500

### `search_inventory()` — `api.py:1010`

Patch `osism.api.get_inventory_path`, `osism.api.os.path.exists`, `osism.api.subprocess.run`, `osism.api.get_hosts_from_inventory`, and `builtins.open` (facts are read from `/cache/facts/<host>`).

- Missing `name_pattern` query parameter → 422
- Invalid `name_pattern` regex (`"["`) → 400, `"Invalid name_pattern regex: ..."`
- Invalid `host_pattern` regex → 400
- `source=bogus` → 400, `"source must be 'hostvars', 'facts', or omitted for both"`
- Inventory file missing → 503
- Inventory `--list` call with `returncode != 0` → 500
- Hostvars search happy path: `--list` then one `--host` call per host (`subprocess.run` `side_effect` list); only names matching `name_pattern` (case-insensitive) returned with `source == "hostvars"`; secret values masked (`"***"`) before matching
- `host_pattern` filters the host list; `hosts_searched` reflects the filtered count
- `source=facts`: no per-host `ansible-inventory --host` calls; facts file read via `open`, matching fact names returned with `source == "facts"`; facts file absent → host contributes nothing
- `limit` honored: more matches than `limit` → exactly `limit` results, host loop breaks early
- Per-host `subprocess.TimeoutExpired` or bad hostvars JSON → warning, host skipped, search continues
- Facts file with invalid JSON / `IOError` on open → warning, search continues
- Response `query` dict echoes `name_pattern`, `host_pattern`, `source`, `limit`
- Two `get_inventory_path` calls: `prefer_minified=False` for hostvars, default (minified) for the host list

## Mocking hints

- Add `httpx` to `[dev-packages]` in the `Pipfile` first — `from fastapi.testclient import TestClient` raises without it (Starlette ≥ 0.21 TestClient is httpx-based).
- Module-scoped fixture:
  ```python
  from fastapi.testclient import TestClient
  from osism.api import app

  @pytest.fixture(scope="module")
  def client():
      return TestClient(app)
  ```
- Importing `osism.api` instantiates `EventBridge()` (`osism/services/event_bridge.py:304`), which attempts `redis.Redis(...).ping()` against host `redis`; the exception is caught and logged, so no Redis is needed — just expect the log line. The import also runs `dictConfig`.
- `osism.api` does `from osism.tasks import reconciler, openstack` — attribute lookups happen on the module objects, so `mocker.patch("osism.api.openstack.get_baremetal_nodes")` patches the same object as `osism.tasks.openstack.get_baremetal_nodes`. Tests call the patched task *functions* synchronously via the endpoint; no Celery broker is involved.
- `utils.nb` and `utils.redis` are lazy attributes materialized by `osism.utils.__getattr__`; use `monkeypatch.setattr(osism.utils, "nb", fake, raising=False)` (a plain `mocker.patch("osism.utils.nb")` would trigger a real connection attempt during the initial `getattr`).
- Fake `subprocess.run` results: `subprocess.CompletedProcess(args=[...], returncode=0, stdout=json.dumps(payload), stderr="")` (the code passes `text=True`, so `stdout` is a `str`). Timeouts: `side_effect=subprocess.TimeoutExpired(cmd="ansible-inventory", timeout=30)`.
- WebSocket stub — the real `websocket_manager.connect` calls `websocket.accept()`; a bare `AsyncMock` would never accept and `client.websocket_connect` would fail. Use:
  ```python
  fake_manager = mocker.MagicMock()
  async def fake_connect(ws):
      await ws.accept()
  fake_manager.connect = mocker.AsyncMock(side_effect=fake_connect)
  fake_manager.update_filters = mocker.AsyncMock()
  fake_manager.disconnect = mocker.AsyncMock()
  mocker.patch("osism.api.websocket_manager", fake_manager)
  ```
  Then `with client.websocket_connect("/v1/events/openstack") as ws: ws.send_text(...); ws.receive_json()`.
- For the facts branch of `search_inventory`, patch `osism.api.os.path.exists` with a `side_effect` keyed on the path (inventory path → `True`, `/cache/facts/<host>` → as needed) and `builtins.open` with `mock_open(read_data=json.dumps(facts))`.
- A minimal valid webhook body (JSON) for `POST /v1/webhook/netbox`:
  ```json
  {
    "username": "admin",
    "data": {"url": "/api/dcim/devices/1/", "name": "sw1",
             "tags": [{"name": "Managed by OSISM"}],
             "custom_fields": {"device_type": "server"}},
    "snapshots": {},
    "event": "updated",
    "timestamp": "2026-01-01T00:00:00Z",
    "model": "device",
    "request_id": "f8a6e0a4-3c1b-4f53-9c5f-2f4f3b2a1e00"
  }
  ```

## Definition of Done

- [ ] `httpx` added to `[dev-packages]` in `Pipfile` (and `Pipfile.lock` updated)
- [ ] `tests/unit/test_api_endpoints.py` created
- [ ] `pytest --cov=osism.api` shows ≥ 90 % (together with the companion helper issue)
- [ ] `pipenv run pytest tests/unit/test_api_endpoints.py` passes locally

## Dependencies

- Companion issue: *Unit tests for `osism/api.py` — helper functions, Pydantic models, secret masking* (Tier 7, split 1/2).

berendt:
Adds `tests/unit/test_api_endpoints.py`, exercising all HTTP endpoints and the WebSocket endpoint of `osism/api.py` through `fastapi.testclient.TestClient`, as specified in #2360:

- **Health / telemetry**: `/`, `/v1`, `/v1/events`, and both sink endpoints (JSON list, JSON object, malformed body → 500 with endpoint-specific detail)
- **Baremetal**: node list mapping into `BaremetalNode`, empty list, NetBox info for single and multiple nodes (including 422 for a missing `node_names` field), ports, parameters, and error paths asserting the node name/UUID in the detail
- **Notifications**: handler dispatch via `BaremetalEvents`, handler errors → 500, non-UUID `message_id` → 422, and the real default handler for unknown event types → 204
- **SONiC ZTP complete**: 503 without NetBox, `provision_state` update plus `device.save()`, 404 for unknown identifiers, save failures → 500
- **NetBox webhook**: 503 without NetBox, parsed `WebhookNetboxData` handed to `process_netbox_webhook`, processing errors → 500, missing required field → 422
- **WebSocket `/v1/events/openstack`**: connect/disconnect lifecycle, `set_filters` acknowledgment with full and partial filters, ignored invalid JSON and non-filter actions, connection survival after `update_filters` errors (stubbed websocket manager, so no broadcaster task is started)
- **Inventory**: hosts listing with `--limit` handling, hostvars with secret masking and sorted variables, ansible-inventory error mapping (404/500/504), facts from the Redis cache with the separator-less cache key, and the search endpoint (regex/source validation, host filtering, facts source, result limit with early loop exit, masking before matching, per-host failure tolerance, query echo, and both `get_inventory_path` call variants)

`httpx` was already added to the `Pipfile` dev packages by #2359, so no dependency changes are needed. The module-level helpers and Pydantic models are covered separately by `tests/unit/test_api_helpers.py`.

Closes #2360

https://claude.ai/code/session_01P5LudkHsw8ZXnyFmRHQDRE

### DIFF
--- a/tests/unit/test_api_endpoints.py
+++ b/tests/unit/test_api_endpoints.py
@@ -1162,6 +1162,7 @@ def test_search_limit_stops_early(client, mocker):
     body = response.json()
     assert body["count"] == 2
     assert len(body["results"]) == 2
+    assert body["hosts_searched"] == 1  # node-2 is skipped once the limit is hit
     assert body["query"]["limit"] == 2
     assert run.call_count == 2  # node-2 is never queried once the limit is hit
 
--- a/tests/unit/test_api_endpoints.py
+++ b/tests/unit/test_api_endpoints.py
@@ -1167,6 +1167,12 @@ def test_search_limit_stops_early(client, mocker):
     assert run.call_count == 2  # node-2 is never queried once the limit is hit
 
 
+@pytest.mark.parametrize("limit", [0, -1], ids=["zero", "negative"])
+def test_search_rejects_non_positive_limit(client, limit):
+    response = search(client, name_pattern="ansible", limit=limit)
+    assert response.status_code == 422
+
+
 @pytest.mark.parametrize(
     "failure",
     [subprocess.TimeoutExpired(cmd="ansible-inventory", timeout=10), "{not json"],
--- a/tests/unit/test_api_endpoints.py
+++ b/tests/unit/test_api_endpoints.py
@@ -758,7 +758,7 @@ def test_get_host_hostvars_masks_secrets(client, mocker):
     "stderr,expected_status",
     [
         ("Could not match supplied host pattern: node-1", 404),
-        ("Unable to parse /inventory/hosts.yml", 404),
+        ("Unable to parse /inventory/hosts.yml", 500),
         ("some other error", 500),
     ],
     ids=["no-match", "unparsable", "other"],
--- a/tests/unit/test_api_endpoints.py
+++ b/tests/unit/test_api_endpoints.py
@@ -1,0 +1,1205 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Unit tests for the HTTP and WebSocket endpoints in :mod:`osism.api`.
+
+All endpoints are exercised through ``fastapi.testclient.TestClient`` against
+the module-level ``app``. The module-level helpers and Pydantic models are
+covered separately in ``test_api_helpers.py``; response-level effects such as
+secret masking in hostvars responses belong here.
+
+Importing :mod:`osism.api` has import-time side effects: it applies
+``dictConfig(LogConfig().model_dump())`` and pulls in
+``osism.services.event_bridge``, whose module-level ``EventBridge()``
+instantiation attempts a Redis connection. That failure is caught and only
+logged, so these tests need no Redis.
+
+``osism.utils`` materializes its ``nb`` NetBox connection and ``redis`` client
+lazily via a module ``__getattr__``; tests therefore inject fakes with
+``patch.dict("osism.utils.__dict__", {...})`` instead of
+``mocker.patch("osism.utils.nb")``, which would trigger a real connection
+attempt when reading the original attribute.
+
+The endpoints call the Celery task functions in ``osism.tasks.openstack``
+synchronously; patching ``osism.api.openstack.<task>`` replaces the attribute
+on the shared module object, so no Celery broker is involved.
+"""
+
+import json
+import os
+import subprocess
+from unittest.mock import AsyncMock, MagicMock, call, mock_open, patch
+from uuid import UUID
+
+import pytest
+from fastapi import WebSocket
+from fastapi.testclient import TestClient
+
+from osism import api
+
+INVENTORY_PATH = "/inventory/hosts.yml"
+FULL_INVENTORY = "/inventory/hosts.yml"
+MINIFIED_INVENTORY = "/inventory/hosts-minified.yml"
+
+
+@pytest.fixture(scope="module")
+def client():
+    return TestClient(api.app)
+
+
+def completed(stdout="", returncode=0, stderr=""):
+    """Build a CompletedProcess as returned by subprocess.run(text=True)."""
+    return subprocess.CompletedProcess(
+        args=["ansible-inventory"], returncode=returncode, stdout=stdout, stderr=stderr
+    )
+
+
+# ---------------------------------------------------------------------------
+# Health endpoints
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("endpoint", ["/", "/v1"])
+def test_health_endpoints_return_ok(client, endpoint):
+    response = client.get(endpoint)
+    assert response.status_code == 200
+    assert response.json() == {"result": "ok"}
+
+
+def test_events_info_points_to_websocket_endpoint(client):
+    response = client.get("/v1/events")
+    assert response.status_code == 200
+    body = response.json()
+    assert body["result"] == "ok"
+    assert body["websocket_endpoint"] == "/v1/events/openstack"
+
+
+# ---------------------------------------------------------------------------
+# Telemetry sinks
+# ---------------------------------------------------------------------------
+
+
+@pytest.mark.parametrize("endpoint", ["/v1/meters/sink", "/v1/events/sink"])
+@pytest.mark.parametrize(
+    "body",
+    [[{"counter_name": "cpu"}, {"counter_name": "memory"}], {"counter_name": "cpu"}],
+    ids=["list", "object"],
+)
+def test_sink_accepts_json(client, endpoint, body):
+    response = client.post(endpoint, json=body)
+    assert response.status_code == 200
+    assert response.json() == {"result": "ok"}
+
+
+@pytest.mark.parametrize(
+    "endpoint,detail",
+    [
+        ("/v1/meters/sink", "Failed to process meters data"),
+        ("/v1/events/sink", "Failed to process events data"),
+    ],
+)
+def test_sink_rejects_malformed_json(client, endpoint, detail):
+    response = client.post(endpoint, content=b"{not json")
+    assert response.status_code == 500
+    assert response.json()["detail"] == detail
+
+
+# ---------------------------------------------------------------------------
+# get_baremetal_nodes_list
+# ---------------------------------------------------------------------------
+
+
+def test_baremetal_nodes_returns_mapped_nodes(client, mocker):
+    nodes_data = [
+        {
+            "uuid": "uuid-1",
+            "name": "node-1",
+            "power_state": "power on",
+            "provision_state": "active",
+            "maintenance": False,
+            "traits": ["CUSTOM_TRAIT"],
+            "properties": {"cpus": 8},
+        },
+        {
+            "uuid": "uuid-2",
+            "name": "node-2",
+            "power_state": "power off",
+            "provision_state": "available",
+            "maintenance": True,
+            "maintenance_reason": "repair",
+        },
+    ]
+    mocker.patch("osism.api.openstack.get_baremetal_nodes", return_value=nodes_data)
+
+    response = client.get("/v1/baremetal/nodes")
+
+    assert response.status_code == 200
+    body = response.json()
+    assert body["count"] == 2
+    first, second = body["nodes"]
+    assert first["uuid"] == "uuid-1"
+    assert first["name"] == "node-1"
+    assert first["power_state"] == "power on"
+    assert first["provision_state"] == "active"
+    assert first["maintenance"] is False
+    assert first["traits"] == ["CUSTOM_TRAIT"]
+    assert first["properties"] == {"cpus": 8}
+    assert first["driver"] is None
+    assert second["maintenance"] is True
+    assert second["maintenance_reason"] == "repair"
+
+
+def test_baremetal_nodes_empty_list(client, mocker):
+    mocker.patch("osism.api.openstack.get_baremetal_nodes", return_value=[])
+    response = client.get("/v1/baremetal/nodes")
+    assert response.status_code == 200
+    assert response.json() == {"nodes": [], "count": 0}
+
+
+def test_baremetal_nodes_error(client, mocker):
+    mocker.patch(
+        "osism.api.openstack.get_baremetal_nodes",
+        side_effect=RuntimeError("ironic down"),
+    )
+    response = client.get("/v1/baremetal/nodes")
+    assert response.status_code == 500
+    detail = response.json()["detail"]
+    assert detail.startswith("Failed to retrieve baremetal nodes:")
+    assert "ironic down" in detail
+
+
+# ---------------------------------------------------------------------------
+# NetBox info endpoints
+# ---------------------------------------------------------------------------
+
+
+def test_baremetal_node_netbox_info(client, mocker):
+    info = mocker.patch(
+        "osism.api.openstack.get_baremetal_node_netbox_info",
+        return_value={
+            "device_role": "compute",
+            "primary_ip4": "10.0.0.1/24",
+            "primary_ip6": "fd00::1/64",
+            "netbox_url": "https://netbox/dcim/devices/1/",
+        },
+    )
+
+    response = client.get("/v1/baremetal/nodes/node-1/netbox")
+
+    assert response.status_code == 200
+    assert response.json() == {
+        "device_role": "compute",
+        "primary_ip4": "10.0.0.1/24",
+        "primary_ip6": "fd00::1/64",
+        "netbox_url": "https://netbox/dcim/devices/1/",
+    }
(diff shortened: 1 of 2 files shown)


tests/unit/test_api_endpoints.py
```diff
@@ -0,0 +1,1205 @@
...
+        {"name": "vaulted", "value": "***"},
+    ]
+
+
+@pytest.mark.parametrize("url", HOSTVARS_URLS, ids=HOSTVARS_URL_IDS)
+@pytest.mark.parametrize(
+    "stderr,expected_status",
+    [
+        ("Could not match supplied host pattern: node-1", 404),
+        ("Unable to parse /inventory/hosts.yml", 404),
```
ideaship: Both hostvars handlers (`osism/api.py:829` and `:892`) fold `Unable to parse` into the same 404 branch as `Could not match`, so a corrupt or unreadable inventory is reported to the client as `Host '<host>' not found in inventory` — a server/config fault surfaced as a missing resource, which sends an operator hunting for a host that may well exist. This row pins both the wrong status and the false detail string.

Rather than dropping the row, please fix it properly in a separate commit in this PR: split the guard so `Could not match` stays 404 while `Unable to parse` returns a 500-class status, then update this row to expect that. Keep the `Could not match -> 404` and `other error -> 500` rows unchanged.

tests/unit/test_api_endpoints.py
```diff
@@ -0,0 +1,1205 @@
...
+        mocker,
+        hosts=["node-1", "node-2"],
+        hostvars={"node-1": failure, "node-2": {"ansible_host": "10.0.0.2"}},
+    )
+
+    response = search(client, name_pattern="ansible", source="hostvars")
+
+    assert response.status_code == 200
+    body = response.json()
+    assert body["hosts_searched"] == 2
```
ideaship, line 1192: `hosts_searched` is `len(hosts_to_search)` computed before the loop (`osism/api.py:1095`), so it counts hosts the `limit` break at `:1098` never visits — it's a candidate count, not hosts actually searched. In a separate commit, initialize it to 0 and increment once per host entered in the loop, then update the assertions to the corrected counts.

This particular case stays `== 2` (an attempted-but-failed host is still searched); the overcount to fix is the limit-short-circuit path, so add a `hosts_searched` assertion to `test_search_limit_stops_early`, which currently only checks `run.call_count`.

tests/unit/test_api_endpoints.py
```diff
@@ -0,0 +1,1205 @@
...
+            "host": "node-1",
+            "name": "ansible_hostname",
+            "value": "node-1",
+            "source": "facts",
+        }
+    ]
+    assert run.call_count == 1  # only the --list call, no per-host lookups
+
+
+def test_search_limit_stops_early(client, mocker):
```
ideaship, line 1149: `limit` has no lower bound (`osism/api.py:1021`, `limit: int = 100`), so `limit=0` or a negative value passes validation and reaches the accumulation loop. In a separate commit, constrain it (e.g. `limit: int = Query(default=100, gt=0)`) and add a test asserting a 422 for non-positive values. Nothing here asserts the current behavior, so no existing row needs to change.

ideaship (changes requested):

ideaship (approved):

ideaship (approved):

PR #2501, opened 2026-07-22
Cap openstacksdk <4.11 to fix openstack.cloud collection break

ideaship:
## Problem

openstacksdk `4.17.0` is incompatible with the galaxy-pinned
`openstack.cloud` `2.5.0` ansible collection. The collection's shared
`openstack_cloud_from_module` helper references `openstack.version`, which the
newer sdk no longer exposes, so every `openstack.cloud` module in the
osism-ansible manager container fails with:

```
AttributeError: module 'openstack' has no attribute 'version'
```

This broke **all** testbed `periodic-midnight` jobs on 2026-07-22 (they passed
the night before). The sdk bump rode in via:

- osism/python-osism#2292
- osism/python-osism#2395

## Change

Revert to the proven-good set that matches collection `2.5.0`, and cap it so it
cannot silently drift again:

- `requirements.txt` + `Pipfile` + `Pipfile.lock`: `openstacksdk` `4.17.0` → `4.10.0`,
  `python-openstackclient` `10.2.1` → `9.0.0` (openstackclient 10.x requires
  openstacksdk `>=4.16`, so the two must move together).
- `.github/renovate.json`: `allowedVersions` caps (`openstacksdk <4.11`,
  `python-openstackclient <10`) covering **all** managers, so neither
  `requirements.txt` (pip_requirements) nor the `Pipfile` (pipenv) can outrun the
  collection again.

`Pipfile.lock` was regenerated with `pipenv lock` (re-adds `requestsexceptions`
for the 4.10.0 tree; incidentally refreshes `certifi` and `platformdirs`).

## Follow-up

The caps are the compatibility contract: raise them **only together** with an
`openstack.cloud` bump in `osism/release`. Longer term, the sdk is pinned
redundantly in both `requirements.txt` and the `Pipfile` (two Renovate managers,
no generation link) — that duplication should be single-sourced.


berendt:
Not required. Bump of the OpenStack collection was sufficient.

PR #2532, opened 2026-07-30
SONiC E2E: provision NetBox with docker compose

ideaship:
## Purpose

Draft, opened to get a CI timing number for the compose migration. It extends the SONiC E2E harness from:

- osism/python-osism#2483

with four scenario commits and then replaces the kind + Helm NetBox fixture with docker compose. Not for review yet — see *Relationship to #2483* below.

## Why

`python-osism-sonic-e2e` timed out (buildset [`8ea91237`](https://zuul.services.osism.tech/t/osism/buildset/8ea91237f334425183c7818867cda1f4), 2457s). It was aborted mid-generation, ~2 minutes short of finishing.

The job did not hang. Zuul's `timeout` is one budget spanning **pre-run + run**, and the job normally consumed 80–92% of its 2400s: seven successive runs of identical code took 1919–2212s pre+run. What tipped it over was node-local — `ensure-pip : Update package lists` took **301s** instead of the usual 2–6s, adding ~409s to the pre-run. Not shared-mirror contention: the sibling jobs in the same buildset ran that identical task in 5s, 2s and 2s.

The underlying problem was that ~1650s of a ~2100s run was fixed setup: `helm install --wait` 704–792s and seeding 888–1011s.

## What was measured

Kubernetes contributed nothing the test used, while the chart's defaults were a large part of the cost: PostgreSQL capped at **150m CPU**, and a `netbox-worker` whose `wait-for-backend` init container hit `CrashLoopBackOff` and extended `helm --wait`. Image pulls were never the issue (netbox, 243 MB, pulled in 20.6s).

Comparable phases (provisioning + base example seeding), all measured on one 16-core host:

| fixture | provisioning | base seeding | total |
|---|---|---|---|
| kind + Helm, chart defaults (what CI runs today) | 460s | 433s | **893s** |
| kind + Helm, resource caps removed, worker off | 349s | 302s | 651s |
| **docker compose** | **241s** | **272s** | **~504s** |

Also measured: NetBox's expensive startup is one-time per fresh database — the image's entrypoint guards `migrate`/`trace_paths`/`reindex` behind `migrate --check`, so first boot took 339s and subsequent boots 50s and 60s.

## Changes

- **`nodeset: ubuntu-noble-large`** — the job ran a Kubernetes control plane, a four-container NetBox stack and the Ansible client on `SCS-2V-4-10` (2 vCPU / 4 GB). Node size alone was worth 1.53× on the deploy and 2.19× on seeding.
- **`tests/e2e/compose.yaml`** — postgres 17.10 / valkey 8.1 / netbox v4.5.10, with healthchecks and `depends_on: service_healthy`. NetBox is configured entirely through the environment variables its own baked `configuration.py` reads, so no config file is authored; the Helm chart had been *overriding* that file. `API_TOKEN_PEPPERS` is deliberately unset, because the image's bootstrap only ever creates a v2 token — which is why a v1 token still has to be minted for pynetbox / `netbox.netbox`.
- **`tests/e2e/deploy_netbox.sh`** — `docker compose up --wait` replaces cluster creation, chart install and readiness polling. No port-forward: the API is published on `127.0.0.1:8080` directly, removing that process and its liveness guard.
- **`tests/e2e/run.sh`** — keeps ownership of the lifecycle trap and teardown; diagnostics now capture the NetBox *application* log (which was unavailable when diagnosing the original failure — the cause had to be inferred from Kubernetes probe events).
- **`playbooks/pre-sonic-e2e.yml`** — kind, kubectl and Helm installs removed along with their six pinned digests. `ensure-docker` already provides `docker-compose-plugin`. The `accept_ra=2` sysctl is retained deliberately.

`timeout` stays at 2400.

## Verification

Full local run: `make sonic-e2e` exit 0, `OK: exports match the golden files`. **All 10 device configs byte-identical with no `--regenerate`** — same NetBox version in, same configs out. End to end 9m34s.

What this PR is for is the CI number: pre+run against the ~2100s baseline. The projection is ~1250–1350s, extrapolated from 16 cores onto 8 vCPU and therefore unverified.

## Relationship to #2483

This branch contains #2483's commits plus four scenario additions plus this migration. It is a draft opened to obtain CI timing; once green, the intent is to fold the result back rather than keep two overlapping PRs open.

## Not included

- `--parallel 4` seeding (measured 302s → 167s) — deferred deliberately, since it is the only change that could alter NetBox state non-deterministically, and keeping it separate makes a golden mismatch unambiguously attributable.
- netbox-manager is **untouched** and keeps its kind harness, so `netbox-manager-e2e` cannot regress.
- Removing the `accept_ra` workaround, and pinning postgres/valkey by digest.

## Known follow-ups

- The harness seeds every file in `tests/e2e/scenario/resources/`, tracked or not. Harmless in CI's clean checkout, but local WIP silently joins the fixture.
- `ubuntu-noble-large` has no `min-ready` (`ubuntu-noble` has 2), so this job now pays a cold VM boot per run — queue latency, outside the build timeout.


ideaship:
Superseded. The E2E suite was rebuilt on docker compose plus synthetic in-repo fixtures, and is being reopened as a sequence of smaller PRs; a tracking issue will collect them. Closing rather than retargeting because this branch's history carries the earlier `run.sh` name and the parallel-seeding commits that were deliberately dropped (concurrent seeding deadlocks on `dcim_device` row locks).

PR #2534, opened 2026-07-30
DO NOT MERGE: measure postgres fsync=off

ideaship:

Identical to the draft in

- osism/python-osism#2532

plus `postgres -c fsync=off -c synchronous_commit=off -c full_page_writes=off`.

**Question:** is any of the seeding time storage-bound?

Measured locally first: PGDATA is only ~83 MB and the database 28 MB, so it already sits entirely in shared buffers and the page cache. Reads never reach disk, so only the write path can be affected — which is why this uses postgres flags rather than moving PGDATA to a tmpfs. Same effect, no RAM cost, no OOM risk on the 4 GB flavor.

**Read the seeding phase, not the total.** The expected effect (order 10–30s of ~555s) is below the ~100s run-to-run variance already observed between nodes, so total pre+run cannot resolve it. Compare the netbox-manager `Runtime:` line against the 554.8s baseline in #2532. A null result here is a legitimate and useful answer.


ideaship:

**Result: no effect.** Base seeding 557.5s vs the 554.8s baseline in #2532 (+2.7s, i.e. noise). pre+run 1324s vs 1335s, also within variance.

Confirms the seeding path is **not storage-bound** — consistent with the local finding that PGDATA is only ~83 MB and already fully cached. Neither these flags nor a tmpfs is worth pursuing.

PR #2535, opened 2026-07-30
DO NOT MERGE: measure seeding with --parallel 4

ideaship:

Identical to the draft in

- osism/python-osism#2532

plus `--parallel 4` on the netbox-manager seeding calls.

**Question:** how much does parallel seeding save in CI, and does `ubuntu-noble-large` finally earn its keep?

Seeding is the largest remaining item — 555s of the 1335s pre+run, i.e. 42%. Locally `--parallel 4` cut it 302s → 167s (1.81×).

This doubles as a test of the node upgrade. Compose measured **1235s on 2 vCPU** vs **1335s on 8 vCPU** — no gain, because the critical path is single-threaded (sequential Django migrations, then one Ansible process issuing serial API calls). Parallel seeding is the first change that could actually use the extra cores, so if the nodeset upgrade is ever going to pay off, it is here.

Ordering is preserved: netbox-manager sorts resource files by leading number, runs groups sequentially, and parallelises only *within* a group. Static analysis of the 16 files in the `300-` group found 251 distinct objects touched with zero overlap. The golden comparison is itself the race detector — interleaved seeding would alter the generated configs, so a green run is meaningful evidence.


ideaship:

**Result: large win.** Base seeding **250.0s vs 554.8s** baseline = **2.22x**, better than the 1.81x measured locally. pre+run **1025s vs 1335s** (-310s). Goldens matched, so no race manifested.

The overlay pass was unchanged (49.8s vs 47.4s) exactly as predicted — its three files sit in distinct numeric groups and serialise regardless.

Worth adopting in #2532.

PR #2536, opened 2026-07-30
DO NOT MERGE: parallel seeding on 2 vCPU

ideaship:

Identical to

- osism/python-osism#2532

except `nodeset: ubuntu-noble` (2 vCPU / 4 GB) instead of `ubuntu-noble-large` (8 vCPU / 32 GB).

**Question:** does the larger flavor earn its keep now that seeding is parallel?

Serially it did not — compose measured **1235s** pre+run on 2 vCPU vs **1335s** on 8 vCPU, no gain, because the critical path was single-threaded. `--parallel 4` is the first change that can use extra cores, and on 8 vCPU it gave **2.22×** on the base seeding pass (555s → 250s). Four workers on 2 vCPU must fare worse; how much worse *is* the value of the bigger node.

Reference points for the base example seeding pass:

| config | base seeding |
|---|---|
| 2 vCPU, serial | 525s |
| 8 vCPU, serial | 555s |
| 8 vCPU, `--parallel 4` | 250s |
| **2 vCPU, `--parallel 4`** | **this run** |

If it lands near 250s, the nodeset commit can be dropped and the quota saved. If it lands near 400s+, the upgrade is doing real work.

Note this also exercises the `SEED_PARALLEL` escape hatch plumbing on the constrained flavor, where contention is most likely to bite.


ideaship:

**Result: the larger flavor buys ~45s.**

| config | base seeding |
|---|---|
| 2 vCPU, serial | 525s |
| 8 vCPU, serial | 555s |
| 8 vCPU, `--parallel 4` | 250.0s / 242.6s |
| **2 vCPU, `--parallel 4`** | **289.9s** |

So parallel seeding works on 2 vCPU too, ~47s slower than on 8 vCPU. The seeding phase measures repeatably (±7s across runs), so that difference is real signal rather than noise.

pre+run was 1025s (2 vCPU) vs 922s (8 vCPU), but roughly half that gap is `compose up`, which ranged 354–432s across six runs independent of node size — i.e. noise.

No memory pressure or OOM on the small node, which was the main concern with four concurrent workers on 4 GB.

Conclusion: ~45s (~4.5%) does not obviously justify 4× the CI quota plus the cold-boot from no `min-ready`. Dropping the nodeset commit from #2532 looks right.

Issue #2400, opened 2026-06-19
Meta: Extend Celery/Redis integration tests (Tier 1 + Tier 2)


berendt:
## Context

#2368 laid the foundation for integration tests: the `python-osism-integration-tests` Zuul job (`playbooks/test-integration.yml`) starts a single `redis:7-alpine` container, installs the package with pipenv, and runs `pytest tests/integration` with `REDIS_HOST=localhost` and `OSISM_REQUIRE_REDIS=1`. The shared infrastructure — skip-when-Redis-unreachable logic plus the `celery_app` / `celery_worker` session fixtures — lives in `tests/integration/conftest.py`.

Three test modules exist so far:

- `test_celery.py` — noop round-trip + worker visibility
- `test_locking.py` — Redlock + task-lock round-trip
- `test_redis_streams.py` — task-output push/finish/fetch

This meta issue tracks the next round of integration tests. The guiding principle: first exhaust everything that works with the **existing single-Redis infrastructure** (Tier 1), then add the few tests that need only one extra test dependency (Tier 2 — `httpx` for the FastAPI `TestClient`). Tests that require additional services (RabbitMQ, NetBox, OpenStack) are explicitly **out of scope** here and will be tracked separately.

All new tests live under `tests/integration/`, carry `pytestmark = pytest.mark.integration`, stay green locally (skipped without Redis), and run for real in CI.

## Tier 1 — no infrastructure change (Redis only)

- [ ] #2401 — RedisSemaphore / create_netbox_semaphore
- [ ] #2402 — Ansible vault password round-trip
- [ ] #2403 — check_ansible_facts freshness
- [ ] #2404 — Celery error / state / revoke paths
- [ ] #2405 — EventBridge Redis pub/sub round-trip

## Tier 2 — one extra test dependency (`httpx` for FastAPI TestClient)

Shared prerequisite: add `httpx` to `[dev-packages]` in `Pipfile` so `fastapi.testclient.TestClient` is importable. `fastapi` itself is already a runtime dependency.

- [ ] #2406 — API facts endpoints via FastAPI TestClient
- [ ] #2407 — WebSocket endpoint `/v1/events/openstack` via TestClient

PR #2547, opened 2026-08-03
Add integration test for the Ansible Vault password round trip

### LINKED ISSUE #2402: Integration test: Ansible vault password round-trip

## Description

This issue adds `tests/integration/test_vault.py`, which exercises the Ansible vault password round trip (Fernet encryption in, Redis storage, Fernet decryption out) against the live Redis of the integration CI job. It also makes the one production change the test needs to be self-contained: the keyfile path inside `osism.utils.get_ansible_vault_password()` becomes a module constant that a test can monkeypatch.

### 1. Integration-test lane

Already exists: the `python-osism-integration-tests` Zuul job (`playbooks/test-integration.yml`) starts a `redis:7-alpine` container and runs `pytest tests/integration` with `REDIS_HOST=localhost` and `OSISM_REQUIRE_REDIS=1`. Without a reachable Redis, `pytest_collection_modifyitems` in `tests/integration/conftest.py:70` skips every `integration`-marked test. The marker is registered under `--strict-markers` in `setup.cfg:184-187`.

This issue adds one test module to that lane. No change to `tests/integration/conftest.py`, `playbooks/test-integration.yml`, `Pipfile`, or `setup.cfg`. The new tests use neither the `celery_app` nor the `celery_worker` fixture. `cryptography` is not a direct dependency but is installed transitively (production code already imports it the same way, `osism/commands/vault.py:25`), so the test may import `Fernet` and `InvalidToken` from `cryptography.fernet` at module level.

### 2. The read path, and the one production change

Already exists: `get_ansible_vault_password()` (`osism/utils/__init__.py:351`) opens the keyfile, builds a `Fernet`, and reads the `ansible_vault_password` key through `_init_redis()` (line 362). It raises `ValueError("Ansible vault password is not set in Redis")` when the key is absent (line 364), decrypts the token (line 366), raises `ValueError("Ansible vault password is empty or contains only whitespace")` when the plaintext is blank (lines 369-372), and logs "Unable to get ansible vault password" before re-raising any exception (lines 375-377). The keyfile path is a local variable, `keyfile = "/share/ansible_vault_password.key"` (line 354). The issue text says "monkeypatch the module constant", but no such constant exists, and a local variable cannot be monkeypatched.

This issue adds a module-level constant `ANSIBLE_VAULT_KEYFILE = "/share/ansible_vault_password.key"` in `osism/utils/__init__.py`, placed with the existing module-level definitions near `LOCAL_FACT_HOSTS` (`osism/utils/__init__.py:25`). `get_ansible_vault_password()` drops its local variable and opens `ANSIBLE_VAULT_KEYFILE` directly, reading the module global at call time so that `monkeypatch.setattr(utils, "ANSIBLE_VAULT_KEYFILE", str(path))` redirects it. The default path stays byte-identical; there is no behavior change.

### 3. The commands

Already exist: `SetPassword` (`osism/commands/vault.py:17`) carries the class attribute `keyfile` (line 18), generates and writes a Fernet key when the file is missing (lines 28-34), reads the password from piped stdin when `sys.stdin.isatty()` is false (lines 39-40), and stores it Fernet-encrypted under `ansible_vault_password` (lines 46-48). `UnsetPassword` deletes that key (line 57). `Check` (`osism/commands/vault.py:127`, own `keyfile` at line 134) verifies the chain: keyfile present, valid Fernet key, key set in Redis (line 210), token decryptable (line 229), plaintext non-empty (line 247); with `--format script` it prints a single `PASSED` or `FAILED: <reason>` line. The CLI entry points are `osism set vault password` / `osism vault password set`, `osism vault password unset` / `osism noset vault password`, and `osism vault check` (`setup.cfg:126`, `:137`, `:171`, `:173-174`). Unit tests in `tests/unit/commands/test_vault.py` already cover all of these against a mocked Redis, including the instantiation pattern (`vault.SetPassword(MagicMock(), MagicMock())`, then `cmd.get_parser("test").parse_args([...])`) and the piped-stdin stub `_piped_stdin` (`tests/unit/commands/test_vault.py:205`).

This issue adds an in-process round trip of `SetPassword`, `Check`, and `UnsetPassword` against the real Redis, reusing those instantiation patterns with real crypto and a real store instead of mocks. The commands and `get_ansible_vault_password()` share one Redis client: `utils.redis` is a lazily initialized module attribute (`osism/utils/__init__.py:700-713`) backed by the same `_init_redis()` singleton (lines 28-40) and configured by `osism/settings.py:21-23`.

### 4. The test module

New file `tests/integration/test_vault.py`, with the SPDX header, a module docstring in the style of `tests/integration/test_redis_streams.py`, and `pytestmark = pytest.mark.integration`. Two fixtures:

- `vault_keyfile(monkeypatch, tmp_path)` picks `tmp_path / "ansible_vault_password.key"` and monkeypatches `utils.ANSIBLE_VAULT_KEYFILE`, `vault.SetPassword.keyfile`, and `vault.Check.keyfile` to that path. It does not create the file; tests that need a pre-existing key write one from `Fernet.generate_key()` themselves, and the command round trip leaves it absent so `SetPassword` generates it for real.
- `clean_vault_key` (autouse) runs `utils.redis.delete("ansible_vault_password")` before and after each test. The key name is fixed and the Redis database is shared by the whole integration session, so cleanup must not depend on the test passing.

Six tests:

- `test_round_trip_returns_plaintext`: write a Fernet key to the keyfile, store `Fernet(key).encrypt(b"hunter2")` under `ansible_vault_password` via `utils.redis.set`, and assert `utils.get_ansible_vault_password()` returns `"hunter2"`.
- `test_missing_redis_key_raises_value_error`: keyfile present, Redis key absent; `pytest.raises(ValueError, match="not set in Redis")`.
- `test_blank_password_raises_value_error`: parametrized over `""` and `"  \t "`; store the encrypted value and expect `pytest.raises(ValueError, match="empty or contains only whitespace")`.
- `test_token_from_other_key_raises_invalid_token`: store a token encrypted with a second, different Fernet key; `pytest.raises(InvalidToken)`. The generic `except` at `osism/utils/__init__.py:375` logs and re-raises, so `InvalidToken` reaches the caller unchanged.
- `test_missing_keyfile_raises_file_not_found`: leave the keyfile uncreated; `get_ansible_vault_password()` raises `FileNotFoundError` from the `open` call, before Redis is touched.
- `test_command_round_trip`: monkeypatch `sys.stdin` with a stub whose `isatty()` returns `False` and whose `read()` returns `"hunter2\n"`; run `SetPassword.take_action`, which generates the keyfile at the monkeypatched path and stores the encrypted password; assert `utils.get_ansible_vault_password()` returns `"hunter2"` (the command strips the trailing newline); run `Check.take_action` parsed with `["--format", "script"]` and `Check._find_secrets_file` monkeypatched to return `None`, assert it returns 0 and prints `PASSED` (via `capsys`); run `UnsetPassword.take_action`, assert `utils.redis.get("ansible_vault_password")` is `None` and a final `get_ansible_vault_password()` call raises `ValueError` matching "not set in Redis".

## Motivation

`get_ansible_vault_password()` is the seam through which conductor workers obtain the vault secret: `get_vault()` (`osism/tasks/conductor/utils.py:72`) wraps it in a `VaultLib` to decrypt secrets from NetBox data, and silently falls back to an empty `VaultLib` when it raises. The function is glue between Fernet and Redis, and nothing currently executes it against a real store: `tests/unit/commands/test_vault.py` mocks Redis throughout, and `get_ansible_vault_password()` itself has no test at all. Whether the bytes `SetPassword` writes actually come back through a live client and decrypt to the same string is exactly the contract a mock cannot check.

#2368 built the lane for this: a real Redis in CI, automatic skips without one. Meta issue #2400 lists the vault password round trip as Tier 1, meaning it needs only that single Redis, no extra service, and no new dependency. This issue closes that item with one test module and a small testability hoist in `osism/utils/__init__.py`.

## Affected Areas

- `osism/utils/__init__.py` (hoist the hardcoded keyfile path out of `get_ansible_vault_password()` into the new module constant `ANSIBLE_VAULT_KEYFILE`; no behavior change)
- `tests/integration/test_vault.py` (new: `vault_keyfile` and `clean_vault_key` fixtures plus six tests covering the direct round trip, four error paths, and the command round trip)

## Acceptance Criteria

- [ ] Define `ANSIBLE_VAULT_KEYFILE = "/share/ansible_vault_password.key"` at module level in `osism/utils/__init__.py` and make `get_ansible_vault_password()` open that constant instead of its local `keyfile` variable, leaving the default path byte-identical.
- [ ] Add `tests/integration/test_vault.py` with the SPDX header, a module docstring, `pytestmark = pytest.mark.integration`, and the two fixtures described above.
- [ ] Verify the direct round trip: a password encrypted with the keyfile's Fernet key and stored under `ansible_vault_password` is returned as plaintext by `utils.get_ansible_vault_password()`.
- [ ] Verify the absent-key path: with no `ansible_vault_password` key in Redis, `get_ansible_vault_password()` raises `ValueError` matching "not set in Redis".
- [ ] Verify the blank-password path for both `""` and a whitespace-only string: the stored token decrypts, and `get_ansible_vault_password()` raises `ValueError` matching "empty or contains only whitespace".
- [ ] Verify the upstream decrypt error: a token encrypted with a different Fernet key makes `get_ansible_vault_password()` raise `cryptography.fernet.InvalidToken`.
- [ ] Verify the missing-keyfile path: with `ANSIBLE_VAULT_KEYFILE` pointing at a nonexistent file, `get_ansible_vault_password()` raises `FileNotFoundError`.
- [ ] Verify the command round trip in-process: `SetPassword.take_action` with a piped-stdin stub generates the keyfile at the monkeypatched path and stores the encrypted password; `Check.take_action` with `--format script` and `_find_secrets_file` returning `None` returns 0 and prints `PASSED`; `UnsetPassword.take_action` removes the key so `utils.redis.get("ansible_vault_password")` returns `None` and `get_ansible_vault_password()` raises `ValueError` matching "not set in Redis".
- [ ] Delete the `ansible_vault_password` Redis key before and after every test through the autouse fixture, so a failing test leaves no state behind for the other integration tests sharing the Redis database.
- [ ] Confirm the `python-osism-integration-tests` job runs the new module and passes with no change to `tests/integration/conftest.py`, `playbooks/test-integration.yml`, `Pipfile`, or `setup.cfg`.
- [ ] Confirm the unit-test job stays green after the constant hoist, in particular `tests/unit/commands/test_vault.py`, which patches the vault command classes rather than the new constant.

## Non-Goals

- Driving the CLI as a subprocess (`osism set vault password`): a child process offers no way to redirect the keyfile path (there is no environment-variable override), so the commands run in-process, as `tests/unit/commands/test_vault.py` already does.
- Testing the `View` command (`osism/commands/vault.py:60`): it shows a vault-encrypted file via `/usr/local/bin/ansible-vault` and never reads the `ansible_vault_password` Redis key. The issue's "view" citation (`osism/commands/vault.py:210`) is the Redis `get` inside `Check`, which this plan covers instead.
- Exercising Check's secrets-file decryption step (`osism/commands/vault.py:260-312`): it needs ansible-core, which the integration environment does not install (`tests/conftest.py` stubs the `ansible` package when it is missing), and its `/opt/configuration` search would tie the test to the host filesystem. `_find_secrets_file` is stubbed to return `None`.
- Rewiring `SetPassword.keyfile` and `Check.keyfile` (`osism/commands/vault.py:18`, `:134`) to the new constant: both are already monkeypatchable class attributes that the unit tests patch directly, and folding them into the constant would change command classes this issue does not need to touch.
- The remaining Tier-1 paths belong to the open siblings: RedisSemaphore to #2401, facts freshness to #2403, Celery error, state, and revoke paths to #2404, EventBridge pub/sub to #2405.
- Tests needing RabbitMQ, NetBox, or OpenStack stay out of scope for the whole round, per #2400.

## References

- #2400 (Meta issue, Tier 1 list), #2368 (integration-test foundation)
- `tests/integration/conftest.py` (skip logic; this module uses no session fixture)
- `tests/integration/test_redis_streams.py`, `tests/integration/test_api_facts.py` (module style and seed-and-clean fixture pattern; the latter landed via #2411)
- `tests/unit/commands/test_vault.py` (command instantiation, keyfile monkeypatching, `_piped_stdin` stub)
- `osism/commands/vault.py` (the three commands), `osism/utils/__init__.py:351-377` (the function under test), `osism/tasks/conductor/utils.py:72-90` (its consumer)
- `playbooks/test-integration.yml` (CI job), `setup.cfg:184-187` (marker registration)

**Executability score:** 8/10

---

berendt:
## What

`get_ansible_vault_password()` (`osism/utils/__init__.py:355`) glues Fernet to
Redis: the key comes from a keyfile, the token from Redis. Conductor workers
read the vault secret through it (`osism/tasks/conductor/utils.py:75`). Every
existing test mocks one side of that chain, so nothing checks that the bytes
`osism set vault password` writes come back through a live client and decrypt
to the same string.

This adds one module to the integration lane that runs the chain against the
live Redis of the `python-osism-integration-tests` job, plus the two changes
around it that make running it safe: one source for the keyfile path, and an
integration suite that cannot be pointed at a deployment's Redis database.

Part of #2400. Follow-up to #2368.

## Changes (in commit order)

1. **Read the Ansible Vault keyfile path from the settings module**: the path
   existed three times — as a local in `get_ansible_vault_password()` that no
   test can redirect, and as a literal in `SetPassword` and `Check`. It becomes
   `settings.ANSIBLE_VAULT_KEYFILE`, overridable through the environment
   variable of the same name, and every consumer reads it at call time. The
   `keyfile` class attributes are dropped: nothing sets them, cliff builds a
   fresh command instance per invocation, and a class body is evaluated once at
   import — so reading a constant there would copy the value rather than
   reference it, leaving three independent variables that merely happen to be
   equal. They must not be able to drift: `Check` exists to verify the chain
   `get_ansible_vault_password()` walks and would print `PASSED` off a different
   variable than the code it checks, and `SetPassword` writes a new Fernet key
   when the file is absent, so a path disagreement there is not an error but a
   fork of the key material. The default path is unchanged.

2. **Refuse to run the integration tests against Redis database 0**: the suite
   gated only on whether Redis answers a ping, and `REDIS_DB` defaults to `0`.
   Some of the keys it touches are fixed global names rather than per-run
   UUIDs, so `pytest tests/integration` run on, or pointed at, a manager node
   mutates that deployment's state — and CI stays green either way, because it
   provisions a fresh Redis. The job now sets `REDIS_DB=15`, and
   `tests/integration/conftest.py` exits the session when the target is
   database 0 without `OSISM_ALLOW_DEFAULT_REDIS_DB` set. `REDIS_DB` also feeds
   `Config.broker_url` (`osism/tasks/__init__.py:28`), so the one variable moves
   the direct client, the Celery broker and the result backend together. It
   cannot be set from inside the conftest — `osism/settings.py` reads the
   environment at import and the conftest imports `settings`. The reachability
   check still runs first, so a local run without Redis keeps skipping instead
   of failing. README updated.

3. **Add integration tests for the Ansible Vault password round trip**: new
   module `tests/integration/test_vault.py` with four tests.
   - the direct round trip: a password encrypted with the keyfile's Fernet key
     and stored under `ansible_vault_password` comes back as plaintext;
   - no Redis key: `ValueError` matching "not set in Redis";
   - token written under a different Fernet key: `InvalidToken` — the real
     exception type, which the mocked unit test cannot pin;
   - the command round trip in-process: `SetPassword` with a piped-stdin stub
     generates the keyfile and stores the token, `Check --format script`
     returns 0 and prints `PASSED`, `UnsetPassword` removes the key again.

   The branches that follow decryption (blank password, missing keyfile) are
   pure Python and already covered by `tests/unit/utils/test_init_task_output.py`,
   so they stay there and the Redis-gated job keeps to what needs Redis.

No new dependency: `cryptography` is already imported the same way by
`osism/commands/vault.py:25`.

## Notes

- `Check`'s last step decrypts a `secrets.yml` through ansible's `VaultLib`.
  ansible-core is not installed in this environment — it lives in the optional
  `[ansible]` extra (`setup.cfg:34`) while the job installs `--dev` plus
  `pip install .` — so `tests/conftest.py` stubs it, and the stubbed `decrypt()`
  returns `b""` without raising. Pointing `--path` at an encrypted fixture would
  report success for any password, so the step is skipped rather than mocked.
- The commands run in-process rather than as a subprocess so the keyfile fixture
  applies to them.
- #2402 listed rewiring `SetPassword.keyfile` and `Check.keyfile` as a Non-Goal.
  They are rewired here anyway, for the reasons in change 1.

## Test plan

- `black --check` and `flake8` report all changed files clean.
- The unit and integration jobs cover the rest: `tests/unit/commands/test_vault.py`
  and `tests/unit/utils/test_init_task_output.py` for the settings move, the
  `python-osism-integration-tests` job for the new module. The integration job
  is also what exercises the `REDIS_DB=15` switch — a green run there is the
  check that nothing in the suite depended on database 0.
- To run the integration module by hand:
  `REDIS_HOST=localhost REDIS_DB=15 OSISM_REQUIRE_REDIS=1 pytest tests/integration/test_vault.py`
  against a throwaway `redis:7-alpine` container.

Closes #2402

---

_Implemented by [planwerk-agent](https://github.com/planwerk/planwerk-agent) with Claude:claude-opus-5[1m]_

### DIFF
--- a/osism/utils/__init__.py
+++ b/osism/utils/__init__.py
@@ -351,10 +351,8 @@ def get_openstack_connection():
 def get_ansible_vault_password():
     from cryptography.fernet import Fernet
 
-    keyfile = "/share/ansible_vault_password.key"
-
     try:
-        with open(keyfile, "r") as fp:
+        with open(settings.ANSIBLE_VAULT_KEYFILE, "r") as fp:
             key = fp.read()
         f = Fernet(key)
 
--- a/tests/integration/test_vault.py
+++ b/tests/integration/test_vault.py
@@ -1,0 +1,125 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Ansible vault password round-trip integration tests.
+
+``get_ansible_vault_password`` is the seam through which conductor workers get
+the vault secret (``osism.tasks.conductor.utils.get_vault``). It glues Fernet to
+Redis: the key comes from a keyfile, the token from Redis. These tests run that
+chain against the live Redis, together with the ``osism vault`` commands that
+write and remove the token.
+"""
+
+import sys
+from unittest.mock import MagicMock
+
+import pytest
+from cryptography.fernet import Fernet, InvalidToken
+
+from osism import settings, utils
+from osism.commands import vault
+
+pytestmark = pytest.mark.integration
+
+VAULT_KEY = "ansible_vault_password"
+
+
+@pytest.fixture
+def vault_keyfile(monkeypatch, tmp_path):
+    """Redirect the keyfile path to ``tmp_path`` and return it.
+
+    The file itself is not created: tests that need a pre-existing Fernet key
+    write one themselves, and the command round trip leaves it absent so
+    ``SetPassword`` generates it for real.
+    """
+    keyfile = tmp_path / "ansible_vault_password.key"
+    monkeypatch.setattr(settings, "ANSIBLE_VAULT_KEYFILE", str(keyfile))
+    return keyfile
+
+
+@pytest.fixture(autouse=True)
+def clean_vault_key():
+    """Delete the vault key before and after each test.
+
+    The key name is part of the contract under test, so it cannot be
+    namespaced per run the way the sibling modules namespace theirs, and the
+    Redis database is shared by the whole integration session -- the cleanup
+    must not depend on the test passing. Deleting a name a real deployment
+    also owns is safe only because ``conftest.py`` refuses to run against the
+    default database.
+    """
+    utils.redis.delete(VAULT_KEY)
+    yield
+    utils.redis.delete(VAULT_KEY)
+
+
+def _write_key(keyfile):
+    """Write a fresh Fernet key to ``keyfile`` and return it."""
+    key = Fernet.generate_key()
+    keyfile.write_text(key.decode("utf-8"))
+    return key
+
+
+def _piped_stdin(text):
+    stdin = MagicMock()
+    stdin.isatty.return_value = False
+    stdin.read.return_value = text
+    return stdin
+
+
+def test_round_trip_returns_plaintext(vault_keyfile):
+    """A password encrypted with the keyfile's key is read back as plaintext."""
+    key = _write_key(vault_keyfile)
+    utils.redis.set(VAULT_KEY, Fernet(key).encrypt(b"hunter2"))
+
+    assert utils.get_ansible_vault_password() == "hunter2"
+
+
+def test_missing_redis_key_raises_value_error(vault_keyfile):
+    """Without the Redis key the read path raises ``ValueError``."""
+    _write_key(vault_keyfile)
+
+    with pytest.raises(ValueError, match="not set in Redis"):
+        utils.get_ansible_vault_password()
+
+
+def test_token_from_other_key_raises_invalid_token(vault_keyfile):
+    """A token written under a different Fernet key raises ``InvalidToken``."""
+    _write_key(vault_keyfile)
+    utils.redis.set(VAULT_KEY, Fernet(Fernet.generate_key()).encrypt(b"hunter2"))
+
+    with pytest.raises(InvalidToken):
+        utils.get_ansible_vault_password()
+
+
+def test_command_round_trip(vault_keyfile, monkeypatch, capsys):
+    """``set`` writes what ``check`` accepts and ``unset`` removes again."""
+    set_password = vault.SetPassword(MagicMock(), MagicMock())
+    monkeypatch.setattr(sys, "stdin", _piped_stdin("hunter2\n"))
+    set_password.take_action(set_password.get_parser("test").parse_args([]))
+
+    assert vault_keyfile.exists()
+    assert utils.get_ansible_vault_password() == "hunter2"
+
+    check = vault.Check(MagicMock(), MagicMock())
+    # The last check step decrypts a secrets.yml with ansible's VaultLib.
+    # ansible-core is not installed in this environment (it lives in the
+    # optional [ansible] extra), so tests/conftest.py stubs VaultLib and its
+    # decrypt() is a no-op that would report success for any password -- the
+    # step is skipped rather than mocked, since pointing --path at an
+    # encrypted fixture would only look like coverage. Worth covering for real
+    # if ansible-core is ever added to the test dependencies.
+    monkeypatch.setattr(vault.Check, "_find_secrets_file", lambda self: None)
+    rc = check.take_action(check.get_parser("test").parse_args(["--format", "script"]))
+
+    assert rc == 0
+    # ``--format script`` prints one token, ``PASSED`` or ``FAILED: <reason>``.
+    # Compared after stripping, so the surrounding whitespace stays free to
+    # change while the contract itself is still asserted.
+    assert capsys.readouterr().out.strip() == "PASSED"
+
+    unset = vault.UnsetPassword(MagicMock(), MagicMock())
+    unset.take_action(unset.get_parser("test").parse_args([]))
+
+    assert utils.redis.get(VAULT_KEY) is None
+    with pytest.raises(ValueError, match="not set in Redis"):
+        utils.get_ansible_vault_password()
--- a/README.md
+++ b/README.md
@@ -29,12 +29,18 @@ same virtualenv; they are skipped automatically when Redis is not running.
 
 ```
 docker run -d -p 6379:6379 redis:7-alpine
-REDIS_HOST=localhost pipenv run pytest tests/integration
+REDIS_HOST=localhost REDIS_DB=15 pipenv run pytest tests/integration
 ```
 
 > **Warning:** The suite mutates live state on the configured Redis. Most keys
-> are per-run (UUID-based), but the task-lock test reads, writes and removes the
-> fixed global key `osism:task_lock` on the selected `REDIS_DB`. Always point it
-> at a disposable Redis (such as the throwaway container above); running it
-> against a Redis shared with a real OSISM deployment would clobber an active
-> operator lock.
+> are per-run (UUID-based), but some are fixed global names on the selected
+> `REDIS_DB`: the task-lock test reads, writes and removes `osism:task_lock`,
+> and the vault test removes `ansible_vault_password` — which a real deployment
+> cannot recover, since no other copy of it exists on the system. Always point
+> the suite at a disposable Redis (such as the throwaway container above).
+>
+> To make that hard to get wrong, the suite refuses to run against `REDIS_DB=0`,
+> the database a deployment uses. Set `REDIS_DB` to a spare database as shown
+> above, or `OSISM_ALLOW_DEFAULT_REDIS_DB=1` if the Redis itself is disposable.
+> `REDIS_DB` moves the direct client, the Celery broker and the result backend
+> together.
--- a/osism/commands/vault.py
+++ b/osism/commands/vault.py
@@ -11,12 +11,10 @@ import sys
 from cliff.command import Command
 from loguru import logger
 
-from osism import utils
+from osism import settings, utils
 
 
 class SetPassword(Command):
-    keyfile = "/share/ansible_vault_password.key"
-
     def get_parser(self, prog_name):
         parser = super(SetPassword, self).get_parser(prog_name)
         return parser
@@ -25,12 +23,14 @@ class SetPassword(Command):
         from cryptography.fernet import Fernet
         from prompt_toolkit import prompt
 
-        if os.path.isfile(self.keyfile):
-            with open(self.keyfile, "r") as fp:
+        keyfile = settings.ANSIBLE_VAULT_KEYFILE
+
+        if os.path.isfile(keyfile):
+            with open(keyfile, "r") as fp:
                 key = fp.read()
         else:
             key = Fernet.generate_key()
-            with open(self.keyfile, "w+") as fp:
+            with open(keyfile, "w+") as fp:
                 fp.write(key.decode("utf-8"))
 
         f = Fernet(key)
@@ -131,8 +131,6 @@ class Check(Command):
     and optionally tests decryption against a secrets.yml file.
     """
 
-    keyfile = "/share/ansible_vault_password.key"
-
     def get_parser(self, prog_name):
         parser = super(Check, self).get_parser(prog_name)
         parser.add_argument(
@@ -174,15 +172,16 @@ class Check(Command):
         format = parsed_args.format
         passed = 0
         failed = 0
+        keyfile = settings.ANSIBLE_VAULT_KEYFILE
 
         # Step 1: Check keyfile exists
-        if os.path.isfile(self.keyfile):
+        if os.path.isfile(keyfile):
             if format == "log":
-                logger.info(f"Keyfile exists: {self.keyfile}")
+                logger.info(f"Keyfile exists: {keyfile}")
             passed += 1
         else:
             if format == "log":
-                logger.error(f"Keyfile not found: {self.keyfile}")
+                logger.error(f"Keyfile not found: {keyfile}")
             elif format == "script":
                 print("FAILED: keyfile_missing")
             failed += 1
@@ -191,7 +190,7 @@ class Check(Command):
 
         # Step 2: Check keyfile contains a valid Fernet key
         try:
-            with open(self.keyfile, "r") as fp:
+            with open(keyfile, "r") as fp:
                 key = fp.read()
             f = Fernet(key)
             if format == "log":
--- a/osism/settings.py
+++ b/osism/settings.py
@@ -22,6 +22,14 @@ REDIS_HOST: str = os.getenv("REDIS_HOST", "redis")
 REDIS_PORT: int = int(os.getenv("REDIS_PORT", "6379"))
 REDIS_DB: int = int(os.getenv("REDIS_DB", "0"))
 
+# Keyfile holding the Fernet key that encrypts the Ansible Vault password in
+# Redis. Read through this module by every consumer -- 'osism set vault
+# password' creates the file when it is missing, so a reader disagreeing with
+# the writer about the path would silently fork the key material.
+ANSIBLE_VAULT_KEYFILE = os.getenv(
+    "ANSIBLE_VAULT_KEYFILE", "/share/ansible_vault_password.key"
+)
+
 
 NETBOX_URL = os.getenv("NETBOX_API", os.getenv("NETBOX_URL"))
 NETBOX_TOKEN = str(
--- a/playbooks/test-integration.yml
+++ b/playbooks/test-integration.yml
@@ -59,6 +59,13 @@
           {{ python_venv_dir }}/bin/pipenv run pytest tests/integration
       environment:
         REDIS_HOST: localhost
+        # The tests write and delete keys under names a real deployment also
+        # uses (the vault password is stored under a fixed key), so they must
+        # never run against the database a manager node keeps its state in.
+        # REDIS_DB also feeds the Celery broker and result backend
+        # (osism/tasks/__init__.py), so this moves every client at once.
+        # tests/integration/conftest.py refuses to run against database 0.
+        REDIS_DB: "15"
         # Fail the session if Redis is unreachable instead of skipping every
         # test, so a Redis-startup failure turns the job red rather than green.
         OSISM_REQUIRE_REDIS: "1"
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -31,6 +31,22 @@ WORKER_NODE_PREFIX = WORKER_NAME.split("@", 1)[0] + "@"
 WORKER_BOOT_TIMEOUT = 60
 
 
+# Database a deployment uses by default (``osism/settings.py``). The
+# integration tests delete keys under names a real deployment also uses, so
+# running them here would destroy production state -- see _reject_default_db.
+DEFAULT_REDIS_DB = 0
+
+
+def _default_db_allowed():
+    """Return ``True`` when running against Redis database 0 was opted into."""
+    return os.environ.get("OSISM_ALLOW_DEFAULT_REDIS_DB", "").lower() not in (
+        "",
+        "0",
+        "false",
+        "no",
+    )
+
+
 def _require_redis():
     """Return ``True`` when an unreachable Redis must fail the session.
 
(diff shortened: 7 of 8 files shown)


tests/integration/test_vault.py
```diff
@@ -0,0 +1,140 @@
...
+
+
+@pytest.fixture(autouse=True)
+def clean_vault_key():
+    """Delete the vault key before and after each test.
+
+    The key name is fixed and the Redis database is shared by the whole
+    integration session, so the cleanup must not depend on the test passing.
+    """
+    utils.redis.delete(VAULT_KEY)
```
ideaship, line 50: This deletes a key a real deployment owns, and nothing in the suite distinguishes a disposable Redis from a live one. `tests/integration/conftest.py:50-67` gates only on reachability, `REDIS_DB` defaults to `0` (`osism/settings.py:23`), and `playbooks/test-integration.yml` overrides only `REDIS_HOST` — so `pytest tests/integration` run on, or pointed at, a manager node deletes that deployment's vault password. It's unrecoverable: no other copy exists on the system, and `osism/tasks/conductor/utils.py` starts raising for every conductor task until someone who remembers the plaintext re-runs `osism set vault password`. CI is safe because it provisions a fresh Redis, which is exactly why this won't show up as a red job.

The sibling modules avoid this by namespacing (`itest-<uuid>` in `test_locking.py:20`, `test_redis_streams.py:21`, `test_api_facts.py:56`), but that doesn't apply here — the key name is part of the contract under test. Please isolate the database instead. Save-and-restore isn't enough: a `SIGINT` between capture and teardown still loses the value.

Suggested: add `REDIS_DB: "15"` to the `environment:` block at `playbooks/test-integration.yml:60`, plus a guard in `tests/integration/conftest.py` that hard-fails when `settings.REDIS_DB == 0` without an explicit opt-in. That protects every future integration test, not just this module. `REDIS_DB` also feeds `Config.broker_url` (`osism/tasks/__init__.py:28`), so one variable moves the direct client, the Celery broker and the result backend together.

Note it can't be set from *inside* `tests/integration/conftest.py` — `osism/settings.py:23` reads the env var at import and the conftest imports `settings` at line 18. If you'd rather keep the fix module-local, patch both `utils._redis` and `utils.redis` (the latter is cached into module globals by the `__getattr__` at `osism/utils/__init__.py:703-705`, so patching `_redis` alone leaves the command path on the original client).

osism/utils/__init__.py
```diff
@@ -24,6 +24,10 @@
 # from the freshness check avoids permanent, unactionable stale warnings.
 LOCAL_FACT_HOSTS = frozenset({"localhost", "127.0.0.1", "::1"})
 
+# Keyfile holding the Fernet key that encrypts the Ansible Vault password in
+# Redis. A module constant rather than a local, so tests can redirect it.
```
ideaship: "so tests can redirect it" holds for one of the three consumers. `get_ansible_vault_password` reads this global at call time (`:359`), but `SetPassword` and `Check` evaluate it in their class bodies (`osism/commands/vault.py:18,134`), so after import they hold their own copies. `utils.ANSIBLE_VAULT_KEYFILE` and `SetPassword.keyfile` are two independent variables that happen to be equal — there's no runtime relationship between them. This is the drift risk raised earlier in the PR, and it's worth noting that the suggestion there — use the constant in the command classes — is what this diff already does; reading it in a class body copies the value rather than referencing it.

Worth fixing structurally rather than documenting, for two reasons that outlive this PR:

**`Check` exists to verify the worker path.** Its docstring promises to verify "the full chain: keyfile, Redis storage, decryption" — the chain being `get_ansible_vault_password`, which is what conductor tasks call. If the two bindings ever differ, `osism vault check` prints `PASSED` while every conductor task raises `InvalidToken`. A health check that reads a different variable than the code it checks can't do its job.

**`SetPassword` turns a path disagreement into forked key material.** `osism/commands/vault.py:28-34` generates and writes a new Fernet key when the file is absent, so a wrong path isn't an error — it writes key B to the wrong location and stores a token under B in Redis, while `/share/ansible_vault_password.key` still holds key A. The workers then fail until someone with the plaintext re-runs `osism set vault password`. A create-if-missing operation must not disagree with its readers about the path.

Neither is a live bug — all three copies currently hold the same literal and nothing writes them outside tests. What makes it worth doing now is the obvious next change to this code: making the path configurable. This constant sits in `osism/utils/__init__.py` rather than `osism/settings.py`, where every other tunable in the repo lives. Someone moving it there and switching `get_ansible_vault_password` over would leave both commands hardcoded to the old default — a half-applied config with the fork above as its payload. Compare `osism/services/event_bridge.py:52`, which reads `os.getenv("REDIS_DB", "0")` instead of `settings.REDIS_DB`; the same shape has already happened once here.

Suggested shape — one late-bound source, in the place the repo keeps config:

```python
# osism/settings.py
ANSIBLE_VAULT_KEYFILE = os.getenv(
    "ANSIBLE_VAULT_KEYFILE", "/share/ansible_vault_password.key"
)

# osism/commands/vault.py — drop the `keyfile` class attributes, and in each take_action:
keyfile = settings.ANSIBLE_VAULT_KEYFILE
```

The `keyfile` class attributes lose nothing: no CLI arg sets them, no subclass overrides them, and cliff builds a fresh command instance per invocation, so their only consumers in the tree are `monkeypatch.setattr` calls. The test-side cost is small and centralised — `tests/unit/commands/test_vault.py:199` and `:343` are the only two patch sites, both inside helpers (`_make_set_password`, `_make_check_keyfile`), so each becomes one line patching `settings.ANSIBLE_VAULT_KEYFILE`. `tests/unit/utils/test_init_connections.py:59` already uses that idiom for the `REDIS_*` settings. The fixture in this PR (`tests/integration/test_vault.py:39-41`) collapses from three patches and a five-line docstring to one patch.

Worth saying why this lands on the hoist rather than on pre-existing code: before it, there were three copies of the literal — duplicated, but self-evidently so, and no one could believe patching one moved another. The hoist removed the *visibility* of the duplication without removing the duplication, and added a comment asserting a live link that doesn't exist. That's what makes it easy to get wrong later.

tests/integration/test_vault.py
```diff
@@ -0,0 +1,140 @@
...
+def test_token_from_other_key_raises_invalid_token(vault_keyfile):
+    """A token written under a different Fernet key raises ``InvalidToken``."""
+    _write_key(vault_keyfile)
+    utils.redis.set(VAULT_KEY, Fernet(Fernet.generate_key()).encrypt(b"hunter2"))
+
+    with pytest.raises(InvalidToken):
+        utils.get_ansible_vault_password()
+
+
+def test_missing_keyfile_raises_file_not_found(vault_keyfile):
```
ideaship: This one never touches Redis, but it carries the module's `integration` mark, so it only runs in the Redis-gated job — and it duplicates `test_get_ansible_vault_password_keyfile_missing` in `tests/unit/utils/test_init_task_output.py:423` exactly. Either move it to the unit suite or drop it.

More broadly: `test_init_task_output.py:379-446` already covers all six behaviors of this function with mocks. The cases that earn their integration mark are `test_round_trip_returns_plaintext` and `test_command_round_trip` — real Fernet tokens and a real bytes round trip through Redis catch encoding regressions the mocked tests can't, and the `set` → `check` → `unset` chain is new coverage. The blank-password and foreign-token cases (lines 87-103) are pure-Python branches the unit tests already pin; trimming them keeps the Redis-gated job focused on what actually needs Redis. Your call.

tests/integration/test_vault.py
```diff
@@ -0,0 +1,140 @@
...
+    monkeypatch.setattr(sys, "stdin", _piped_stdin("hunter2\n"))
+    set_password.take_action(set_password.get_parser("test").parse_args([]))
+
+    assert vault_keyfile.exists()
+    assert utils.get_ansible_vault_password() == "hunter2"
+
+    check = vault.Check(MagicMock(), MagicMock())
+    # The last check step decrypts a secrets.yml found on the host. There is
+    # none in CI, and searching for one would tie the test to the filesystem.
+    monkeypatch.setattr(vault.Check, "_find_secrets_file", lambda self: None)
```
ideaship, line 111: Stubbing `_find_secrets_file` is the right call, but the stated reason is weaker than the real one, and the gap invites an "improvement" that would quietly break it.

ansible-core isn't installed here — it lives in the optional `[ansible]` extra (`setup.cfg:34`) while this job installs `--dev` plus `pip install .` — so `tests/conftest.py:82` registers the ansible stubs. The stubbed `VaultLib.decrypt` returns `b""` without raising, while `is_encrypted` really does check the `$ANSIBLE_VAULT` prefix (`tests/conftest.py:53-61`). So step 6 wouldn't merely be host-dependent: given any encrypted file it would take the success branch and log "Vault password successfully decrypted" for *any* password.

That's what makes "there is none in CI" risky as the recorded reason — the natural response to it is to add an encrypted fixture and pass `--path`, which looks like better coverage and is a vacuous assertion. Suggest naming the structural cause instead, and what would change it:

```
# The last check step decrypts a secrets.yml with ansible's VaultLib.
# ansible-core is not installed in this environment, so conftest stubs
# VaultLib and its decrypt() is a no-op that would report success for any
# password -- the step is skipped rather than mocked. Worth covering for
# real if ansible-core is ever added to the test dependencies.
```

ideaship (changes requested):

osism/utils/__init__.py
berendt: Done — moved to `osism/settings.py` in the shape you sketched, including the env var, and dropped both `keyfile` class attributes in favour of a local read in each `take_action`. The two unit helpers (`_make_set_password`, `_make_check_keyfile`) are now one `monkeypatch.setattr(settings, ...)` each, and the integration fixture is a single patch.

The point about the hoist removing the visibility of the duplication without removing the duplication is fair, and it is the part I would not have caught from the diff alone — three literals are obviously three things, one constant read in three places looks like one. The commit message now carries the `SetPassword` fork as the reason rather than the tidiness argument.

tests/integration/test_vault.py
berendt: Isolated the database rather than the key, as suggested — `REDIS_DB: "15"` in `playbooks/test-integration.yml`, plus a guard in `tests/integration/conftest.py` that exits the session when `settings.REDIS_DB == 0` unless `OSISM_ALLOW_DEFAULT_REDIS_DB` is set.

Two details worth flagging:

- The guard runs *after* the reachability check, not before. Ahead of it, a plain `pytest tests/` on a laptop with no Redis and no `REDIS_DB` set would hard-exit 1 where it used to skip. Behind it, the only runs that can trip the guard are the ones that would actually reach a database 0 — which is exactly the manager-node case.
- `README.md:25-41` documented `REDIS_HOST=localhost pipenv run pytest tests/integration`, which the guard would have rejected. The example now sets `REDIS_DB=15` and the warning names the vault key and its unrecoverability alongside the task lock.

It is its own commit, since it protects the whole suite rather than this module.

tests/integration/test_vault.py
berendt: Dropped `test_missing_keyfile_raises_file_not_found` — it was a duplicate carrying an `integration` mark for no reason — and also dropped the blank-password parametrisation, which is a post-decryption branch already pinned twice in `test_init_task_output.py`.

Kept `test_token_from_other_key_raises_invalid_token`. The unit suite mocks `Fernet` wholesale, so its nearest equivalent (`test_get_ansible_vault_password_fernet_decrypt_raises`) asserts a `RuntimeError` it injected itself; nothing outside this module checks that a genuine mismatched token surfaces as `InvalidToken` rather than being reshaped by the `except Exception` wrapper. It also goes through real Redis bytes. Four tests now.

tests/integration/test_vault.py
berendt: Rewritten along those lines — the comment now names the stub as the cause and says why an encrypted fixture with `--path` would be a vacuous assertion rather than better coverage, so the trap is recorded next to the thing that would spring it. Also carried the reason into the commit message and the PR body, both of which said "there is none in CI".

ideaship (approved):

PR #2548, opened 2026-08-03
Add integration tests for the Redis semaphore

### LINKED ISSUE #2401: Integration test: RedisSemaphore / create_netbox_semaphore

## Description

This issue adds live-Redis integration tests for the distributed semaphore in `osism/utils/__init__.py`: the `RedisSemaphore` class and the `create_netbox_semaphore` factory. All tests go into one new module, `tests/integration/test_semaphore.py`, and run in the existing `python-osism-integration-tests` CI job. No production code changes.

Concrete boundaries:

1. **The semaphore implementation already exists.** `RedisSemaphore` (`osism/utils/__init__.py:115`) tracks holders in a Redis sorted set. The constructor prefixes the caller's key, so a semaphore built with `key="x"` uses the Redis key `semaphore:x` (`osism/utils/__init__.py:154`). `acquire()` (`osism/utils/__init__.py:159`) runs the server-side Lua script `_ACQUIRE_LUA` (`osism/utils/__init__.py:131`), which first removes holders scored older than `HOLDER_EXPIRY = 60` seconds (`osism/utils/__init__.py:123`) via `ZREMRANGEBYSCORE`, then checks capacity with `ZCARD` and reserves a slot with `ZADD`. On a full set it retries every 10 ms (`osism/utils/__init__.py:188`) until the timeout and returns `False` (`osism/utils/__init__.py:190`); it never raises on timeout. `release()` (`osism/utils/__init__.py:192`) removes the holder with `ZREM` and returns without touching Redis when `self.identifier` is `None`. `__enter__` (`osism/utils/__init__.py:198`) raises `TimeoutError("Could not acquire semaphore semaphore:<key>")` when it cannot acquire (`osism/utils/__init__.py:201`); `__exit__` releases (`osism/utils/__init__.py:204`). `create_netbox_semaphore(netbox_url, max_connections=None)` (`osism/utils/__init__.py:505`) defaults `max_connections` to `settings.NETBOX_MAX_CONNECTIONS` (default 5, `osism/settings.py:100`) and derives the key `netbox_semaphore_<first 8 hex chars of md5(url)>` (`osism/utils/__init__.py:520-523`); the resulting Redis key is therefore `semaphore:netbox_semaphore_<md5-prefix>`, not the `semaphore:netbox:<url>` format the draft body assumed. The factory wires in the module-global client from `_init_redis()` (`osism/utils/__init__.py:28`) and `timeout=30` (`osism/utils/__init__.py:525-530`). **This issue adds** tests that exercise this behavior against a real Redis and assert the key format as implemented.

2. **The integration-test infrastructure already exists.** `tests/integration/conftest.py:70` skips integration-marked tests when Redis is unreachable and turns that into a session failure when `OSISM_REQUIRE_REDIS` is set. The Zuul job `python-osism-integration-tests` (`.zuul.yaml:139`) runs `playbooks/test-integration.yml`, which starts a single `redis:7-alpine` container and runs `pytest tests/integration` with `REDIS_HOST=localhost` and `OSISM_REQUIRE_REDIS=1`. The `integration` marker is registered in `setup.cfg:187`, and `testpaths = tests/unit` keeps the module out of plain `pytest` runs. **This issue adds** one module that plugs into this unchanged. The `celery_app` / `celery_worker` fixtures are not used; the tests need Redis only.

3. **Unit coverage already exists and stops short of a real server.** `tests/unit/utils/test_init_semaphore.py` runs the production Lua against `fakeredis` and states in its module docstring that "A faithful many-client race needs a real Redis and lives outside the unit suite" (`tests/unit/utils/test_init_semaphore.py:8`). `tests/unit/utils/test_init_locks.py` covers the constructor, `release()`, the context manager and `create_netbox_semaphore` key derivation with mocks. **This issue adds** the race test that docstring delegates here, plus live-server checks of the behaviors `fakeredis` reimplements in Python and can get wrong: sorted-set commands (`ZADD` / `ZCARD` / `ZREMRANGEBYSCORE`), server-side Lua evaluation, and real timing.

Module design. The module follows `tests/integration/test_locking.py`: SPDX header, module docstring, `pytestmark = pytest.mark.integration`, and a unique key per test so runs stay independent of prior state. Each test passes `itest-sem-<uuid4>` as the constructor key; any raw seed or inspection through the Redis client must target the prefixed form `semaphore:itest-sem-<uuid4>`. A module-scoped `redis_client` fixture builds a `redis.Redis` from `osism.settings` (`REDIS_HOST` / `REDIS_PORT` / `REDIS_DB`, the same pattern as `_redis_reachable` in `tests/integration/conftest.py:50-64`), yields it, and closes it in teardown. The client is created without `decode_responses`, so `zrange` returns `bytes`; compare members against `sem.identifier.encode()` or decode them, as `tests/unit/utils/test_init_semaphore.py:95` does. Overflow and timeout tests pass `timeout=0.5` or shorter so the module finishes in a few seconds.

The ten test functions:

1. `test_acquire_up_to_maxsize`: three `RedisSemaphore` instances on one shared key with `maxsize=3` each return `True` from `acquire()`; `ZCARD` of the prefixed key is 3; all are released afterwards and `ZCARD` drops to 0.
2. `test_acquire_beyond_maxsize_returns_false`: with `maxsize=1` and one holder held, a second instance's `acquire(timeout=0.5)` returns `False` and the holder count stays 1.
3. `test_release_frees_slot`: with `maxsize=1`, after `release()` on the holder, a previously refused instance's `acquire()` returns `True`.
4. `test_release_without_acquire_is_noop`: calling `release()` on an instance that never acquired returns without error, `identifier` stays `None`, and `ZCARD` stays 0.
5. `test_expired_holder_reclaimed`: seed an aged holder with `redis_client.zadd("semaphore:<key>", {"stale": time.time() - 61})` (61 seconds is past `HOLDER_EXPIRY = 60`); a fresh instance with `maxsize=1` acquires immediately, and the sorted set then contains only the new holder's identifier (the stale member was removed by `ZREMRANGEBYSCORE` inside the Lua script).
6. `test_context_manager_roundtrip`: inside `with RedisSemaphore(redis_client, key, 1, timeout=5):` the holder count is 1; after the block it is 0.
7. `test_context_manager_timeout_raises`: seed a freshly scored holder into the prefixed key, then `with pytest.raises(TimeoutError):` around a `with` on an instance with `maxsize=1, timeout=0.2`; the message is `Could not acquire semaphore semaphore:<key>`.
8. `test_create_netbox_semaphore_key_and_maxsize`: for a unique URL, `create_netbox_semaphore(url, max_connections=2)` returns a `RedisSemaphore` whose `key` equals `"semaphore:netbox_semaphore_" + hashlib.md5(url.encode()).hexdigest()[:8]` and whose `maxsize` is 2; the returned instance's `acquire()` and `release()` round-trip against the live Redis (this exercises the `_init_redis()` client for real, which the mocked unit tests cannot).
9. `test_create_netbox_semaphore_default_maxsize`: without `max_connections`, the returned semaphore's `maxsize` equals `settings.NETBOX_MAX_CONNECTIONS`.
10. `test_concurrent_acquire_never_exceeds_maxsize`: 12 threads, released together by a `threading.Barrier`, each construct their own `RedisSemaphore` on one shared key with `maxsize=3` and call `acquire(timeout=0.5)`; successful threads hold their slot until all threads finish. Exactly 3 acquisitions succeed, `ZCARD` is 3 while they are held, and releasing all brings it to 0. The shared `redis_client` is safe here; `redis.Redis` is thread-safe through its connection pool.

## Motivation

The semaphore is what caps concurrent NetBox API requests: `osism/tasks/netbox.py:31` builds one per NetBox URL via `utils.create_netbox_semaphore(nb.base_url)`. If the atomic acquire admits more holders than `maxsize`, NetBox gets hammered by exactly the load the semaphore exists to prevent; if it admits fewer, tasks serialize and slow down.

Today the only evidence the acquire path works comes from `fakeredis` and mocks. `fakeredis` reimplements sorted sets and Lua evaluation in Python, which is exactly the surface this class depends on, and the unit suite says itself that the many-client race belongs in a real-Redis suite (`tests/unit/utils/test_init_semaphore.py:8`). The integration job from #2368 provides that real Redis, and the meta issue #2400 lists this module first among the Tier 1 gaps.

This issue closes the gap: the semaphore's contended path, its expiry reclaim and its factory run against `redis:7-alpine` in CI, and the race check the unit suite delegated finally exists.

## Affected Areas

- `tests/integration/test_semaphore.py` (new module: the `redis_client` fixture and the ten tests listed in the Description; the only file this issue touches)

## Acceptance Criteria

- [ ] Add `tests/integration/test_semaphore.py` with an SPDX header, a module docstring, and `pytestmark = pytest.mark.integration`, following `tests/integration/test_locking.py`.
- [ ] Use a unique `itest-sem-<uuid4>` constructor key per test, and target the prefixed `semaphore:itest-sem-<uuid4>` key for every raw seed or inspection.
- [ ] Verify `maxsize` holders acquire: with `maxsize=3`, three instances on one key return `True` from `acquire()` and `ZCARD` is 3 (`test_acquire_up_to_maxsize`).
- [ ] Verify overflow times out: on a full semaphore, `acquire(timeout=0.5)` returns `False` without raising and leaves the holder count unchanged (`test_acquire_beyond_maxsize_returns_false`).
- [ ] Verify `release()` frees a slot: after one holder releases, a subsequent `acquire()` returns `True` (`test_release_frees_slot`).
- [ ] Verify `release()` without a prior acquire is a no-op: no exception, `identifier` stays `None`, `ZCARD` stays 0 (`test_release_without_acquire_is_noop`).
- [ ] Verify expired-holder reclaim: a member seeded with score `time.time() - 61` is removed by the next `acquire()` on a `maxsize=1` semaphore, the acquire returns `True`, and the sorted set contains only the new holder's identifier (`test_expired_holder_reclaimed`).
- [ ] Verify the context manager acquires and releases: holder count is 1 inside the `with` block and 0 after it (`test_context_manager_roundtrip`).
- [ ] Verify the context manager raises `TimeoutError` with the message `Could not acquire semaphore semaphore:<key>` when the semaphore stays full past `timeout=0.2` (`test_context_manager_timeout_raises`).
- [ ] Verify `create_netbox_semaphore(url, max_connections=2)` returns a semaphore with `key == "semaphore:netbox_semaphore_" + hashlib.md5(url.encode()).hexdigest()[:8]` and `maxsize == 2`, and that it acquires and releases against the live Redis (`test_create_netbox_semaphore_key_and_maxsize`).
- [ ] Verify `create_netbox_semaphore(url)` with `max_connections=None` falls back to `settings.NETBOX_MAX_CONNECTIONS` (`test_create_netbox_semaphore_default_maxsize`).
- [ ] Verify the race invariant: 12 barrier-synchronized threads on one key with `maxsize=3` produce exactly 3 successful `acquire(timeout=0.5)` calls while holders are held, `ZCARD` is 3, and releasing all holders brings it to 0 (`test_concurrent_acquire_never_exceeds_maxsize`).
- [ ] Verify the module skips instead of failing when Redis is unreachable and `OSISM_REQUIRE_REDIS` is unset: `pytest tests/integration/test_semaphore.py` on a machine without Redis reports every test as skipped with the reason from `tests/integration/conftest.py:79`.
- [ ] Verify `pytest tests/integration` passes in the `python-osism-integration-tests` Zuul job with the new module included.

## Non-Goals

- Changes to `RedisSemaphore` or `create_netbox_semaphore` in `osism/utils/__init__.py`: this issue asserts the behavior as implemented, including the `semaphore:netbox_semaphore_<md5-prefix>` key format.
- `fakeredis`-based semaphore tests: `tests/unit/utils/test_init_semaphore.py` owns that layer.
- Coverage of the other Tier 1 areas: the vault password round-trip is #2402, facts freshness is #2403, Celery error/state/revoke paths are #2404, and EventBridge pub/sub is #2405.
- New services or test dependencies: the meta issue #2400 pins Tier 1 to the existing single-Redis job; `Pipfile` and `playbooks/test-integration.yml` stay untouched.

## References

- Meta issue #2400 (Tier 1 / Tier 2 plan) and #2368 (integration-test foundation).
- `osism/utils/__init__.py:115` (`RedisSemaphore`) and `osism/utils/__init__.py:505` (`create_netbox_semaphore`).
- `tests/integration/conftest.py` (skip logic and fixtures), `tests/integration/test_locking.py` (module conventions).
- `tests/unit/utils/test_init_semaphore.py` (fakeredis Lua coverage and the delegated race note), `tests/unit/utils/test_init_locks.py` (mock-based coverage).
- `.zuul.yaml:139` and `playbooks/test-integration.yml` (the `python-osism-integration-tests` job).
- `setup.cfg:187` (registration of the `integration` marker).

**Executability score:** 9/10

---

berendt:
`RedisSemaphore` caps concurrent NetBox API requests: `osism/tasks/netbox.py:31` builds one per NetBox URL through `utils.create_netbox_semaphore(nb.base_url)`. Its holders live in a sorted set and are admitted by a server-side Lua script, so what the class does is Redis behaviour: the expiry reclaim, the capacity limit, and the atomicity that stops two clients from taking the same free slot. Until now the only coverage came from `fakeredis` and mocks, and `fakeredis` reimplements exactly that surface in Python.

One commit, one new file, no production code and no dependency changes.

`Add integration tests for the Redis semaphore` adds `tests/integration/test_semaphore.py` with ten tests against the live Redis of the `python-osism-integration-tests` job:

- `test_acquire_up_to_maxsize`, `test_acquire_beyond_maxsize_returns_false`, `test_release_frees_slot` and `test_release_without_acquire_is_noop` cover the capacity limit and both release paths.
- `test_expired_holder_reclaimed` seeds a holder aged past `HOLDER_EXPIRY` directly through the Redis client, so the `ZREMRANGEBYSCORE` reclaim runs without waiting out 60 seconds.
- `test_context_manager_roundtrip` and `test_context_manager_timeout_raises` cover both `with sem:` paths, including the exact `TimeoutError` message.
- `test_create_netbox_semaphore_key_and_maxsize` and `test_create_netbox_semaphore_default_maxsize` pin the key the helper derives from a NetBox URL (`semaphore:netbox_semaphore_<md5-prefix>`), its `maxsize`, and an acquire/release round trip against the live server.
- `test_concurrent_acquire_never_exceeds_maxsize` is the many-client race the unit suite defers to a real server (`tests/unit/utils/test_init_semaphore.py:8`): twelve barrier-synchronised threads contend for three slots and exactly three get in.

Each test uses a uuid key and deletes it afterwards, so runs stay independent of one another.

## Verification

Against a throwaway `redis:7-alpine` container, `pytest tests/integration/test_semaphore.py` reports 10 passed, repeated three times without flakiness. With Redis stopped it reports 10 skipped rather than failing, per the skip logic in `tests/integration/conftest.py`.

The race test was checked for teeth: replaying the same twelve threads with a deliberately non-atomic acquire (`ZCARD` then `ZADD` in two round trips) admitted 6 to 7 holders where `maxsize` is 3, so the test fails loudly on the bug the Lua script prevents.

Closes #2401

### DIFF
--- a/tests/integration/test_semaphore.py
+++ b/tests/integration/test_semaphore.py
@@ -1,0 +1,149 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Redis-semaphore integration tests against a live Redis.
+
+``RedisSemaphore`` caps concurrent NetBox API requests: ``osism.tasks.netbox``
+builds one per NetBox URL through ``create_netbox_semaphore``. Holders live in a
+sorted set and are admitted by a server-side Lua script (``ZREMRANGEBYSCORE``
+plus ``ZCARD`` plus ``ZADD``).
+
+The script's logic is not what needs a live server:
+``tests/unit/utils/test_init_semaphore.py`` runs the production ``_ACQUIRE_LUA``
+against ``fakeredis``, which executes Lua through ``lupa``. What only the real
+server provides is Redis' own Lua sandbox and ``redis.call`` bindings, the
+numeric coercion of the ``now``, ``maxsize`` and expiry arguments that ``lupa``
+only approximates, and the race the unit suite explicitly defers to here:
+many clients contending for the same free slot at the same moment.
+"""
+
+import concurrent.futures
+import hashlib
+import threading
+import uuid
+
+import pytest
+
+from osism import utils
+
+pytestmark = pytest.mark.integration
+
+
+def semaphore_redis_key(key):
+    """The prefixed key ``RedisSemaphore`` stores its sorted set under."""
+    return f"semaphore:{key}"
+
+
+@pytest.fixture
+def redis_client():
+    """The shared Redis client the semaphore itself uses.
+
+    ``create_netbox_semaphore`` wires this very client into the semaphores it
+    builds, so seeding and inspecting through it observes what the code under
+    test sees. It is a cached module global held for the whole session, so it
+    is not closed here: that would break every later test that touches it.
+    """
+    return utils.redis
+
+
+@pytest.fixture
+def semaphore_key(redis_client):
+    """A key unique to this test, removed from Redis afterwards.
+
+    ``RedisSemaphore`` prefixes what it is handed, so the sorted set lives under
+    ``semaphore:<key>``. Raw seeding and inspection use that prefixed form.
+    """
+    key = f"itest-sem-{uuid.uuid4()}"
+    yield key
+    redis_client.delete(semaphore_redis_key(key))
+
+
+def test_release_frees_slot(redis_client, semaphore_key):
+    """The slot a holder gives up is granted to the caller that was refused.
+
+    The unit suite covers a slot freed by expiry, never one freed by
+    ``release`` and granted to a waiter inside its retry loop.
+    """
+    holder = utils.RedisSemaphore(redis_client, semaphore_key, 1, timeout=5)
+    assert holder.acquire() is True
+
+    waiter = utils.RedisSemaphore(redis_client, semaphore_key, 1, timeout=0.5)
+    assert waiter.acquire() is False
+
+    holder.release()
+
+    assert waiter.acquire() is True
+    assert redis_client.zcard(waiter.key) == 1
+
+    waiter.release()
+
+
+def test_create_netbox_semaphore_key_and_maxsize(redis_client):
+    """The NetBox helper derives its key from the URL and works against Redis."""
+    url = f"https://netbox-{uuid.uuid4()}.example"
+    url_hash = hashlib.md5(url.encode()).hexdigest()[:8]
+    redis_key = f"semaphore:netbox_semaphore_{url_hash}"
+
+    sem = utils.create_netbox_semaphore(url, max_connections=2)
+
+    try:
+        assert isinstance(sem, utils.RedisSemaphore)
+        assert sem.key == redis_key
+        assert sem.maxsize == 2
+
+        assert sem.acquire() is True
+        members = {member.decode() for member in redis_client.zrange(redis_key, 0, -1)}
+        assert members == {sem.identifier}
+
+        sem.release()
+        assert redis_client.zcard(redis_key) == 0
+    finally:
+        redis_client.delete(redis_key)
+
+
+def test_concurrent_acquire_never_exceeds_maxsize(redis_client, semaphore_key):
+    """Under a real race only ``maxsize`` of many contending clients get in.
+
+    The atomic Lua script is what makes this hold: a capacity check and a slot
+    reservation split across two round trips would let several clients observe
+    the same free slot and all take it. This is the many-client race the unit
+    suite defers to a live server; ``fakeredis`` never reproduces it, because a
+    single sequential client cannot over-admit.
+
+    One round only lands the interleaving that exposes an over-admitting
+    implementation about nine times in ten, so the race is run repeatedly. A
+    round costs a full acquire timeout, since the losers each wait theirs out,
+    which is why the timeout here is much shorter than elsewhere in this file.
+
+    Every thread hands its result back through a future, so a worker that dies
+    fails the test. Left unchecked it would only warn, and the race this test
+    exists for would silently shrink to the threads that survived.
+    """
+    maxsize = 3
+    thread_count = 12
+    rounds = 5
+    redis_key = semaphore_redis_key(semaphore_key)
+
+    def contend(barrier):
+        sem = utils.RedisSemaphore(redis_client, semaphore_key, maxsize, timeout=0.1)
+        barrier.wait()
+        return sem if sem.acquire() else None
+
+    for _ in range(rounds):
+        # A timeout on the barrier turns a thread that never arrives into a
+        # failed assertion rather than a hung CI job.
+        barrier = threading.Barrier(thread_count, timeout=30)
+
+        with concurrent.futures.ThreadPoolExecutor(max_workers=thread_count) as pool:
+            futures = [pool.submit(contend, barrier) for _ in range(thread_count)]
+            acquired = [sem for sem in (future.result() for future in futures) if sem]
+
+        # Nothing is released before every future has been collected, so the
+        # holder count is stable here.
+        assert len(acquired) == maxsize
+        assert redis_client.zcard(redis_key) == maxsize
+
+        for sem in acquired:
+            sem.release()
+
+        assert redis_client.zcard(redis_key) == 0
+        redis_client.delete(redis_key)


tests/integration/test_semaphore.py
```diff
@@ -0,0 +1,225 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Redis-semaphore integration tests against a live Redis.
+
+``RedisSemaphore`` caps concurrent NetBox API requests: ``osism.tasks.netbox``
+builds one per NetBox URL through ``create_netbox_semaphore``. Holders live in a
+sorted set and are admitted by a server-side Lua script (``ZREMRANGEBYSCORE``
+plus ``ZCARD`` plus ``ZADD``), so what these tests exercise is Redis behaviour:
+the expiry reclaim, the capacity limit, and the atomicity that stops two
+clients from taking the same free slot. ``fakeredis`` reimplements all of that
```
ideaship: This rationale isn't accurate, and it's load-bearing — it's the stated reason the file exists. `tests/unit/utils/test_init_semaphore.py:4-7` runs the production `_ACQUIRE_LUA` against fakeredis **with `lupa`** (`Pipfile:51-52` declares both), so the unit suite already executes the real server-side script rather than reimplementing it in Python or asserting on a mock's recorded calls.

There is a genuine argument for a live server here, and it's narrower: Redis' own Lua sandbox, its `redis.call` bindings, and the numeric coercion of the `now` / `maxsize` / `expiry` arguments, which `lupa` only approximates — plus the many-client race that the unit docstring itself carves out at `:8` as needing "a real Redis". Worth stating that instead. The commit message carries the same claim and would want the same correction.

tests/integration/test_semaphore.py
```diff
@@ -0,0 +1,225 @@
...
+@pytest.fixture
+def semaphore_key(redis_client):
+    """A key unique to this test, removed from Redis afterwards.
+
+    ``RedisSemaphore`` prefixes what it is handed, so the sorted set lives under
+    ``semaphore:<key>``. Raw seeding and inspection use that prefixed form.
+    """
+    key = f"itest-sem-{uuid.uuid4()}"
+    yield key
+    redis_client.delete(f"semaphore:{key}")
```
ideaship: The `"semaphore:"` prefix is rebuilt by hand at eleven places in this file (`:47, 60, 65, 77, 93, 105, 114, 130, 143, 162, 198`) while `sem.key` — which is public — is used only at `:168`. A prefix change would break eleven assertions for reasons unrelated to what each test is checking.

Not all eleven want the same treatment. Where a semaphore is already in scope (`:60, 65, 77, 93, 105`) `sem.key` substitutes directly. `:114, 130, 143` need only a one-line reorder to construct the semaphore first. For this fixture and the per-thread constructor at `:198` there's no instance yet, so a small helper is the right shape:

```python
def semaphore_redis_key(key):
    return f"semaphore:{key}"
```

Keep `:162` hand-written — that one *is* the assertion in `assert sem.key == redis_key`, and substituting `sem.key` there would delete the test.

tests/integration/test_semaphore.py
```diff
@@ -0,0 +1,225 @@
...
+
+
+def test_acquire_up_to_maxsize(redis_client, semaphore_key):
```
ideaship: Scope question for the file as a whole, and it splits three ways rather than two — the overlap with the unit suite isn't all the same kind.

**Genuinely redundant — delete.** `test_acquire_up_to_maxsize` (`:50`) and `test_acquire_beyond_maxsize_returns_false` (`:68`) are `tests/unit/utils/test_init_semaphore.py:43 test_acquire_never_exceeds_maxsize` split in two, and the unit version asserts strictly more (it also pins that the refused caller leaves the set unchanged). `test_expired_holder_reclaimed` (`:108`) is weaker than its unit twin at `:79`: seeding at `time.time() - HOLDER_EXPIRY - 1` (`:115`) means the holder is *already* expired, so the first `eval` reclaims it and the retry loop never runs, whereas `test_reclaim_boundary_advances_across_retries` seeds at "now" with a shrunk `HOLDER_EXPIRY` so the holder ages out mid-loop. Mutation check: hoist `now = time.time()` out of the `while` in `acquire()` (`osism/utils/__init__.py:173`) — the unit test goes red, this one stays green.

**Not redundant, but no live-Redis signal — move to `tests/unit/`, don't delete.** No unit test currently covers `release()` short-circuiting on `identifier is None` (`:98`), `__enter__`/`__exit__` (`:128`), the `TimeoutError` raise from `__enter__` (`:141`), or the `max_connections` fallback (`:181`). Deleting these loses real coverage; they're simply at the wrong layer. `:98` issues no Lua at all — `release()` returns before `zrem`, so the only server contact is a `zcard` on a key nobody wrote — `:128` exercises four lines of pure Python, and `:181` performs no semaphore operation whatsoever.

**Keep.** `test_release_frees_slot` (`:82`) is not a duplicate: the unit suite covers a slot freed by *expiry*, never one freed by *release* and granted to a waiter mid-retry-loop. `test_create_netbox_semaphore_key_and_maxsize` (`:158`) is a real roundtrip through the production helper. `test_concurrent_acquire_never_exceeds_maxsize` (`:188`) is the file's reason to exist.

That leaves the Redis-gated job running what actually needs Redis, and drops ~1.2 s of real `time.sleep` retry loops (`timeout=0.5`, `0.5`, `0.2`) from it.

tests/integration/test_semaphore.py
```diff
@@ -0,0 +1,225 @@
...
+        members = {member.decode() for member in redis_client.zrange(redis_key, 0, -1)}
+        assert members == {sem.identifier}
+
+        sem.release()
+        assert redis_client.zcard(redis_key) == 0
+    finally:
+        redis_client.delete(redis_key)
+
+
+def test_create_netbox_semaphore_default_maxsize():
```
ideaship: This one performs no semaphore operation, and its assertion reads the same setting the helper reads (`osism/utils/__init__.py:516-517`), so it largely asserts that a value equals itself. It does still detect a literal substitution or a deleted fallback branch, so it's worth keeping rather than dropping — but as a unit test with a sentinel:

```python
mocker.patch.object(settings, "NETBOX_MAX_CONNECTIONS", 17)
assert utils.create_netbox_semaphore("https://netbox.example").maxsize == 17
```

That pins the fallback branch instead of the tautology. Note it isn't free of Redis today: `create_netbox_semaphore` calls `_init_redis()` (`osism/utils/__init__.py:528`), which constructs a client and pings, so the test currently needs a reachable server to build an object it never uses. Moved to the unit suite it wants `_init_redis` mocked — `tests/unit/utils/test_init_connections.py` already has the idiom.

tests/integration/test_semaphore.py
```diff
@@ -0,0 +1,225 @@
...
+    """Without ``max_connections`` the helper falls back to the setting."""
+    sem = utils.create_netbox_semaphore(f"https://netbox-{uuid.uuid4()}.example")
+
+    assert sem.maxsize == settings.NETBOX_MAX_CONNECTIONS
+
+
+def test_concurrent_acquire_never_exceeds_maxsize(redis_client, semaphore_key):
```
ideaship, line 103: This is the only test that exercises the invariant under real concurrency, and it runs the race exactly once. Worth being precise about what that covers, because the obvious reading is stronger than the truth.

A revert of the atomic script to separate round trips is caught deterministically elsewhere: `tests/unit/utils/test_init_locks.py:61-83` and five neighbours assert `redis.eval.assert_called_once_with(_ACQUIRE_LUA, …)`, so they fail whether or not an interleaving lands. And the race cannot be reintroduced *inside* the script, since Redis executes an `EVAL` atomically. So this test is not what stands between a straight revert and a release.

Where it is the only guard is a refactor rather than a revert — `register_script`/`evalsha`, say — where those mock assertions get rewritten to match the new call, and part of the capacity decision moves client-side in the process. The rewritten shape tests bless the new shape, the sequential fakeredis tests cannot over-admit with a single client, and this is the only test left that would notice.

For that case, one round is roughly a 1-in-10 miss. With `acquire()` mutated to three round trips, this test caught it in 54 of 60 cold-process runs against a live Redis (and in 0 of 100 runs against fakeredis, which is why it belongs here rather than in the unit suite). Wrapping the barrier, threads and assertions in `for _ in range(5)` with a `redis_client.delete(redis_key)` between rounds takes the miss probability to ~1e-5.

One cost note, since this file's runtime is already mostly waiting: a round costs a full acquire timeout, because the nine losers each wait theirs out — 0.51 s measured. Five rounds at `timeout=0.5` would add ~2 s. Dropping *this test's* acquire timeout to 0.1 makes a round 0.11 s with detection unchanged (58 of 60 measured), so five rounds cost about what one round costs today.

tests/integration/test_semaphore.py
```diff
@@ -0,0 +1,225 @@
...
+    maxsize = 3
+    thread_count = 12
+    redis_key = f"semaphore:{semaphore_key}"
+    # A timeout on the barrier turns a thread that never arrives into a failed
+    # assertion rather than a hung CI job.
+    barrier = threading.Barrier(thread_count, timeout=30)
+    acquired = []
+    acquired_lock = threading.Lock()
+
+    def contend():
```
ideaship: `contend()` has no error handling, and nothing propagates a worker failure back to the test thread. A thread that dies inside `acquire()` goes through `threading.excepthook`, pytest's threadexception plugin downgrades it to a `PytestUnhandledThreadExceptionWarning`, and `setup.cfg:184` sets `addopts = -ra --strict-markers` with no `filterwarnings = error` — so the test still passes.

Concretely: three threads win their slots and the other nine die on a transient error after crossing the barrier but before contending. `len(acquired) == 3` and `zcard == 3` both hold, the job goes green, and the twelve-way race this file exists to exercise silently degenerated into a three-way one. That matters more here than the general unchecked-thread pattern, because this is the only behavioural check on the atomicity guarantee — and it compounds the single-round point above: the one round that carries the signal can quietly not happen.

Simplest fix is to drop the manual threads for a pool and call `result()` on every future, so propagation is automatic rather than opt-in. It also removes the `acquired_lock` bookkeeping:

```python
with concurrent.futures.ThreadPoolExecutor(max_workers=thread_count) as pool:
    futures = [pool.submit(contend) for _ in range(thread_count)]
    acquired = [sem for sem in (f.result() for f in futures) if sem]
```

with `contend()` returning `sem` or `None`. Note the barrier keeps working unchanged: a thread that dies before `barrier.wait()` leaves the rest to time out after 30 s with `BrokenBarrierError`, which now surfaces as a failure instead of a warning.

ideaship (changes requested):

tests/integration/test_semaphore.py
berendt: Corrected. The docstring no longer claims fakeredis reimplements the script in Python; it now names `tests/unit/utils/test_init_semaphore.py` as running the real `_ACQUIRE_LUA` through lupa, and states the narrower case for a live server: Redis' own Lua sandbox and `redis.call` bindings, the numeric coercion of the `now`/`maxsize`/expiry arguments, and the many-client race. The commit message carries the same correction.

tests/integration/test_semaphore.py
berendt: Added `semaphore_redis_key()` and used it in the fixture and in the concurrency test. After the deletions below, the only hand-written prefix left is the one in `test_create_netbox_semaphore_key_and_maxsize`, which is the assertion itself, as you noted. `test_release_frees_slot` now asserts through `waiter.key`.

tests/integration/test_semaphore.py
berendt: Agreed on the first group: `test_acquire_up_to_maxsize`, `test_acquire_beyond_maxsize_returns_false` and `test_expired_holder_reclaimed` are gone. The mutation check confirms the point about the reclaim test.

The second group is deleted rather than moved, because the unit coverage already exists — in `tests/unit/utils/test_init_locks.py`, not in `test_init_semaphore.py`:

* `release()` short-circuit → `test_release_without_acquire_is_noop:206` (asserts `zrem` not called)
* `__enter__`/`__exit__` → `test_enter_returns_self_when_acquire_succeeds:221`, `test_exit_calls_release_and_returns_false:243`, `test_exit_does_not_swallow_exceptions:254`
* `TimeoutError` from `__enter__` → `test_enter_raises_timeout_error_when_acquire_fails:233`
* `max_connections` fallback → `test_create_netbox_semaphore_uses_settings_default:358`

Moving them would have duplicated those. The file is down to the three keepers.

tests/integration/test_semaphore.py
berendt: Deleted rather than moved: `tests/unit/utils/test_init_locks.py:358` `test_create_netbox_semaphore_uses_settings_default` is already the test you describe — it patches `_init_redis` to a MagicMock, sets `NETBOX_MAX_CONNECTIONS` to the sentinel 7 and asserts `maxsize == 7`. Good catch on the Redis dependency: it was reaching a live server to build an object it never used.

tests/integration/test_semaphore.py
berendt: Five rounds now, with the acquire timeout dropped to 0.1 for this test and a `delete()` on the key between rounds. The docstring states what the repetition buys and why the timeout is short here, so neither reads as an arbitrary constant.

tests/integration/test_semaphore.py
berendt: Switched to a `ThreadPoolExecutor`; `contend()` returns the semaphore or `None` and every future is collected with `result()`, so a dead worker fails the test. The `acquired_lock` bookkeeping is gone. The barrier is rebuilt per round and still carries its 30s timeout.

ideaship (approved):

PR #2549, opened 2026-08-03
Add integration tests for the Ansible facts freshness check

### LINKED ISSUE #2403: Integration test: check_ansible_facts freshness

## Description

This issue adds live-Redis integration tests for `osism.utils.check_ansible_facts()` (`osism/utils/__init__.py:596`), the freshness check that `osism apply` runs before executing plays (`osism/commands/apply.py:422`). The function scans Redis for `ansible_facts*` keys with a cursor-based `SCAN` loop, reads `ansible_date_time.epoch` from each JSON value, and emits loguru warnings for a missing cache or stale hosts. The new module `tests/integration/test_facts.py` seeds real keys through the live Redis client and asserts on the captured log output.

Concrete boundaries:

1. The function under test already exists and is unit-tested against a mocked Redis: `tests/unit/utils/test_init_task_output.py:450` covers scan failure, no keys, stale, fresh, missing epoch, and malformed JSON with `MagicMock` clients. What the mocks cannot exercise is the real wire path: the cursor loop over `SCAN` (`osism/utils/__init__.py:617`), keys and values arriving as `bytes` and being decoded (`osism/utils/__init__.py:638`), and `json.loads` over data that round-tripped through a Redis server. This issue adds exactly that live coverage and changes no source code.

2. The integration infrastructure already exists from #2368: `tests/integration/conftest.py:70` skips `integration`-marked tests when Redis is unreachable, or fails the session under `OSISM_REQUIRE_REDIS=1`, and the `python-osism-integration-tests` Zuul job (`playbooks/test-integration.yml`) runs `pytest tests/integration` against a fresh `redis:7-alpine` container. This issue adds one test module under that roof and needs no infrastructure change (Tier 1 in #2400).

3. The `loguru_logs` fixture already reaches the integration suite: it lives in `tests/conftest.py:86`, and pytest applies a `conftest.py` to every directory below it, so modules in `tests/integration/` can request it as-is. The draft's note that it "may need to be made available" does not hold; nothing moves.

4. The seeding convention already exists in `tests/integration/test_api_facts.py:38` (from #2406): a module-local `seed_facts` fixture writes `ansible_facts<host>` keys through the lazy `osism.utils.redis` client (`osism/utils/__init__.py:700`), uses `itest-<uuid4>` hostnames, and deletes every seeded key in its teardown after `yield`. This issue adds an equivalent module-local `seed_facts` fixture to `test_facts.py`; it is not hoisted into a shared conftest.

The module carries `pytestmark = pytest.mark.integration` (registered in `setup.cfg:186`) and a docstring naming the function under test, and contains six tests:

1. No keys: with no `ansible_facts*` key in Redis, `check_ansible_facts(max_age=300)` emits a WARNING containing "No Ansible facts found in Redis cache" (`osism/utils/__init__.py:625`). Before seeding anything, the test probes for pre-existing keys with `utils.redis.scan_iter(match="ansible_facts*")` and calls `pytest.skip` when any exist. Per the author's decision: against a developer's Redis holding real facts the test must not delete foreign data, while the CI container starts empty, so the test always runs there.

2. Fresh facts: seed one host whose value is `json.dumps({"ansible_date_time": {"epoch": str(int(time.time()))}})` (Ansible stores epoch as a string; `float()` at `osism/utils/__init__.py:657` accepts it). After `check_ansible_facts(max_age=300)`, no WARNING record contains that hostname.

3. Stale facts: seed one host with an epoch 9999 seconds in the past and a second fresh host in the same test. `check_ansible_facts(max_age=300)` emits a WARNING containing `Host '<stale-host>': facts are` and `seconds old` (`osism/utils/__init__.py:679`), a WARNING containing "Run 'osism sync facts' to update facts." (`osism/utils/__init__.py:676`), and no WARNING containing the fresh hostname. Do not assert the exact count in "stale for N host(s)": a shared local Redis may hold unrelated stale facts that raise N.

4. Missing epoch: seed a host with `json.dumps({"ansible_hostname": "x"})`. The call emits a DEBUG record containing "facts missing ansible_date_time.epoch" with the hostname (`osism/utils/__init__.py:653`), no WARNING containing the hostname, and completes without raising.

5. Malformed JSON: seed a host with the value `{ not valid json`. Inside the function `json.loads` raises `json.JSONDecodeError`, which the `except (json.JSONDecodeError, ValueError, TypeError)` block at `osism/utils/__init__.py:660` catches. The call emits a DEBUG record starting with "Skipping malformed ansible_facts entry", no WARNING containing the hostname, and completes without raising. The debug call at `osism/utils/__init__.py:664` uses printf-style `%r` placeholders that loguru's `str.format` formatting leaves untouched, so the captured message is the literal text "Skipping malformed ansible_facts entry for key %r: %r"; assert on the prefix, not on interpolated values.

6. Empty value: seed a host with the empty string. `r.get(key)` returns falsy bytes and the `if not data: continue` branch at `osism/utils/__init__.py:647` skips the host silently. The call emits no WARNING and no DEBUG containing the hostname and completes without raising.

Two rules keep the module deterministic against a non-empty Redis. Hostnames are `itest-<uuid4>`, which never collide with `LOCAL_FACT_HOSTS` (`osism/utils/__init__.py:25`), whose facts the function skips at `osism/utils/__init__.py:643`. Every log assertion is scoped to a seeded hostname or to a message unique to its scenario, never "no warnings at all".

Pass `max_age=300` explicitly in every call. The default `settings.FACTS_MAX_AGE` (`osism/settings.py:39`) is 43200 seconds and would tie the stale scenario to the environment. Importing `osism.utils` at module level is safe without Redis: `utils.redis` is a lazy attribute resolved on first access (`osism/utils/__init__.py:700`), and the skip logic in `tests/integration/conftest.py:70` prevents any test body from running when Redis is down.

## Motivation

`osism apply` calls `check_ansible_facts()` before running plays (`osism/commands/apply.py:422`) to warn operators about a missing or outdated facts cache. A wrong answer in either direction has a cost: a missed stale warning lets plays run against outdated facts, and a false warning sends operators chasing a sync that is not needed.

The existing unit tests mock every Redis interaction, so the parts most likely to regress are exactly the parts they simulate: the cursor-based `SCAN` pagination, the bytes-to-str decoding of keys and values, and JSON parsing of data a real server returned. A regression in the match pattern or the cursor loop would pass the mocked tests unchanged.

Since #2368 the CI job runs a live Redis, so this gap closes at the cost of one test module. This issue covers the `check_ansible_facts` item of the Tier 1 list in #2400; the remaining Tier 1 modules belong to their own sub-issues.

## Affected Areas

- `tests/integration/test_facts.py` (new module: six integration tests plus a module-local `seed_facts` fixture)

## Acceptance Criteria

- [ ] Add `tests/integration/test_facts.py` with `pytestmark = pytest.mark.integration`, a module docstring naming `check_ansible_facts`, and a module-local `seed_facts` fixture that writes `ansible_facts<host>` keys through `osism.utils.redis` and deletes every seeded key in its teardown after `yield`, following `tests/integration/test_api_facts.py:38`.
- [ ] Verify the no-keys path: with no `ansible_facts*` key present, `check_ansible_facts(max_age=300)` emits a WARNING containing "No Ansible facts found in Redis cache", captured with `loguru_logs`.
- [ ] Verify the no-keys test skips instead of deleting: when `utils.redis.scan_iter(match="ansible_facts*")` finds pre-existing keys, the test calls `pytest.skip` with a reason naming the leftover keys, and no key is deleted.
- [ ] Verify the fresh path: after seeding one host with a current epoch (string form) and calling `check_ansible_facts(max_age=300)`, no WARNING record contains that hostname.
- [ ] Verify the stale path: after seeding one stale host (epoch 9999 seconds old) and one fresh host, `check_ansible_facts(max_age=300)` emits a WARNING containing `Host '<stale-host>': facts are` and `seconds old`, a WARNING containing "Run 'osism sync facts' to update facts.", and no WARNING containing the fresh hostname.
- [ ] Verify the missing-epoch path: after seeding a host whose JSON lacks `ansible_date_time.epoch`, the call emits a DEBUG record containing "facts missing ansible_date_time.epoch" and the hostname, emits no WARNING containing the hostname, and completes without raising.
- [ ] Verify the malformed-JSON path: after seeding the value `{ not valid json`, `json.JSONDecodeError` is caught inside `check_ansible_facts` and the call emits a DEBUG record starting with "Skipping malformed ansible_facts entry", emits no WARNING containing the hostname, and completes without raising.
- [ ] Verify the empty-value path: after seeding a host with the empty string, the call emits no WARNING and no DEBUG containing the hostname and completes without raising.
- [ ] Confirm every log assertion is scoped to a seeded hostname or to a message unique to its scenario, so pre-existing keys in a shared local Redis cannot fail the fresh, stale, missing-epoch, malformed-JSON, or empty-value tests.
- [ ] Run `pytest tests/integration/test_facts.py` twice in a row against the same live Redis: both runs pass, which shows the teardown removed every seeded key.
- [ ] Run `pytest tests/integration/test_facts.py` without a reachable Redis: every test is reported as skipped with the reason from `tests/integration/conftest.py:70`.
- [ ] Confirm the `python-osism-integration-tests` CI job executes the new tests (not skipped) and passes.

## Non-Goals

- Changing `check_ansible_facts()` or any other source code; this issue is test-only. The draft cites the function at line 563, which has since moved to `osism/utils/__init__.py:596`; the citation is corrected here, the function is untouched.
- Covering the Redis-unreachable branch (`osism/utils/__init__.py:621`) in the integration suite: forcing a `SCAN` failure would mean stopping the shared Redis mid-session and would break the other integration tests. `tests/unit/utils/test_init_task_output.py:461` covers it with a mock.
- Moving `loguru_logs` or the seed fixture into a shared conftest: module-local fixtures mirror `tests/integration/test_api_facts.py` and keep the change to one file.
- Sibling test modules: the Redis semaphore is #2401 (pull request #2548), the vault password round-trip is #2402 (pull request #2547), Celery error, state and revoke paths are #2404, and the EventBridge pub/sub round-trip is #2405.
- Tests that need RabbitMQ, NetBox, or OpenStack services; #2400 excludes them from this round.
- Adding dependencies: Tier 1 in #2400 runs on the existing single-Redis setup; the `httpx` prerequisite belonged to the Tier 2 issues #2406 and #2407, both closed.

## References

- Meta issue #2400 (Tier 1 list) and #2368 (integration-test foundation)
- `osism/utils/__init__.py:596` (function under test), `osism/utils/__init__.py:25` (`LOCAL_FACT_HOSTS`)
- `osism/commands/apply.py:422` (production call site)
- `tests/unit/utils/test_init_task_output.py:450` (unit coverage of the same scenarios, mocked Redis)
- `tests/conftest.py:86` (`loguru_logs` fixture)
- `tests/integration/conftest.py` (skip logic and session fixtures)
- `tests/integration/test_api_facts.py` (seeding convention, from #2406 / pull request #2411)
- `playbooks/test-integration.yml` (CI job), `setup.cfg:186` (marker registration)

---

berendt:
Closes #2403.

`check_ansible_facts()` (`osism/utils/__init__.py:596`) is what `osism apply` calls before running plays (`osism/commands/apply.py:422`) to warn operators about a missing or outdated facts cache. Its unit tests (`tests/unit/utils/test_init_task_output.py:450`) drive it through a `MagicMock` client, so the parts most likely to regress are the parts the mocks simulate: the cursor-based `SCAN` loop, the bytes decoding of keys and values, and JSON parsing of data a real server returned.

This adds `tests/integration/test_facts.py`, which exercises those paths against the live Redis the `python-osism-integration-tests` job provides. No source file changes.

## The commit

- `Add integration tests for the Ansible facts freshness check` — one new module with six tests and a module-local `seed_facts` fixture.

## What the tests cover

| Test | Seeded value | Checks |
|---|---|---|
| `test_no_facts_warns_about_empty_cache` | nothing | the "No Ansible facts found in Redis cache" warning |
| `test_fresh_facts_produce_no_warning` | current epoch | no warning names the host |
| `test_stale_facts_report_the_stale_host_only` | one host at `now-9999`, one fresh | the stale host is named with its age, the fresh one is not |
| `test_facts_without_epoch_are_skipped` | facts without `ansible_date_time.epoch` | the skip is logged at debug level, no warning |
| `test_malformed_json_is_skipped` | `{ not valid json` | `json.JSONDecodeError` is caught, the skip is logged at debug level |
| `test_empty_value_is_skipped` | empty string | the host is skipped silently |

Two properties keep the module deterministic against a Redis that is not empty. Hostnames are `itest-<uuid4>`, and every assertion is scoped to a seeded hostname or to a message unique to its scenario, so unrelated keys cannot fail the suite. `check_ansible_facts()` scans the whole database, so the empty-cache case cannot be produced on a Redis holding real facts; that test skips and names the leftover keys rather than deleting data the suite does not own.

`max_age=300` is passed explicitly everywhere, since the `settings.FACTS_MAX_AGE` default of 12 hours would tie the stale case to the environment.

## Verification

Against a throwaway `redis:7-alpine`:

- `pytest tests/integration/test_facts.py` — 6 passed, run twice in a row, with `KEYS 'ansible_facts*'` empty afterwards.
- With a foreign `ansible_facts` key present: the empty-cache test is reported as skipped, the key survives, and the other five tests stay green.
- Without Redis: 6 skipped; with `OSISM_REQUIRE_REDIS=1` the session exits non-zero.
- `black --check` and `flake8` are clean.

Three mutations of `check_ansible_facts()` were each caught by the matching test and then reverted: disabling the staleness comparison failed `test_stale_facts_report_the_stale_host_only`, downgrading the empty-cache warning failed `test_no_facts_warns_about_empty_cache`, and removing the missing-epoch debug line failed `test_facts_without_epoch_are_skipped`.

Part of #2400 (Tier 1).

### DIFF
--- a/tests/integration/test_facts.py
+++ b/tests/integration/test_facts.py
@@ -1,0 +1,189 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Integration tests for ``osism.utils.check_ansible_facts()``.
+
+The function scans Redis for ``ansible_facts*`` keys with a cursor-based
+``SCAN`` loop and reads ``ansible_date_time.epoch`` from each JSON value to
+decide whether the cached facts are stale, reporting the outcome through
+loguru. The unit tests cover the same scenarios against a ``MagicMock``
+client; running them against a live Redis exercises what the mocks stand in
+for: the cursor loop, keys and values arriving as ``bytes``, and JSON that
+round-tripped through a Redis server. The suite is skipped automatically when
+Redis is not reachable (see ``conftest.py``).
+"""
+
+import json
+import time
+import uuid
+
+import pytest
+
+from osism import utils
+
+pytestmark = pytest.mark.integration
+
+# Passed explicitly to every call: the ``settings.FACTS_MAX_AGE`` default is 12
+# hours, which would tie the stale case to the environment.
+MAX_AGE = 300
+
+
+@pytest.fixture
+def seed_facts():
+    """Seed ``ansible_facts<host>`` keys and remove them after the test."""
+    keys = []
+
+    def _seed(host, value):
+        key = f"ansible_facts{host}"
+        utils.redis.set(key, value)
+        keys.append(key)
+
+    yield _seed
+
+    for key in keys:
+        utils.redis.delete(key)
+
+
+def _host():
+    """Return a hostname unique to one test.
+
+    The ``itest-`` prefix keeps it clear of ``LOCAL_FACT_HOSTS``, whose facts
+    ``check_ansible_facts()`` skips outright.
+    """
+    return f"itest-{uuid.uuid4()}"
+
+
+def _facts(epoch):
+    """Return a facts blob carrying ``epoch``.
+
+    Ansible stores the epoch as a string, and ``check_ansible_facts()`` casts
+    it with ``float()``, so the string form is what the test seeds.
+    """
+    return json.dumps({"ansible_date_time": {"epoch": str(int(epoch))}})
+
+
+def _messages(records, level, needle):
+    """Return the captured messages at ``level`` that contain ``needle``."""
+    return [
+        record["message"]
+        for record in records
+        if record["level"] == level and needle in record["message"]
+    ]
+
+
+def test_no_facts_warns_about_empty_cache(loguru_logs):
+    """An empty facts cache is reported as such.
+
+    ``check_ansible_facts()`` scans the whole database, so a Redis holding
+    real facts cannot produce this case. Skip there rather than deleting data
+    that does not belong to the suite; the CI container starts empty.
+    """
+    leftover = sorted(
+        key.decode() for key in utils.redis.scan_iter(match="ansible_facts*")
+    )
+    if leftover:
+        pytest.skip(f"Redis already holds ansible_facts keys: {leftover}")
+
+    utils.check_ansible_facts(max_age=MAX_AGE)
+
+    assert _messages(loguru_logs, "WARNING", "No Ansible facts found in Redis cache")
+
+
+def test_fresh_facts_produce_no_warning(seed_facts, loguru_logs):
+    """A host whose facts are current is not reported."""
+    host = _host()
+    seed_facts(host, _facts(time.time()))
+
+    utils.check_ansible_facts(max_age=MAX_AGE)
+
+    assert not _messages(loguru_logs, "WARNING", host)
+
+
+def test_stale_facts_report_the_stale_host_only(seed_facts, loguru_logs):
+    """Facts older than ``max_age`` are reported, fresh ones are not."""
+    stale_host = _host()
+    fresh_host = _host()
+    seed_facts(stale_host, _facts(time.time() - 9999))
+    seed_facts(fresh_host, _facts(time.time()))
+
+    utils.check_ansible_facts(max_age=MAX_AGE)
+
+    stale_messages = _messages(
+        loguru_logs, "WARNING", f"Host '{stale_host}': facts are"
+    )
+    assert stale_messages
+    assert all("seconds old" in message for message in stale_messages)
+    assert _messages(loguru_logs, "WARNING", "Run 'osism sync facts' to update facts.")
+    assert not _messages(loguru_logs, "WARNING", fresh_host)
+
+
+def test_stale_facts_are_reported_across_scan_batches(seed_facts, loguru_logs):
+    """Every stale host is reported when ``SCAN`` needs more than one call.
+
+    ``check_ansible_facts()`` collects the keys of every batch before reading
+    them. The other tests seed too few keys to leave the first batch, so
+    replacing the loop with a single ``SCAN`` call keeps them green; only
+    enough hosts to force a second call covers it.
+    """
+    hosts = [_host() for _ in range(200)]
+    for host in hosts:
+        seed_facts(host, _facts(time.time() - 9999))
+
+    # COUNT is a hint, not a limit, so the number of keys one call covers is
+    # not guaranteed. Fail loudly if a single batch swallows the keyspace
+    # instead of passing without having exercised pagination.
+    cursor, _ = utils.redis.scan(0, match="ansible_facts*", count=100)
+    assert cursor != 0
+
+    utils.check_ansible_facts(max_age=MAX_AGE)
+
+    # One line per host proves the batches accumulated; exactly one line
+    # proves they were not counted twice, which matters because SCAN may
+    # return a key more than once and the collected keys are not deduplicated.
+    for host in hosts:
+        assert len(_messages(loguru_logs, "WARNING", f"Host '{host}': facts are")) == 1
+
+
+def test_facts_without_epoch_are_skipped(seed_facts, loguru_logs):
+    """Facts lacking ``ansible_date_time.epoch`` are skipped, not reported."""
+    host = _host()
+    seed_facts(host, json.dumps({"ansible_hostname": "node-1"}))
+
+    utils.check_ansible_facts(max_age=MAX_AGE)
+
+    assert _messages(
+        loguru_logs, "DEBUG", f"Host '{host}': facts missing ansible_date_time.epoch"
+    )
+    assert not _messages(loguru_logs, "WARNING", host)
+
+
+def test_malformed_json_is_skipped(seed_facts, loguru_logs):
+    """A value that is not JSON is skipped without failing the check."""
+    host = _host()
+    seed_facts(host, "{ not valid json")
+
+    utils.check_ansible_facts(max_age=MAX_AGE)
+
+    skipped = _messages(loguru_logs, "DEBUG", "Skipping malformed ansible_facts entry")
+    assert any(host in message for message in skipped)
+    assert not _messages(loguru_logs, "WARNING", host)
+
+
+def test_empty_value_is_skipped(seed_facts, loguru_logs):
+    """An empty value is skipped silently and the scan continues.
+
+    Absence of output says nothing about what happened to the remaining keys,
+    so stale hosts are seeded alongside: turning the ``continue`` into a
+    ``return`` or ``break`` drops their report and fails the test.
+    """
+    host = _host()
+    seed_facts(host, "")
+    stale_hosts = [_host() for _ in range(5)]
+    for stale_host in stale_hosts:
+        seed_facts(stale_host, _facts(time.time() - 9999))
+
+    utils.check_ansible_facts(max_age=MAX_AGE)
+
+    assert not _messages(loguru_logs, "WARNING", host)
+    assert not _messages(loguru_logs, "DEBUG", host)
+    for stale_host in stale_hosts:
+        assert _messages(loguru_logs, "WARNING", f"Host '{stale_host}': facts are")
--- a/osism/utils/__init__.py
+++ b/osism/utils/__init__.py
@@ -661,11 +661,12 @@ def check_ansible_facts(max_age=None):
             truncated_value = data
             if isinstance(truncated_value, (bytes, str)):
                 truncated_value = truncated_value[:200]
-            logger.debug(
-                "Skipping malformed ansible_facts entry for key %r: %r",
-                key,
-                truncated_value,
-                exc_info=True,
+            # loguru formats with str.format and knows no exc_info kwarg, so
+            # printf-style placeholders would be logged verbatim and the
+            # traceback dropped.
+            logger.opt(exception=True).debug(
+                f"Skipping malformed ansible_facts entry for key {key!r}: "
+                f"{truncated_value!r}"
             )
             continue
 


tests/integration/test_facts.py
```diff
@@ -0,0 +1,156 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Integration tests for ``osism.utils.check_ansible_facts()``.
+
+The function scans Redis for ``ansible_facts*`` keys with a cursor-based
+``SCAN`` loop and reads ``ansible_date_time.epoch`` from each JSON value to
+decide whether the cached facts are stale, reporting the outcome through
+loguru. The unit tests cover the same scenarios against a ``MagicMock``
+client; running them against a live Redis exercises what the mocks stand in
+for: the cursor loop, keys and values arriving as ``bytes``, and JSON that
```
ideaship, line 10: The cursor loop is listed first among the things a live Redis exercises, and it is the one path this file does not reach. `check_ansible_facts()` calls `r.scan(cursor, match="ansible_facts*", count=100)` (`osism/utils/__init__.py:617`), and the largest scenario here seeds two keys. Measured against `redis:7-alpine` on an empty database:

| keys seeded | first cursor | loop iterations |
|---|---|---|
| 2 | 0 | 1 |
| 50 | 0 | 1 |
| 64 | 0 | 1 |
| 100 | 127 | 2 |
| 200 | 193 | 2 |
| 300 | 138 | 3 |

So on the container the job provides, the `while True` body runs exactly once and `keys.extend(batch)` never accumulates. Worth noting the threshold is well above "seed a few more": 64 keys still completes in one call, because `SCAN` walks hash buckets rather than matching keys and `count=100` covers the whole table until it grows past that.

The consequence is concrete. Refactoring the loop to a single call — `cursor, keys = r.scan(0, match="ansible_facts*", count=100)`, a plausible simplification — leaves this file entirely green: 20 of 20 runs pass at two keys. A 200-host version fails 20 of 20, reporting only 100–102 of the 200 stale hosts. That is the regression the docstring implies is covered, and the mocked pagination test at `tests/unit/utils/test_init_task_output.py:639-658` is what currently catches it.

Either exercise it or drop "the cursor loop" from the list; the bytes and round-tripped-JSON claims are real and worth keeping. I would exercise it, for a reason beyond the refactor above: Redis documents that `SCAN` may return the same element more than once while the keyspace rehashes, and `osism/utils/__init__.py:614-620` extends a flat list with no dedup — which would inflate the count at `:673-676` and print a host's warning twice. A mocked test that hard-codes distinct keys per batch encodes exactly-once semantics that live `SCAN` does not promise, so that class of defect is only reachable here.

If you exercise it, a behavioural assertion beats counting `scan()` calls — a call-count spy means wrapping the live client, which reintroduces the mock this file exists to escape:

```python
def test_stale_facts_across_scan_batches(seed_facts, loguru_logs):
    hosts = [_host() for _ in range(200)]
    for host in hosts:
        seed_facts(host, _facts(time.time() - 9999))

    # Precondition: a single SCAN cannot cover this keyspace, so the loop in
    # check_ansible_facts() must iterate. COUNT is documented as a hint, so this
    # fails loudly if a future Redis makes 200 keys insufficient rather than
    # silently passing without testing pagination.
    cursor, _ = utils.redis.scan(0, match="ansible_facts*", count=100)
    assert cursor != 0

    utils.check_ansible_facts(max_age=MAX_AGE)

    assert _messages(loguru_logs, "WARNING", f"stale for {len(hosts)} host(s)")
    for host in hosts:
        assert _messages(loguru_logs, "WARNING", f"Host '{host}': facts are")
```

Asserting every host is what proves the batches accumulated. The commit message carries the same claim about the SCAN loop and would want the same correction.

tests/integration/test_facts.py
```diff
@@ -0,0 +1,156 @@
...
+    assert _messages(
+        loguru_logs, "DEBUG", f"Host '{host}': facts missing ansible_date_time.epoch"
+    )
+    assert not _messages(loguru_logs, "WARNING", host)
+
+
+def test_malformed_json_is_skipped(seed_facts, loguru_logs):
+    """A value that is not JSON is skipped without failing the check.
+
+    The debug call uses printf-style placeholders that loguru's ``str.format``
```
ideaship: This docstring is accurate about the behaviour, but it is worth fixing rather than documenting. loguru formats with `str.format`, so the template at `osism/utils/__init__.py:664-669` discards `key` and `truncated_value` entirely — the recorded message is the literal `Skipping malformed ansible_facts entry for key %r: %r`. `exc_info=True` is inert too: loguru has no such kwarg, so it is swallowed as a format argument and `record["exception"]` stays `None`. Verified against loguru 0.7.3. An operator hitting corrupt cached facts gets a message naming neither the host, nor the payload, nor the parse error — on the one path where all three matter.

That also makes line 144 weaker than it reads. With no host in the message, *any* malformed `ansible_facts*` entry in a shared Redis satisfies it: this test's own key could never be processed and the assertion would still pass. It is also the exception to the commit message's claim that "every assertion is scoped to a seeded hostname or to a message unique to its scenario" — that sentence wants qualifying either way.

One line in `osism/utils/__init__.py`:

```python
logger.opt(exception=True).debug(
    f"Skipping malformed ansible_facts entry for key {key!r}: {truncated_value!r}"
)
```

then this paragraph goes away and line 144 becomes `assert _messages(loguru_logs, "DEBUG", host)`. Confirmed that the fix produces `Skipping malformed ansible_facts entry for key b'ansible_factsitest-…': b'{ not valid json'` with a `JSONDecodeError` attached.

If you would rather keep the diff tests-only, that splits cleanly — with the fix applied, this file and the affected unit modules stay green (12 and 77 tests), because every assertion is a substring match on a prefix the interpolated message retains. But please drop this paragraph in this PR regardless: nothing will turn red to signal that prose describing placeholders is stale once they are gone. The same pattern exists at `osism/commands/console.py:124` and would want the same treatment.

tests/integration/test_facts.py
```diff
@@ -0,0 +1,156 @@
...
+
+def test_empty_value_is_skipped(seed_facts, loguru_logs):
+    """An empty value is skipped silently."""
+    host = _host()
+    seed_facts(host, "")
+
+    utils.check_ansible_facts(max_age=MAX_AGE)
+
+    assert not _messages(loguru_logs, "WARNING", host)
+    assert not _messages(loguru_logs, "DEBUG", host)
```
ideaship, line 187: Both assertions here are absence-only and the seeded input produces no log output at all, so nothing distinguishes "skipped the empty entry and kept scanning" from "skipped it and stopped". Line 156 in particular has almost no failure mode: the only reachable DEBUG message containing a hostname is the missing-epoch one at `osism/utils/__init__.py:653-655`, which an empty value never reaches, so even removing `if not data: continue` — making `json.loads("")` raise — leaves this green.

Measured over 200 trials per case, replacing that `continue` with an early exit:

| test | `continue` → `return` | `continue` → `break` |
|---|---|---|
| as written (one empty key) | 0/200 caught | 0/200 caught |
| co-seeding 1 stale host | 200/200 | 78/200 |
| co-seeding 5 stale hosts | 200/200 | 173/200 |

Co-seeding a stale host and asserting it is still reported costs two lines and catches the early-`return` case deterministically — order-independently, because stale hosts are reported after the loop, so an early return skips the summary regardless of which key `SCAN` returns first. Five stale hosts also covers most of the `break` case. The correct implementation passed 200/200 in every configuration, so there is no flakiness cost.

tests/integration/test_facts.py
```diff
@@ -0,0 +1,156 @@
...
+pytestmark = pytest.mark.integration
+
+# Passed explicitly to every call: the ``settings.FACTS_MAX_AGE`` default is 12
+# hours, which would tie the stale case to the environment.
+MAX_AGE = 300
+
+
+@pytest.fixture
+def seed_facts():
+    """Seed ``ansible_facts<host>`` keys and remove them after the test."""
```
ideaship, line 32: This fixture is byte-for-byte identical to `tests/integration/test_api_facts.py:38-51`, docstring included. Both build `f"ansible_facts{host}"` — the key format the two suites exist to pin, one for `check_ansible_facts()`, one for the inventory-facts endpoints — so changing that format and updating only one copy leaves the other seeding the old layout and still passing.

Nothing to change in this PR. In a commit of its own afterwards: move `seed_facts` into the existing `tests/integration/conftest.py` and delete both copies, and add a unique-host fixture there to replace `_host()` here and the six inline `f"itest-{uuid.uuid4()}"` in `test_api_facts.py`. That one wants to be a factory rather than a plain string fixture, since `test_stale_facts_report_the_stale_host_only` needs two hosts in one test.

ideaship (changes requested):

ideaship (approved):

PR #2550, opened 2026-08-03
Add integration tests for Celery error, state and revoke paths

### LINKED ISSUE #2404: Integration test: Celery error, state and revoke paths

## Description

Extend the Celery integration coverage beyond the existing noop round-trip with three lifecycle tests: exception propagation through the result backend, the `STARTED` state transition, and revocation of a running task. All three run against the live Redis broker and the real worker subprocess that the existing session fixtures provide.

The repository walk settles the constraint the issue flagged for the revoke test, and widens it: none of the three tasks registered on the `ansible` app (`gather_facts`, `run`, `noop` in `osism/tasks/ansible.py`) raises deliberately or blocks, so every one of the three scenarios needs a worker-registered test-only task, not just revoke. Per the author's decision during elaboration, these tasks live in a new module under `tests/integration/`, not in the shipped package.

Concrete boundaries:

1. Worker fixture. Already exists: the `celery_worker` fixture (`tests/integration/conftest.py:97`) starts a real worker subprocess (`celery -A osism.tasks.ansible worker -n ci-worker@%h -Q osism-ansible -c 1`) with `GATHER_FACTS_SCHEDULE=0`, gates readiness on a ping answered by `ci-worker@...`, and raises `RuntimeError("Celery worker exited early with code ...")` when the process dies during boot. This issue adds two things to that `subprocess.Popen` call: `--include=tests.integration.worker_tasks` on the celery command line, and an explicit `cwd` pointing at the repository root, derived as `Path(__file__).resolve().parents[2]`, so the worker resolves the module against the checkout regardless of where pytest was launched. A broken import path makes the worker exit at boot and surfaces through the existing early-exit `RuntimeError`, not as a hang.

2. Test-only tasks. Already exists: `osism/tasks/ansible.py` registers `gather_facts`, `run` and `noop`; `noop` returns immediately, and the two Ansible runners neither raise nor block in a controlled way (`run` calls `utils.check_task_lock_and_exit()`, which on a set lock calls `exit(1)`; the resulting `SystemExit` would kill the pool child rather than produce a clean `FAILURE`). This issue adds `tests/integration/worker_tasks.py`, which imports `app` from `osism.tasks.ansible` and registers two tasks with explicit names:
   - `itest_fail(message)`, named `osism.tasks.ansible.itest_fail`, raises `RuntimeError(message)`.
   - `itest_block(seconds)`, named `osism.tasks.ansible.itest_block`, calls `time.sleep(seconds)` and returns `seconds`.

   The explicit names matter: `Config.task_routes` (`osism/tasks/__init__.py:34`) routes by the pattern `osism.tasks.ansible.*`. An auto-generated name (`tests.integration.worker_tasks.itest_fail`) would land on the `default` queue, which the worker does not consume, and `result.get()` would time out. `RuntimeError` is a builtin, so the client-side `get()` re-raises the original exception class after deserialization from the Redis result backend.

3. Lifecycle tests. Already exists: `tests/integration/test_celery.py` covers the noop round-trip and worker visibility. This issue adds the new module `tests/integration/test_celery_lifecycle.py` (the file name the issue offered) with the SPDX header, a module docstring, `pytestmark = pytest.mark.integration`, a polling helper, and the three tests described below. The tests import the task signatures with `from tests.integration.worker_tasks import itest_block, itest_fail` (the `tests` package has `__init__.py` files at both levels, so the absolute import works under pytest). `test_celery.py` is not modified.

4. STARTED tracking. Already exists: `task_track_started = True` in `Config` (`osism/tasks/__init__.py:33`). This issue only observes it; no configuration changes.

5. Revocation. Already exists: `osism.utils.revoke_task` (`osism/utils/__init__.py:454`; the issue's `:421` predates unrelated changes to the file) builds a fresh Celery app from `Config` and calls `app.control.revoke(task_id, terminate=True)`, returning `True` on success and `False` when that raises. This issue exercises the `True` path against the live broker.

The three tests, concretely:

- `test_failing_task_propagates_exception(celery_worker)`: dispatch `itest_fail` with a unique marker message (`f"itest-fail-{uuid.uuid4()}"`, following the `uuid` pattern in `tests/integration/test_locking.py`); wrap `result.get(timeout=60)` in `pytest.raises(RuntimeError)` and assert the marker message is preserved in the raised exception; afterwards assert `result.failed() is True` and `result.state == "FAILURE"`.
- `test_started_state_is_reported(celery_worker)`: dispatch `itest_block(3)`; poll with the helper until the result reaches one of `{"STARTED", "SUCCESS", "FAILURE"}` and assert the observed state is `"STARTED"` (with a 3 s task body and a 0.1 s poll interval the intermediate state cannot be missed); then assert `result.get(timeout=60) == 3` and `result.state == "SUCCESS"`.
- `test_revoke_running_task(celery_worker)`: dispatch `itest_block(300)`; wait until the state is `"STARTED"`, because revoking a still-queued task would only discard it from the queue and never exercise `terminate=True` against a running pool child; assert `utils.revoke_task(result.id) is True`; poll until the state reaches `"REVOKED"`; assert that `result.get(timeout=10)` raises `celery.exceptions.TaskRevokedError`; finally dispatch `ansible.noop.delay()` and assert `.get(timeout=60) is True`, which proves the prefork pool respawned the terminated child and the session-scoped worker stays usable for tests that run afterwards.

The polling helper `_wait_for_state(result, states, timeout=30)` loops on `result.state` against a `time.monotonic()` deadline with a 0.1 s sleep between reads, returns the first state contained in `states`, and calls `pytest.fail(...)` naming the task id, the expected states and the last observed state when the deadline passes. Every wait in the module goes through this helper or through `result.get(timeout=...)`; the module contains no unbounded blocking call.

No other files change: the Zuul job (`playbooks/test-integration.yml`) already runs `pytest tests/integration`, the `integration` marker is already declared under `[tool:pytest]` in `setup.cfg`, and no new dependency is needed. `celery==5.6.3` (pinned in `Pipfile`) provides both the worker `--include` option and `celery.exceptions.TaskRevokedError`; both were checked against the installed version.

## Motivation

The operator-facing flows in this repository rely on exactly the three result-backend behaviours that have no test today. `handle_task` (`osism/tasks/__init__.py:422`) drives waiting and the CTRL+C revoke prompt, which calls `utils.revoke_task` (`osism/tasks/__init__.py:459`), and `task_track_started` is what makes a running task distinguishable from a queued one. The integration tests introduced by #2368 prove only the happy path: a noop task returns `True` through the result backend.

A regression in any of these paths (an exception swallowed by the result backend, `STARTED` no longer reported, revocation silently failing) would reach production unnoticed today. After this issue, the `python-osism-integration-tests` Zuul job fails instead.

## Affected Areas

- `tests/integration/worker_tasks.py` (new: `itest_fail` and `itest_block` registered on the `ansible` app under `osism.tasks.ansible.itest_*` names)
- `tests/integration/conftest.py` (the `celery_worker` fixture gains `--include=tests.integration.worker_tasks` and an explicit `cwd` at the repository root)
- `tests/integration/test_celery_lifecycle.py` (new: `_wait_for_state` helper and the three lifecycle tests)

## Acceptance Criteria

- [ ] Define `itest_fail(message)` (raises `RuntimeError(message)`) and `itest_block(seconds)` (sleeps, returns `seconds`) in `tests/integration/worker_tasks.py`, registered on the app imported from `osism.tasks.ansible` with the explicit names `osism.tasks.ansible.itest_fail` and `osism.tasks.ansible.itest_block`.
- [ ] Start the worker with `--include=tests.integration.worker_tasks` and `cwd` set to the repository root in the `celery_worker` fixture; the fixture's existing readiness gate passes within its 60 s deadline.
- [ ] Verify failure propagation: `result.get(timeout=60)` on a dispatched `itest_fail` raises `RuntimeError` carrying the dispatched marker message, and afterwards `result.failed()` is `True` and `result.state` is `"FAILURE"`.
- [ ] Verify the STARTED transition: after dispatching `itest_block(3)`, polling observes `"STARTED"`, then `result.get(timeout=60)` returns `3` and `result.state` is `"SUCCESS"`.
- [ ] Verify revocation: after a dispatched `itest_block(300)` reaches `"STARTED"`, `utils.revoke_task(result.id)` returns `True`, polling observes `"REVOKED"` within 30 s, and `result.get(timeout=10)` raises `celery.exceptions.TaskRevokedError`.
- [ ] Call `revoke_task` only after `"STARTED"` was observed, never on a still-queued task.
- [ ] Verify worker survival: at the end of the revoke test, `ansible.noop.delay().get(timeout=60)` returns `True`.
- [ ] Fail instead of hanging when a state is never reached: `_wait_for_state` calls `pytest.fail` naming the task id and the last observed state once its `time.monotonic()` deadline (default 30 s) passes.
- [ ] Mark `tests/integration/test_celery_lifecycle.py` with `pytestmark = pytest.mark.integration` and the `# SPDX-License-Identifier: Apache-2.0` header (also on `worker_tasks.py`); without a reachable Redis, `pytest tests/integration` reports the three new tests as skipped, and with `OSISM_REQUIRE_REDIS=1` the session fails (existing conftest behaviour, unchanged).
- [ ] Keep all changes inside `tests/integration/`; the pull request diff contains no file outside that directory.
- [ ] Pass the `python-osism-integration-tests` Zuul job.

## Non-Goals

- No test tasks in `osism/tasks/ansible.py` or anywhere else in the shipped package; the author decided during elaboration that they live under `tests/integration/`.
- No coverage of the `revoke_task` failure path (`False` when `app.control.revoke` raises, `osism/utils/__init__.py:472`): with the CI broker up there is no honest way to trigger it from an integration test.
- No tests requiring RabbitMQ, NetBox or OpenStack services; #2400 excludes them from this test round.
- No EventBridge pub/sub coverage, which is #2405's scope; RedisSemaphore, vault password and facts freshness are carried by the sibling issues #2401, #2402 and #2403.

## References

- Meta issue #2400 (test-round framing and tier definitions); test foundation laid in #2368.
- `tests/integration/conftest.py` (skip logic, `celery_app` / `celery_worker` fixtures).
- `tests/integration/test_celery.py` (existing noop round-trip and worker-visibility tests).
- `osism/tasks/__init__.py` (`Config` with broker URL, `task_track_started`, `task_routes`; `handle_task` revoke prompt).
- `osism/tasks/ansible.py` (the `ansible` Celery app and its three production tasks).
- `osism/utils/__init__.py` (`revoke_task`).
- `playbooks/test-integration.yml` and `.zuul.yaml` (definition and check/gate wiring of `python-osism-integration-tests`).
- Celery 5.6 documentation: task states and `task_track_started` (https://docs.celeryq.dev/en/stable/userguide/tasks.html), revoking workers' tasks (https://docs.celeryq.dev/en/stable/userguide/workers.html).

**Executability score:** 9/10

---

berendt:
Extends the Celery integration coverage beyond the noop round-trip with three
lifecycle tests against the live Redis broker and the real worker subprocess.

One commit, e240b64:

- `tests/integration/worker_tasks.py` (new) registers two test-only tasks on
  the `ansible` app: `itest_fail` raises `RuntimeError`, `itest_block` sleeps
  and returns its argument. The explicit `osism.tasks.ansible.*` names route
  them onto the `osism-ansible` queue the worker consumes; with their
  auto-generated names they would land on the unconsumed `default` queue.
- `tests/integration/conftest.py` starts the session worker with
  `--include=tests.integration.worker_tasks` and pins its `cwd` to the
  repository root so the module resolves regardless of where pytest was
  launched. A broken include surfaces as the fixture's existing early-exit
  `RuntimeError` instead of a hang.
- `tests/integration/test_celery_lifecycle.py` (new) adds the three tests: an
  exception raised in the worker reaches the caller through
  `AsyncResult.get()` and leaves the task in `FAILURE`; `task_track_started`
  reports `STARTED` while a task runs; `osism.utils.revoke_task` terminates a
  running task, the state ends `REVOKED`, `get()` raises `TaskRevokedError`,
  and a final noop round-trip proves the prefork pool respawned its child.
  All waits go through a `time.monotonic()` deadline helper that fails the
  test instead of hanging.

The tests carry `pytestmark = pytest.mark.integration`, skip without a
reachable Redis, and run for real in the `python-osism-integration-tests`
job. `black --check` and `flake8` pass locally.

Closes #2404

### DIFF
--- a/tests/integration/conftest.py
+++ b/tests/integration/conftest.py
@@ -12,6 +12,7 @@ green outside the dedicated CI job.
 import os
 import subprocess
 import time
+from pathlib import Path
 
 import pytest
 
@@ -29,6 +30,24 @@ WORKER_NAME = "ci-worker@%h"
 # to gate readiness on *this* worker rather than any worker on the broker.
 WORKER_NODE_PREFIX = WORKER_NAME.split("@", 1)[0] + "@"
 WORKER_BOOT_TIMEOUT = 60
+# Test-only tasks (see that module's docstring for why they exist). The worker
+# imports it relative to its cwd, which is pinned to the repository root below.
+WORKER_INCLUDE = "tests.integration.worker_tasks"
+
+
+def _find_repo_root():
+    """Return the repository root, found by its ``setup.cfg`` marker.
+
+    Walking up from this file instead of hard-coding ``parents[2]`` fails
+    loudly here rather than as an opaque import error at worker boot.
+    """
+    for parent in Path(__file__).resolve().parents:
+        if (parent / "setup.cfg").exists():
+            return parent
+    raise RuntimeError("no repository root with setup.cfg above conftest.py")
+
+
+REPO_ROOT = _find_repo_root()
 
 
 def _require_redis():
@@ -100,7 +119,10 @@ def celery_worker(celery_app):
     The worker runs from the same virtualenv as the tests.
     ``GATHER_FACTS_SCHEDULE=0`` prevents registration of the periodic
     ``gather_facts`` task, which would try to run ``/run.sh`` in containers that
-    do not exist in CI.
+    do not exist in CI. ``--include`` registers the test-only lifecycle tasks
+    from ``tests/integration/worker_tasks.py``; ``cwd`` is pinned to the
+    repository root so that import resolves regardless of where pytest was
+    launched. A broken include surfaces as an early worker exit below.
     """
     proc = subprocess.Popen(
         [
@@ -114,8 +136,11 @@ def celery_worker(celery_app):
             WORKER_QUEUE,
             "-c",
             "1",
+            "--include",
+            WORKER_INCLUDE,
         ],
         env={**os.environ, "GATHER_FACTS_SCHEDULE": "0"},
+        cwd=REPO_ROOT,
     )
 
     try:
--- a/tests/integration/test_celery_lifecycle.py
+++ b/tests/integration/test_celery_lifecycle.py
@@ -1,0 +1,113 @@
+# SPDX-License-Identifier: Apache-2.0
+
+"""Celery task-lifecycle integration tests: failure, STARTED and revocation.
+
+These extend the noop round-trip in ``test_celery.py`` with the result-backend
+behaviour the operator-facing flows rely on: an exception raised in the worker
+reaches the caller through ``AsyncResult.get()``, ``task_track_started``
+reports ``STARTED`` while a task runs, and ``osism.utils.revoke_task``
+terminates a running task. The tasks they dispatch are registered on the
+worker via ``tests/integration/worker_tasks.py``.
+"""
+
+import time
+import uuid
+
+import pytest
+
+from osism import utils
+from tests.integration.worker_tasks import itest_block, itest_fail
+
+pytestmark = pytest.mark.integration
+
+# Timeouts in seconds, sized for slow CI machines: how long a dispatched task
+# may take to deliver its result, and how long a state transition may take to
+# become visible in the result backend.
+RESULT_TIMEOUT = 60
+STATE_TIMEOUT = 30
+# Task durations in seconds: long enough that polling cannot miss STARTED, and
+# far enough past the revoke that the revoke test never races its task's
+# natural end. ``itest_block`` carries its own hard time limit, so a test that
+# fails before revoking does not block the worker for the full duration.
+BLOCK_SHORT = 3
+BLOCK_LONG = 300
+
+
+def _wait_for_state(result, states, timeout=STATE_TIMEOUT):
+    """Poll ``result.state`` until it is one of ``states`` and return it.
+
+    Fails the test instead of hanging when the deadline passes.
+    """
+    deadline = time.monotonic() + timeout
+    state = result.state
+    while time.monotonic() < deadline:
+        state = result.state
+        if state in states:
+            return state
+        time.sleep(0.1)
+    pytest.fail(
+        f"task {result.id} did not reach {states} within {timeout}s "
+        f"(last state: {state})"
+    )
+
+
+def test_failing_task_propagates_exception(celery_worker):
+    """An exception raised in the worker reaches the caller via ``get()``.
+
+    ``RuntimeError`` is a builtin, so the result backend re-raises the
+    original exception class with the original message.
+    """
+    message = f"itest-fail-{uuid.uuid4()}"
+
+    result = itest_fail.delay(message)
+
+    with pytest.raises(RuntimeError, match=message):
+        result.get(timeout=RESULT_TIMEOUT)
+    assert result.failed() is True
+    assert result.state == "FAILURE"
+
+
+def test_started_state_is_reported(celery_worker):
+    """``task_track_started`` reports ``STARTED`` while a task runs.
+
+    The task body runs for ``BLOCK_SHORT`` seconds against a 0.1 s poll
+    interval, so a run that observes ``SUCCESS`` or ``FAILURE`` first either
+    never reported ``STARTED`` or lost the whole task duration to a stalled
+    poll loop.
+    """
+    result = itest_block.delay(BLOCK_SHORT)
+
+    state = _wait_for_state(result, {"STARTED", "SUCCESS", "FAILURE"})
+
+    assert state == "STARTED", (
+        f"got {state} first -- either task_track_started stopped working, "
+        f"or this poll loop stalled for the full {BLOCK_SHORT}s window"
+    )
+    assert result.get(timeout=RESULT_TIMEOUT) == BLOCK_SHORT
+    assert result.state == "SUCCESS"
+
+
+def test_revoke_running_task(celery_worker):
+    """``osism.utils.revoke_task`` terminates a running task.
+
+    The revoke is issued only after ``STARTED`` was observed: revoking a
+    still-queued task would merely discard it from the queue and never
+    exercise ``terminate=True`` against a running pool child.
+    """
+    from celery.exceptions import TaskRevokedError
+
+    from osism.tasks import ansible
+
+    result = itest_block.delay(BLOCK_LONG)
+    state = _wait_for_state(result, {"STARTED", "SUCCESS", "FAILURE"})
+    assert state == "STARTED"
+
+    assert utils.revoke_task(result.id) is True
+
+    assert _wait_for_state(result, {"REVOKED", "SUCCESS", "FAILURE"}) == "REVOKED"
+    with pytest.raises(TaskRevokedError):
+        result.get(timeout=RESULT_TIMEOUT)
+
+    # The prefork pool replaces the terminated child; prove the session-scoped
+    # worker stays usable for tests that run after this one.
+    assert ansible.noop.delay().get(timeout=RESULT_TIMEOUT) is True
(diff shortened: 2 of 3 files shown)


tests/integration/test_celery_lifecycle.py
```diff
@@ -0,0 +1,108 @@
...
+    The task body runs for ``BLOCK_SHORT`` seconds against a 0.1 s poll
+    interval, so the intermediate state cannot be missed; observing
+    ``SUCCESS`` or ``FAILURE`` first would mean ``STARTED`` was never
+    reported.
+    """
+    result = itest_block.delay(BLOCK_SHORT)
+
+    state = _wait_for_state(result, {"STARTED", "SUCCESS", "FAILURE"})
+
+    assert state == "STARTED"
```
ideaship: A bare `assert state == "STARTED"` misdirects the triage on the one run where it fires. There are two ways to observe a terminal state first — `task_track_started` stopped working, or this poll loop stalled for the full `BLOCK_SHORT` window — and the docstring above commits to the first ("would mean `STARTED` was never reported") while the second is what a loaded CI machine produces. Naming both costs one line:

```python
assert state == "STARTED", (
    f"got {state} first — either task_track_started stopped working, "
    f"or this poll loop stalled for the full {BLOCK_SHORT}s window"
)
```

The wait set itself is right and shouldn't change: admitting `SUCCESS`/`FAILURE` is what turns a miss into an immediate failure instead of a 30 s timeout. It's the docstring's categorical phrasing that wants the same qualification.

tests/integration/conftest.py
```diff
@@ -29,6 +30,25 @@
...
 WORKER_BOOT_TIMEOUT = 60
+# Test-only tasks (see that module's docstring for why they exist). The worker
+# imports it relative to its cwd, which is pinned to the repository root below.
+WORKER_INCLUDE = "tests.integration.worker_tasks"
+
+
+def _find_repo_root():
+    """Return the repository root, found by its ``setup.cfg`` marker.
+
+    Walking up from this file instead of hard-coding ``parents[2]`` keeps the
```
ideaship: "Keeps the lookup correct if this file moves" promises something `WORKER_INCLUDE = "tests.integration.worker_tasks"` at `:35` doesn't deliver. Move only `conftest.py` and the dotted path still resolves; move the whole `tests/integration/` directory — the more likely change — and the constant breaks while the walk survives. Either derive both from `__package__`, or trim the sentence to the failure-mode half, which holds unconditionally and is the part worth having: a located error here instead of an opaque `ImportError` at worker boot.

tests/integration/worker_tasks.py
```diff
@@ -0,0 +1,31 @@
...
+from osism.tasks.ansible import app
+
+
+@app.task(name="osism.tasks.ansible.itest_fail")
+def itest_fail(message):
+    raise RuntimeError(message)
+
+
+@app.task(name="osism.tasks.ansible.itest_block")
+def itest_block(seconds):
```
ideaship, line 37: Please add a `time_limit` here. If the first `_wait_for_state` in `test_revoke_running_task` times out before the revoke is issued, `pytest.fail()` leaves a 300 s sleep occupying the single-concurrency worker, and teardown burns its full 30 s warm-shutdown window before killing the process. Today that costs 30 s on an already-red run, because this test happens to collect last among the ones taking `celery_worker` — but that safety is invisible at the point where it matters, and it disappears the day someone adds a worker-using test file sorting after `test_celery_lifecycle.py`, which has no second slot to run in.

`BLOCK_LONG = 300` is the right value and shortening it would be the wrong fix — "never race the task's natural end" is the correct trade. The bound belongs on the task, not on the duration.

ideaship (changes requested):

ideaship (approved):

PR #2552, opened 2026-08-03
Add parameter to only erase metadata to baremetal clean

janhorstmann:
When specified the `erase_device_metadata` clean step will be executed, thus only erasing metadata on the disk.
On some hatdware devices secure erase will fail with an error, that prevents the fallback erasure of metadata from being executed. Adding this parameter will allow to trigger the metadata cleaning explicitly and thus provide a workround to clean these nodes, albeit in a less sophisticated and incomplete way.

### DIFF
--- a/osism/commands/baremetal.py
+++ b/osism/commands/baremetal.py
@@ -1184,6 +1184,12 @@ class BaremetalClean(Command):
             help="Clean given baremetal node when in provision state available",
         )
         parser.add_argument(
+            "--metadata-only",
+            default=False,
+            help="Only erase metadata on disks",
+            action="store_true",
+        )
+        parser.add_argument(
             "--all",
             default=False,
             help="Clean all baremetal nodes in provision state available",
@@ -1201,6 +1207,7 @@ class BaremetalClean(Command):
         cloud = parsed_args.cloud
         all_nodes = parsed_args.all
         name = parsed_args.name
+        metadata_only = parsed_args.metadata_only
         yes_i_really_really_mean_it = parsed_args.yes_i_really_really_mean_it
 
         if not all_nodes and not name:
@@ -1213,7 +1220,10 @@ class BaremetalClean(Command):
             )
             return 1
 
-        clean_steps = [{"interface": "deploy", "step": "erase_devices"}]
+        if metadata_only:
+            clean_steps = [{"interface": "deploy", "step": "erase_devices_metadata"}]
+        else:
+            clean_steps = [{"interface": "deploy", "step": "erase_devices"}]
 
         from osism.tasks.openstack import get_cloud_helpers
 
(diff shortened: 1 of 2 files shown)


osism/commands/baremetal.py
```diff
@@ -1213,7 +1220,10 @@ def take_action(self, parsed_args):
             )
             return 1
 
-        clean_steps = [{"interface": "deploy", "step": "erase_devices"}]
+        if metadata_only:
```
ideaship, line 1223: This doesn't fully hold to "Only erase metadata". Further down, `baremetal.py:1256-1259` unconditionally prepends `{"interface": "raid", "step": "delete_configuration"}` for any node whose `raid_interface` is not `no-raid`, so `osism baremetal clean --metadata-only NODE` on a hardware-RAID node emits `[raid delete_configuration, deploy erase_devices_metadata]` and tears down the logical disks in addition to wiping metadata. Ironic runs manual clean steps in the order supplied, so the RAID teardown happens first.

Since the commit message frames this as a workaround for hardware where secure erase errors out, that's exactly the node class where an operator would reach for the light-touch path and be surprised.

Two ways out: gate the prepend on `not metadata_only`, or keep the behavior and reword the help to say what actually happens (e.g. "Erase only disk metadata instead of a full device erase. RAID configuration is still deleted on nodes with a RAID interface."). Which one is your call, but the current pairing of contract and behavior shouldn't ship as-is.

osism/commands/baremetal.py
```diff
@@ -1183,6 +1183,12 @@ def get_parser(self, prog_name):
             type=str,
             help="Clean given baremetal node when in provision state available",
         )
+        parser.add_argument(
+            "--metadata-only",
+            default=False,
+            help="Only erase metadata on disks. ",
```
ideaship: Nit: trailing space in the help string, and the other options in this parser don't end their help with a period — `help="Only erase metadata on disks"` would match the siblings. (Moot if you end up rewording this per the other comment.)

ideaship (changes requested):
### Test coverage

Could you add coverage for the new flag? The harness is already in place at `tests/unit/commands/test_baremetal.py:1325-1358` — `_run_baremetal_clean` and the `ERASE_DEVICES_STEP` / `RAID_DELETE_STEP` constants — and the three sibling `test_clean_*` tests already assert exact `clean_steps` lists, so this is a near-copy of `test_clean_manageable_without_raid_interface`:

```python
ERASE_METADATA_STEP = {"interface": "deploy", "step": "erase_devices_metadata"}


def test_clean_metadata_only_uses_erase_devices_metadata():
    node = FakeNode(provision_state="manageable")
    conn = MagicMock()
    conn.baremetal.find_node.return_value = node
    _run_baremetal_clean(["node1", "--metadata-only"], conn)
    conn.baremetal.set_node_provision_state.assert_called_once_with(
        node.id, "clean", clean_steps=[ERASE_METADATA_STEP]
    )
```

A second test covering `--metadata-only` on a node with a RAID interface would also pin down whatever gets decided for the RAID question below, so the resolution is captured in executable form rather than only in review discussion.

### Commit message

The step is `erase_devices_metadata` (plural `devices`), but the body says `erase_device_metadata`, so a later grep for the actual step name won't find it. Also `hatdware` → `hardware` and `workround` → `workaround` while you're in there.

ideaship (approved):

Issue #2562, opened 2026-08-05
SONiC config-generation E2E golden test: context for the PR series


ideaship:
Context for a series of PRs that add end-to-end golden tests for the SONiC
config generator. Filed so each PR can stay small without a reviewer having to
reconstruct why it exists or what the ones around it do.

## What this adds, and the gap it fills

The SONiC config generator is 12 modules — `config_generator.py` alone is ~2.5k
lines — and its output is a complete `config_db.json` that gets deployed to a
switch. Its unit tests cover the individual helpers; before this series
**nothing exercised the whole path**: seeded NetBox → `sync_sonic()` → generated
config.

These PRs add that. NetBox is provisioned from a compose stack, seeded from
in-repo fixtures, and the exported `config_db.json` for every seeded device is
diffed against a committed golden. Nine golden devices end up covering all 38
`config_db` tables the generator can emit.

## Why golden files rather than more unit tests

Because each golden pins a device's **entire** emitted config, not one table. The
regressions that matter here are cross-cutting: a change to one helper silently
altering an unrelated table, or a table quietly disappearing from the output.
Unit tests assert what they were written to assert and stay green through both.

That is not hypothetical. Building this surfaced two production bugs the
existing unit tests did not catch, and in one case *could not* have caught — its
fixtures were shaped so the failing condition was unexpressible.
Both went out as separate PRs (see below), since neither is test
infrastructure.

## Design decisions worth knowing before reviewing

**NetBox comes from docker compose, not kind + Helm.** An earlier attempt at this
suite (see Superseded below) used a kind cluster. It was abandoned on measurement: ~1650s of a ~2100s
run was fixed setup — installing kind and Helm, then seeding — so the job used
80-92% of a 2400s budget as a matter of course and any node-local hiccup tipped
it over. Compose brings pre+run to ~1244s and removes python-osism's need for
kind/kubectl/helm entirely. A larger CI nodeset was also tried and dropped: 1235s
at 2 vCPU against 1335s at 8, so the bigger node was *slower*.

**Seed data is synthetic and in-repo, not derived from the testbed or the
netbox-manager example.** The motive is independence rather than speed: goldens
built from an input in another repository are hostage to it, and an unrelated
change there would rewrite them. Seeding also drops from 525s to 57.7s, and
coverage improves: the netbox-manager-derived data reached 31 of the 38 emitted
tables, the synthetic fixtures reach all 38.

## Scope of the coverage claim

[...]

PR #2566, opened 2026-08-05
Add synthetic base fixtures and the SONiC E2E job

ideaship:
Part of the series tracked in #2562, which explains the ordering and what each PR covers. Based on the preceding PR in the stack, so review only the top commits here.

**This is where the test first runs end to end, and the largest PR in the
series** — 3528 lines, of which roughly 2700 is generated golden JSON. Four
commits: five synthetic devices and their goldens, the coverage report, a README
section, and the Zuul job.

It is deliberately not split further. The harness is not independently testable
in smaller pieces, and the Zuul job has to land together with the goldens it
compares against, or the check is red on arrival.

**Two things to know before approving:**

Once this job exists it gates changes under `osism/tasks/conductor/`,
`osism/settings.py`, `Makefile` and `.zuul.yaml`. That makes this the point of
no return for the series.

The synthetic-fixture half **has never run in Zuul**. Only the compose change
has a green CI run, on a branch that no longer exists. The check on this PR is
the first real test of the fixtures, so it is the one to watch.

`coverage.py` reports 30 of 38 emitted tables here, and exits non-zero saying
so. That is the honest number for the base fixtures alone; the four scenario PRs
above take it to 38.

### DIFF
--- a/.zuul.yaml
+++ b/.zuul.yaml
@@ -140,6 +140,37 @@
     pre-run: playbooks/pre.yml
     run: playbooks/test-integration.yml
 
+# End-to-end golden test for the SONiC config generator: provisions NetBox
+# with a docker compose stack, seeds it from the in-repo fixtures under
+# tests/e2e/scenario/, runs sync_sonic() and compares the exported config_db
+# files against tests/e2e/golden/. In check the job only runs when files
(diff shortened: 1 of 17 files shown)


berendt:
@ideaship Please check the Sourcery review.

PR #2573, opened 2026-08-05
DO NOT MERGE: CI-only check for the full SONiC E2E series

ideaship:
**Do not review or merge. This PR exists only to get a CI signal, and will be
closed once it has one.**

The series in #2562 is stacked, so eight of its nine PRs are based on
`sonic-e2e-v2-*` branches rather than `main` — and Zuul does not enqueue those.
Only #2563, whose base is `main`, has been checked. That leaves the part of the
series most in need of a check unchecked, including the Zuul job definition the
series itself adds.

This branch is the same commit as the head of #2571, retargeted at `main` so the
whole series gets checked as one change. Nothing here is unique to this PR;
review happens in the stack.

## What this is actually testing

A local `make sonic-e2e` run exercises the harness script directly, so it
validates the fixtures, seeding, generation and golden comparison — but it says
nothing about the CI wiring, which is exactly the part that has never run. This
PR covers the rest:

- the `python-osism-sonic-e2e` job definition and its `files` matcher
- `playbooks/pre-sonic-e2e.yml`, including the `accept_ra=2` sysctl the
  IPv6-only node needs to keep its default route once forwarding is enabled
- `playbooks/test-sonic-e2e.yml` and the pipenv/`[ansible]` install it performs
- the `required-projects` pull of `netbox-manager`, which supplies the seeding
  CLI
- whether the job fits its 2400s timeout on a real CI node rather than a
  developer laptop

## Expected result

Green, with `python-osism-sonic-e2e` passing against the nine committed goldens.
The coverage report is not part of the job, so its exit status does not affect
this run.


ideaship:
Served its purpose — closing.

The full series passed as one change: all ten jobs green, with `python-osism-sonic-e2e` at **800s against its 2400s timeout**. That covered what a local harness run cannot — the job definition and files matcher, both playbooks including the `accept_ra=2` sysctl the IPv6-only node needs, the `required-projects` pull of netbox-manager, and all nine goldens.

For context on the timeout figure: the job this replaces consumed 80-92% of the same budget as a matter of course, which is why a node-local 301s apt stall was enough to fail it. At 800s that stall would still leave well over half the budget unused.

The stack in #2562 now self-checks as it drains — each PR gets a real check once the one below it merges and its base becomes `main`.

